PR #338, opened 2024-10-25
UI design

dechdev:
This PR adds [TailwindCSS](https://tailwindcss.com/) and [ShadCN](https://ui.shadcn.com/) component library. It retains MaterialUI, as that will be kept until all components are transitioned to Tailwind.

This PR also formalizes RPC querying patterns for most methods in Paladin.


peterbroadhurst:
I'm going to close this one out, as it's stale against the code and the design of the did get a significant change from @gabriel-indik that was worked on with design input from @sarah-burwood.

Will leave it with you @dechdev if you want to contribute ideas/code to the individual point items being done inside of #466 and the issues spawned from it.

PR #351, opened 2024-10-30
Work through full verification of the logic on Key Manager we rely on

peterbroadhurst:
Parking for now


peterbroadhurst:
Closing in favor of #539

PR #372, opened 2024-11-04
Add basic TypeScript example app

awrichar:


example/bond/scripts/abi.mjs
peterbroadhurst, line 1: Very glad you did a NPM version of this, rather than requiring users to need gradle to run the samples 👍 

... I do think though this won't have run the actual build. But an NPM user will probably be happy to just run `npx hardhat build` in the `solidity` project.

sdk/typescript/src/README.md
peterbroadhurst, line 1: Brilliant to see the starting point on this 🙌

peterbroadhurst (approved):
This is great. Will make a massive difference to developers to be able to consume the bond sample like this in Typescript/NodeJS (compared to the Golang/Java internal E2E test code).

Also loving the Typescript SDK work that's begun here.

example/bond/scripts/abi.mjs
awrichar: Yes, the README lays out the minimum steps, all of which can be done with just `npm` (in the solidity folder and then here).

PR #377, opened 2024-11-05
Build docker image workflow

dwertent:
The workflow is designed to build the Docker images for both the `paladin core` and the `operator`. Currently, this build process is triggered each time code is pushed to the `main` branch, and the image is tagged using the branch name (`main`, in this case).

To make this workflow function properly, the following repository secrets are required:
- `DOCKERHUB_USERNAME`
- `DOCKERHUB_TOKEN`

Additionally, ensure that the `paladin` and `paladin-operator` repositories have been created in the Docker registry.

For reference, here is an example of a successful workflow run:
[Example Workflow Run](https://github.com/dwertent/paladin/actions/runs/11676665992)


peterbroadhurst (approved):
Looks great. As you say it will need @ryjones help to get the right credentials in place to push to the final LFDT-labs location in dockerhub.

ryjones:
please add a `workflow_dispatch` trigger

PR #378, opened 2024-11-05
Helm release job

dwertent:
Helm-chart for paladin Operator


operator/contractpkg/main.go
```diff
@@ -129,7 +135,11 @@ func (m *ContractMap) process(name string, b *ContractMapBuild) error {
 		for libName, link := range b.LinkedLibs {
 			link = strings.ReplaceAll(link, "_", "-")
 			requiredBuilds = append(requiredBuilds, link)
-			linkedContracts[libName] = fmt.Sprintf(`{{index .status.resolvedContractAddresses "%s"}}`, link)
+			l := fmt.Sprintf(`{{index .status.resolvedContractAddresses "%s"}}`, link)
+			if helmCompatible {
+				l = fmt.Sprintf("{{`%s`}}", l)
```
dwertent, line 140: TODO: This should take place in the Makefile (when copying the contracts)

operator/charts/paladin-operator/values.yaml
```diff
@@ -1,15 +1,17 @@
...
 operator:
   name: paladin-operator
   namespace: paladin
   image:
-    repository: paladin-operator
-    tag: latest
-    pullPolicy: IfNotPresent
+    repository: kaleidoinc/paladin-operator
+    tag: release
+    pullPolicy: Always
```
peterbroadhurst, line 14: Think this is a nice add for now, but we'll likely need to come back to this as part of wider release management strategy as we go forwards.

operator/config/samples/core_v1alpha1_transactioninvoke_zeto_register_anon.yaml
```diff
@@ -23,12 +23,12 @@ spec:
     {
       "name": "Zeto_Anon",
       "implementation": {
-        "implementation":        "{{ index .status.resolvedContractAddresses "zeto-impl-anon" }}",
-        "depositVerifier":       "{{ index .status.resolvedContractAddresses "zeto-g16-check-hashes-value" }}",
-        "withdrawVerifier":      "{{ index .status.resolvedContractAddresses "zeto-g16-check-inputs-outputs" }}",
-        "batchWithdrawVerifier": "{{ index .status.resolvedContractAddresses "zeto-g16-check-inputs-outputs-batch" }}",
-        "verifier":              "{{ index .status.resolvedContractAddresses "zeto-g16-verifier-anon" }}",
-        "batchVerifier":         "{{ index .status.resolvedContractAddresses "zeto-g16-verifier-anon-batch" }}"
+        "implementation":        '{{`{{ index .status.resolvedContractAddresses "zeto-impl-anon" }}`}}',
```
peterbroadhurst, line 26: 🥸 🥸

operator/contractpkg/main.go
peterbroadhurst: Why do we need the non-helm compatible version here? Surely if this is valid YAML, then let's just do this version

peterbroadhurst (approved):
Looks great. One comment/question before merge @dwertent

operator/charts/paladin-operator/values.yaml
dwertent: Definitely!! (same with the paladin image, that will be a nightmare)

operator/config/samples/core_v1alpha1_transactioninvoke_zeto_register_anon.yaml
dwertent: Yes, the helm templating failed otherwise!

PR #388, opened 2024-11-05
[helm] Move config/samples into 'kubectl apply'

onelapahead:
If using `gradle` for E2E, the setup fails bc cert-manager CRDs are not installed (yet):

```
> Task :operator:installOperator
helm dependency build charts/paladin-operator || true
Getting updates for unmanaged Helm repositories...
...Successfully got an update from the "https://prometheus-community.github.io/helm-charts" chart repository
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "paladin" chart repository
...Successfully got an update from the "jetstack" chart repository
...Successfully got an update from the "bitnami" chart repository
Update Complete. ⎈Happy Helming!⎈
Saving 5 charts
Downloading cert-manager from repo https://charts.jetstack.io/
Downloading prometheus from repo https://prometheus-community.github.io/helm-charts
Deleting outdated charts
helm upgrade --install paladin charts/paladin-operator \
                -n default --create-namespace \
                --set operator.namespace=default \
                --set operator.image.repository=paladin.io/paladin-operator \
                --set operator.image.tag=0.0.1 \
                --set certManager.enabled=true \
                --set certManager.namespace=cert-manager \
                --set prometheus.enabled=false
Release "paladin" does not exist. Installing it now.
Error: unable to build kubernetes objects from release manifest: resource mapping not found for name: "selfsigned-issuer" namespace: "" from "": no matches for kind "Issuer" in version "cert-manager.io/v1"
ensure CRDs are installed first
make: *** [helm-install] Error 1

> Task :operator:installOperator FAILED

FAILURE: Build failed with an exception.
```

This is because Helm is actually treating the CRDs as templates (bc cert-manager wants Helm to manage the CRD lifecycle) - so it can't guarantee it installs the `Issuer` CRD before we make an `Issuer`.

However, if you look at what we're doing - we have an odd Make target that copies the samples into the chart for install. We can instead install the samples after the chart installation to avoid the chicked-egg, as well as the nasty Go template within Go template problem.

This does have the drawback that the samples are not managed by the Helm lifecycle - but considering they are samples and this is for the E2E pipeline currently, that seems like a fine tradeoff for more simplicitly.


onelapahead:
After discussing with @dwertent and @peterbroadhurst - we're going to remove `cert-manager` and `prometheus` charts as dependencies.

We can instead document for users want to use `cert-manager` how to install it separately, and our Gradle+make pipeline for E2E can test exactly that.

PR #392, opened 2024-11-05
Fix smart contract installation

dwertent:
In this pull request, I have removed the installation of the smart contracts from the chart so they will now be installed separately as standalone components. Additionally, I've implemented some fixes to improve the overall getting-started experience and the execution of Gradle end-to-end tests


operator/.gitignore
```diff
@@ -28,7 +28,7 @@ go.work
 **/test/e2e/abis/**
 
 # Generated by contractpkg tool
-**/config/samples/core_v1alpha1_smartcontractdeployment_*.yaml
+# **/config/samples/core_v1alpha1_smartcontractdeployment_*.yaml
```
onelapahead, line 31: I'm assuming this might be problematic - but I agree we should be committing codegen files and assessing their diffs if they ever do change.

My experience was there a lot of files being generated/copied around locally, as someone wanting to make a contribution it was a bit convoluted to debug the Helm chart installation process as a result.

doc-site/docs/getting-started/installation.md
```diff
@@ -54,9 +54,14 @@ helm install cert-manager --namespace cert-manager --version v1.16.1 jetstack/ce
...
-    --set paladin.namespace=paladin \
-    --set cert-manager.enabled=false
+helm upgrade --install paladin paladin/paladin-operator -n paladin --create-namespace
+```
+
+### Step 4: Deploy the CRs
+```
+curl -s https://api.github.com/repos/LF-Decentralized-Trust-labs/paladin/operator/config/samples   | \
+jq -r '.[] | select(.name | endswith(".yaml")) | .download_url' | \
+xargs -n 1 kubectl apply -n paladin -f
```
onelapahead, line 64: Darn bit of a shame `kubectl` doesn't support folders...

But speaking from I've seen projects like Prometheus and cert-manager do - I wonder if this is an indicator to us that a "set" of samples should all one big YAML file ?

Then we could do `kubectl apply -f https://raw.githubusercontent.com/LF-Decentralized-Trust-labs/paladin/refs/heads/main/operator/config/samples/gettingstarted.yaml` instead which reads a bit nicer.

And it forces us to curate different example YAMLs, and make it clear what each is good for.

A giant `samples` folder of CRs which actually have dependencies on one another is a bit terse for an end-user / new contributer.

operator/config/samples/core_v1alpha1_transactioninvoke_zeto_register_anon_enc.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   node: node1
   from: zeto.operator
   txType: public
-  toTemplate: '{{`{{ index .status.resolvedContractAddresses "zeto-factory" }}`}}'
+  toTemplate: '{{ index .status.resolvedContractAddresses "zeto-factory" }}'
```
onelapahead, line 12: This is great to me - bc its not concerned with Helm support it can be a bit more readable to users.

onelapahead (approved):
🙇🏻

operator/.gitignore
peterbroadhurst: @dwertent @onelapahead - please pause before we check in 100s of KB of data into our github history, on every recompile of a smart contract we do.

I believe developers will hate a github repo of terrible bloat like that.

dwertent: I removed this code from the Helm chart because Helm was having trouble templating the files. When I adjusted the files to allow Helm to template them successfully, the code failed to decode the objects properly.

In the long term, I think the best solution would be to create separate CRDs for the auto-generated data in the status, or to have the code load the status without creating a complex dependency between the status and the nested object.

For a quick fix to get the system working, I considered simply submitting the files. If we don’t want to go with this approach, another option is to upload the generated files as artifacts, allowing users to pull them from the release. This alternative is less intuitive but avoids committing the generated files and adding them to the Git history.

doc-site/docs/getting-started/installation.md
dwertent: I agree, let's first decide how we want to publish the CRs, once we have a decision I can consolidate the files into a single file.

PR #394, opened 2024-11-06
Pick nodes consistently in "npm run dev", and query fixes

peterbroadhurst:
## Proposed changes

Enhancements to:
- Ensure we re-run queries if we get errors
- Be consistent in `npm run dev` that `node` is on 3000


ui/client/src/contexts/ApplicationContext.tsx
```diff
@@ -32,7 +32,7 @@ export const ApplicationContextProvider = ({ children, colorMode }: Props) => {
         return 0;
       }),
     refetchInterval: constants.UPDATE_FREQUENCY_MILLISECONDS,
-    retry: false
+    retry: true
```
gabriel-indik, line 35: One issue that I see with turning `retry` to `true` is that it seems that in that case `error` will always be null. This means that if the UI cannot reach the Paladin node (for whatever reason) instead of showing an error message, the UI will just freeze. 
@dechdev do you happen to know what would be a good way to handle this scenario?

gabriel-indik (approved):

ui/client/src/contexts/ApplicationContext.tsx
dechdev: Setting `retry: true` will retry the query indefinitely ad. I don't think we want this. Instead, we should have a global max number of retries, and use the retry callback function, seen [here](https://tanstack.com/query/v4/docs/framework/react/reference/useQuery). 

```
retry: boolean | number | (failureCount: number, error: TError) => boolean
- If false, failed queries will not retry by default.
- If true, failed queries will retry infinitely.
- If set to a number, e.g. 3, failed queries will retry until the failed query count meets that number.
defaults to 3 on the client and 0 on the server
```

ui/client/src/queries/common.ts
```diff
@@ -26,16 +26,13 @@ export const generatePostReq = (stringBody: string): RequestInit => {
...
+  const res = await resFn();
   if (!res.ok && !ignoreStatuses.includes(res.status)) {
     throw new Error(errorMsg);
   }
-  try {
-    return (await res.json()).result;
-  } catch {
-    return {};
-  }
+  return (await res.json()).result;
```
gabriel-indik, line 37: The original goal of this try catch was to quietly handle cases where the value returned is not a valid JSON. Note that with this change we assume the response will always be a valid JSON.

peterbroadhurst: Very helpful to have that - I couldn't work that out from the code alone, as it seemed to be deliberately translating errors in a success payload. I thought that was the intent of the code.

A note that I didn't see a change / comments / intent described in the changes under #399, so if you have more updates to make a single approach to the different types of errors coming back to the UI true, that would be great.

The "error" categories include:
- Something that isn't JSON/RPC came back - like a `504` gateway error, which won't be JSON parsable
- An error specific to a JSON/RPC call came back, with an error code, which could be displayed to a user
- An error specific to a JSON/RPC call came back, which in that particular case the UI wants to hide from the user (like the fact there is no domain receipt on the node)
- `null` came back from a query describing a straight forward not found scenario

PR #401, opened 2024-11-06
Fix chart installation 

dwertent:
- Adjust template placeholder decoding.
- Enable replacement of the paladin image in the values file.
- Remove cert-manager from the chart dependencies.

> Note: run `gradle clean -PdeleteCluster=true` after pulling the changes


operator/Makefile
```diff
@@ -271,12 +284,10 @@ delete-node:
 DELETE_CLUSTER ?= false
 .PHONY: clean
 clean: helm-uninstall uninstall-crds ## Clean up all resources.
```
awrichar, line 286: I've found that running `make clean` or `gradle clean` when I have no kind cluster completely fails. We'll need to address this, as it should definitely be possible to iterate on `gradle build / gradle clean` as a developer without ever running kind.

operator/internal/controller/paladin_controller.go
```diff
@@ -914,7 +915,7 @@ issuerRef:
 	if tlsConf.CertSpecTemplate != "" {
 		specTemplate = tlsConf.CertSpecTemplate
 	}
-	template, err := template.New("").Option("missingkey=error").Funcs(sprig.FuncMap()).Parse(specTemplate)
+	template, err := template.New("").Option("missingkey=error").Funcs(sprig.FuncMap()).Parse(adjustTemplatePlaceholders(specTemplate))
```
peterbroadhurst, line 918: I'm sorry @dwertent, I feel a little lost to understand why there's a codechange in the Operator (rather than the generation of the CRs). Going to see if I can understand by building the Helm charts to see what's inside of them.

I assert that from the Operator's perspective, the YAML should be valid. It should contain a valid Go template (or if we don't like go templates, we use a different Template engine like `yq` etc.).

The the thing generating the YAML needs to have templating itself, like Helm, then it's responsible for doing the right escaping in the Helm chart to make the **result** of the helm chart valid YAML-containing-a-go-tempalte.

peterbroadhurst: I generated the helm chart, and the values I thought would be problematic look fine to me
```
"implementation":        '{{`{{ index .status.resolvedContractAddresses "zeto-impl-anon" }}`}}',
```

^^^ looks like should generate the right CR per Helm escaping rules... going to try and see if I can recreate a problem with this

peterbroadhurst (approved):

Issue #397, opened 2024-11-06
bug: Slow Reconciliation of Paladin Nodes and Zeto On-Chain Orchestration


onelapahead:
### What happened?

Currently its taking 20-30min for the operator to standup the 3 Besu node network, the accompanying Paladin nodes, and then deploy and invoke the various smart contracts necessary for Noto, Zeto, and Pente.

### What did you expect to happen?

We would like this to ideally take only 1-2m w/ 5m being an acceptable first improvement if possible.

We want users to have a fast and smooth getting started experience, despite the complexities of the multi-node network and demonstrating programmable privacy.

We also want developers to have an E2E pipeline which allows for faster feedback and iteration times.

### How can we reproduce it (as minimally and precisely as possible)?

Follow "Getting Started" or use `gradle deploy` locally.

### Anything else we need to know?

It is believe these are primarily optimizations that need to be made within the Operator, with the exception of the ability to have dynamic reload of registry and domain configuration by Paladin itself. The latter will be treated as a separate issue / body of work, and the Operator will hack around the lack of dynamic reload for now.

### OS version

```console
sw_vers
ProductName:		macOS
ProductVersion:		14.4.1
BuildVersion:		23E224
```

PR #402, opened 2024-11-07
[operator] Readiness for Paladins and SCD Mappings

onelapahead:
## Description

Adds a readiness check to the `PaladinReconciler` based on if the underlying `StatefulSet` is ready or not, and a controller ref to allow for watching relevant `StatefulSet` updates.

The `SmartContractDeploymentReconciler` is then updated to only reconcile a `Paladin` if the node is `Ready` and if the contracts refer to it.

Renames the `Completed` phase to `Ready`, as a `Paladin` can now go back to `Pending` if the `StatefulSet` every re-reconciles due to config changes.

## Results

One set of enhancements which seemed to help SCD reconcile speed decently in efforts to address to #397, but contracts still get stuck even after the `Paladin` goes back to `Ready` after a restart, so missing a retry or a readiness check in the right place. And ideally we avoid too many pod restarts at once, but will address that in a follow up PR.

I did have other contracts stuck due to encoding issues I believe will be addressed by https://github.com/LF-Decentralized-Trust-labs/paladin/pull/401.


operator/internal/controller/transactioninvoke_controller.go
```diff
@@ -212,12 +214,38 @@ func getContractDeploymentAddress(ctx context.Context, c client.Client, name, na
 
 }
 
+// TODO this is duplicated btwn here and SmartContractDeploymentReconciler, as reconcileAll alludes to, generics
```
peterbroadhurst, line 217: We have some helpers to allow generics. Found it surprisingly hard to support them in k8s data structures, I had to introduce a thing like this to every file:
https://github.com/LF-Decentralized-Trust-labs/paladin/blob/ef6a417de44129e9622934d3306bd1b72fe55d28/operator/internal/controller/transactioninvoke_controller.go#L49-L56

peterbroadhurst (approved):
This seems like a good step forwards on all fronts, but as we've discussed it's probably not quite the end of the journey to faster deploy times.

I note we don't have service-level readiness checks on our pod defs right now, which might be good to add (although I don't know if we can make them part of the checking process for the SCD/TXInvoke paths)

Issue #398, opened 2024-11-06
Dynamic Reload of Configuration


onelapahead:
### What would you like to be added?

The ability for Paladin to dynamically reload its configuration based on changes to config files such as domains and registries. Note that the processes should not restart/crash, but ideally only the impacted internal components are restarted without compromising any in-flight transaction processing or state.

### Why is this needed?

This allows for zero-downtime configuration updates in dynamic deployment environments such as cloud-native Kubernetes clusters.

It will allow for less "hacky" Operator code implemented when addressing https://github.com/LF-Decentralized-Trust-labs/paladin/issues/397.

onelapahead:
If we use `fsnotify` rather than target Linux only and `inotfify` - we'll want to keep https://github.com/hyperledger/firefly-common/issues/156 in mind. Whether we roll our own `fswatcher` like lib in Paladin, or re-use `firefly-common`'s.

PR #403, opened 2024-11-07
[operator] Using Container Lifecycle Hooks and Inflight Changes for Better Ordering and Delaying Restarts

### LINKED ISSUE #397: bug: Slow Reconciliation of Paladin Nodes and Zeto On-Chain Orchestration

onelapahead:
## Description

Follow up from #402. Attempts to close #397.

Initially on Getting Started, Paladin crashes some due to the Postgres sidecar not being healthy by the time the automigrations run. We order the containers so Postgres is the first container, and use a `postStart` hook that must complete before starting Paladin.

Probes are added so the Operator isn't notified Paladin is ready until its RPC port is live. This should help the deploy/invoke controllers so that they don't run unless it can actually receive JSONRPC traffic.

Then, to avoid the constantly restarting due to changing config / deploying smart contracts issue, we introduce a temporary solution of `inflight` changes map lock. This allows us to detect config changes based on the checksum, delay updating the STS until a change interval of 30s has passed, and then release the changes lock allowing the STS to update, which helps initial SCDs -> TIs -> Domains/Registries all succeed. This can be removed once #398 is completed.

## Other Changes

* Fixes the default container annotation as it must be on the pod
* Include VSCode launch config for running the Operator locally in the debugger
* Increase parallelization of reconcilers (note the internal workqueue of client-go / controller-runtime ensures parallel reconciles of the _same_ resource is not possible).
* Expontential backoff of `setupChainId` in the RPC client of Paladin to avoid any startup crashes due to Besu coming up in parallel (or being down for production scenarios)


onelapahead:
> Then, to avoid the constantly restarting due to changing config / deploying smart contracts issue, we introduce a bad hack via preStop that tells Paladin to sleep for 15s before shutting down (but the readiness probe continues to pass, allowing traffic to keep being received) which will allow it to deploy more contracts, invoke transactions, and "queue up" other config updates to hopefully help with the delays we've seen.

This isn''t working as desired, as the STS goes into a non-ready state, causing the SCDs/TIs to wait again. Instead, I'll build an inflightChange map to track when a hash has changed, but delay actually reconciling for a few seconds.

.vscode/launch.json
```diff
@@ -0,0 +1,24 @@
...
+            "name": "Run Controller",
+            "type": "go",
+            "preLaunchTask": "check-running-on-kind",
+            "request": "launch",
+            "mode": "debug",
+            "cwd": "${workspaceRoot}/operator",
+            "program": "${workspaceRoot}/operator/cmd/main.go",
+            "env": { 
+                "KUBE_LOCAL": "true",
+                "WATCH_NAMESPACE": "default"
```
onelapahead, line 17: Note - this assumes you're pointing your operator at `default` per a `gradle deploy`-esque local workflow.

.vscode/settings.json
```diff
@@ -1,6 +1,7 @@
 {
     "go.formatFlags": ["-s"],
     "go.lintTool": "golangci-lint",
+    "go.lintFlags": ["--fast", "--allow-parallel-runners"],
```
onelapahead, line 4: On my (newly installed) VSCode w/ Copilot, I was having horrendous performance with basic code inspection and autocomplete w/o upgrading my `golangci-lint` and including these settings.

operator/internal/controller/paladin_controller.go
```diff
@@ -437,7 +531,7 @@ func (r *PaladinReconciler) createPostgresPVC(ctx context.Context, node *corev1a
 }
 
 func (r *PaladinReconciler) addKeystoreSecretMounts(ss *appsv1.StatefulSet, signers []corev1alpha1.SecretBackedSigner) {
-	paladinContainer := &ss.Spec.Template.Spec.Containers[0]
+	paladinContainer := &ss.Spec.Template.Spec.Containers[1] // TODO this changes depending on PG mode
```
onelapahead, line 534: TODO: need to make this index conditional / contexted so that we can determine if its `1` or `0` depending of PG sidecare mode is enabled.

onelapahead:
> This allows us to detect config changes based on the checksum, delay updating the STS until a change interval of 30s has passed, and then release the changes lock allowing the STS to update, which helps initial SCDs -> TIs -> Domains/Registries all succeed.

Note - this is far from perfect. This does not "lock" the actual change detected, it just guarantees it won't apply any changes but every 30s. Worse case - a final change or two are queued up right after a change is applied via a pod restart, and those then cause a subsequent 30s restart.

But again an overall start up time of 1-3m compared to the previous 5-20m is a vast improvement. This become significantly less complex and faster once dynamic reload is implemented, and therefore feels like an acceptable workaround for the time being, but will defer to @dwertent @peterbroadhurst .

operator/internal/controller/check-psql.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+count=0
+
+until [ $count -ge 3 ]; do
+  psql -U postgres -c 'SELECT 1' && count=$((count + 1))
+  echo "Waiting for postgres to be ready..."
+  sleep 5
```
dwertent, line 8:
```suggestion
  sleep 1
```

dwertent: I think this will save some startup time

operator/Makefile
```diff
@@ -176,8 +176,7 @@ build-installer: manifests generate kustomize ## Build consolidated installer YA
 ##@ Helm Chart - Deployment
 .PHONY: prepare-crd-chart
 prepare-crd-chart: manifests kustomize helmify ## Build CRDs using helmify.
-	$(KUSTOMIZE) build config/default | $(HELMIFY) ${CHART_PATH_CRD}
-	find ${CHART_PATH_CRD}/templates/ -type f ! \( -name '*-crd.yaml' -o -name '_helpers.tpl' \) -exec rm -f {} +
```
dwertent, line 180: This ensures we do not copy files that are not CRDs. Is there a reason to remove it?

operator/hack/check-running-on-kind.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
```
dwertent, line 1: What is this for?

operator/internal/controller/inflight.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"sync"
+	"time"
+)
+
+// InFlight is a struct used to manage in flight requests for a unique identifier.
+type InFlight struct {
+	mux                *sync.RWMutex
+	inFlight           map[string]time.Time
```
dwertent, line 11: I highly recommend using the [maps](https://github.com/goradd/maps) package (`maps.SafeMap`), for asynchronous map operations

jimthematrix:
Testing with the branch, I saw much reduced crashloopfallback status in the paladin pods (where before the fix I saw all 3 paladin pods going through the crashloop before things get resolved/reconciled). even when I do, things get resolved within a couple of seconds, and the entire bootstrapping process was complete in under 60s for me on an M1 MacBook Pro.

operator/internal/controller/check-psql.sh
onelapahead: Yeah its tricky... the Postgres container temporarily starts its Postgres server to do some bootstrapping, shuts down, and then starts back up again.

So I've found the time needs to be longer in order to not accidentally consider Postgres "ready" too soon.

I wonder if instead I can look for the presence of a folder/file to indicate bootstrapping is done, before doing a quick health check.

operator/hack/check-running-on-kind.sh
onelapahead: If you see the VSCode launch config for running the operator in an IDE and the check task - its used by my launch config to prevent a dev from accidentally running the Paladin operator against an unintended cluster. As an SRE working across a lot of clusters daily, I appreciate when projects have safeguards in place for local development.

operator/Makefile
onelapahead: I'm using Kustomize itself to target the `config/crd` kustomization folder (not `config/default`) which therefore only output the CRDs, avoiding the need to delete extra YAML files.

operator/internal/controller/check-psql.sh
dwertent: Interesting. Maybe it's worth querying `pg_isready`? I believe that should respond only once the server is ready

dwertent (approved):
Well done!

operator/internal/controller/paladin_controller.go
onelapahead: Discussed with @dwertent - this code should just "find" the container rather than guess - in case we add any future sidecars (hopefully spinning the UI out separately).

operator/internal/controller/check-psql.sh
onelapahead: Great suggestion - testing that now

operator/internal/controller/inflight.go
onelapahead: Going to hold off on this for now just to avoid extra deps and bc this concurrent map is simple, but will keep that in mind for the future 🙇🏻

PR #404, opened 2024-11-07
E2E run out of band from main

peterbroadhurst:
The E2E is not fully worked through to run in github actions, but it currently runs on every push currently.
Which means there is a misleading red cross on the `main` branch.


hosie (approved):
LGTM

PR #406, opened 2024-11-07
More renaming and cleaning up of types

awrichar:
Rename transaction "extra data" to "domain data".

Remove testbed's TransactionInput type and use the real TransactionInput type.


peterbroadhurst (approved):
This looks good to me.

... do have an observation that we're stepping here towards consolidating more of the testbed. Know it's not ready yet to fully consolidate, but hopefully we can see a trajectory to a place where the stubs in testbed are no longer needed.

e.g. we move to running the full TxMgr in our E2E tests of domains

awrichar:
Yes, definitely trying to align the testbed as closely as possible with the real transaction manager, and hopefully continue replacing mocked pieces with real logic.

PR #409, opened 2024-11-07
Adjust sed command for linux

dwertent:
* Replaced the `sed` with go code
* Adding operator deploy

### DIFF
--- a/operator/build.gradle
+++ b/operator/build.gradle
@@ -138,7 +138,7 @@ task copyZetoSolidity(type: Copy) {
 }
 
 task buildContractSamples(type: Exec, dependsOn: [copySolidity, copyZetoSolidity]) {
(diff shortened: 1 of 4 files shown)


operator/build.gradle
```diff
@@ -138,7 +138,7 @@ task copyZetoSolidity(type: Copy) {
 }
 
 task buildContractSamples(type: Exec, dependsOn: [copySolidity, copyZetoSolidity]) {
-    commandLine "go", "run", "./contractpkg", "./contractpkg/contract_map.json"
+    commandLine "go", "run", "./contractpkg", "generate", "./contractpkg/contract_map.json"
```
awrichar, line 141: If possible, we should declare the inputs/outputs of these tasks so that Gradle knows when to re-run them (can be a follow-up).

awrichar (approved):

PR #417, opened 2024-11-08
[docs] CRDs Reference and Getting Started -> Tutorials

onelapahead:
Autogenerates the CRDs reference docs based on the comments on the Go structs - we could commit this, but worried about the likelihood of it falling out of date, etc. so thought autogenerating at build time might be better (similar to what was done w/ the getting started "samples") ? Would like to hear the maintainers' thoughts.

Otherwise, tried to help the flow of Getting Started. It was never stated you needed to clone the Git repo, and I feel like the Getting Started shouldn't require that (we publish a Helm chart and OCI image set for a reason) - so I added a simple `curl` step to download the KinD config. And then once the Getting Started is done, it didn't seem to guide me to anything that was next.

Then for the Tutorials - felt that's where it was better to explicitly state one should clone the Git repo and that there were NodeJS pre-reqs.


peterbroadhurst (approved):
Great improvements

PR #426, opened 2024-11-11
Adding release action

dwertent:
### Overview

This PR introduces a new GitHub Action workflow called `release`, which can be triggered either by creating a tag or manually.

The `release` action orchestrates the following release steps:

1. **Release Images** – Builds and releases Docker images.
2. **Release Charts** – Packages and releases Helm charts.
3. **Release CLI SDK** – Publishes the CLI SDK.

Each release step is modular and can also be executed independently as a manual action.

### DIFF
--- a/.github/workflows/release-cli-sdk.yaml
+++ b/.github/workflows/release-cli-sdk.yaml
@@ -1,0 +1,28 @@
+name: CLI SDK Release
+on:
+  workflow_call:
+    secrets:
+      NPM_TOKEN:
+        description: 'NPM token'
+        required: true
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          submodules: recursive
+      - name: Install pre-requisites
+        uses: ./.github/actions/setup
+       
+      - name: Publish to npm
+        working-directory: sdk/typescript
+        shell: bash
+        run: |
+          set -e
+          ../../gradlew build
+          npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,75 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - v*
+  workflow_dispatch:
+    inputs:
+      latest:
+        required: false
+        type: boolean
+        default: false
+      tag:
+        required: true
+        type: string
+        description: 'Release tag (chart and images)'
+        
+jobs:
+  set-variables:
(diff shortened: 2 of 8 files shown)


.github/workflows/release-cli-sdk.yaml
```diff
@@ -0,0 +1,28 @@
+name: CLI SDK Release
```
awrichar, line 1: I don't think there's any "CLI" involved here. This is the TypeScript SDK (communicates with Paladin via HTTP).

.github/workflows/release.yaml
```diff
@@ -0,0 +1,75 @@
...
+    uses: ./.github/workflows/release-charts.yaml
+    with:
+      chart_tag: ${{ needs.set-variables.outputs.tag }}
+      images_tag: ${{ needs.set-variables.outputs.tag }}
+      latest: '${{ fromJSON(needs.set-variables.outputs.latest) }}'
+    secrets:
+      CR_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  release-cli-sdk:
+    needs: release-charts
```
awrichar, line 72: Does this mean the releases are always tied together? I can see us potentially iterating rapidly on the TypeScript SDK to add additional methods (it's pretty incomplete right now). But each time we implement a method, we'll be bumping the version and releasing everything in Paladin?

dwertent: No, the TypeScript release can be executed separately. 

The dependency between the chart release and SDK release is in place for scenarios where updates have been made in the SDK that aren’t yet reflected in the chart. In these cases, it’s best to release the chart first, followed by the SDK, to ensure compatibility.

However, if you’re still developing the SDK and the required functionality is already available in the chart, you can manually release the SDK.

.github/workflows/release-typescript-sdk.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Install pre-requisites
+        uses: ./.github/actions/setup
+       
+      - name: Publish to npm
+        working-directory: sdk/typescript
+        shell: bash
+        run: |
+          set -e
+          ../../gradlew build
+          npm publish
```
awrichar, line 26: We should make sure the name in `package.json` matches the LF Decentralized Trust org. I think it should be `@lfdecentralizedtrust-labs/paladin-sdk` based on Discord conversation.

.github/workflows/release.yaml
awrichar: Perfect. I think we'll get to a point that it makes sense to release everything together. For now, I think the SDK will be playing catch-up to expose everything that Paladin already does... and I want to avoid doing releases of the operator if nothing in Paladin has actually changed, as it could be confusing to have a bunch of releases that don't actually have any new functionality.

awrichar (approved):

PR #431, opened 2024-11-13
propagate public tx options

hosie:
When invoking a contract on the pente domain, if I try to submit 2 transactions in quick succession, I am seeing error 
```
ERROR Error dispatching transaction: PD011812: Public transaction rejected: PD012216: Transaction reverted PenteInputNotAvailable
```

I believe the reason for this is that we are estimating the gas based on the current committed block but my first transaction has not been added to a block yet so the 2nd transaction revers because it attempts to spend the states produced by the first.

This highlights the need for a conversation about what the default approach should be for gas estimation for the domain.  The domain specific contracts should be fairly predictable in terms of the gas usage so it is not clear to me that we should be estimating gas for every transaction. 

In lieu of that conversation, in the meantime, this PR enables the `PublicTxOptions` ( including `gas`)  to be specified on input to `ptx_sendTransaction`.  It is already part of the input type for `ptx_sendTransaction` but we were just not propagating it.

### DIFF
--- a/core/go/internal/privatetxnmgr/sequencer_dispatch.go
+++ b/core/go/internal/privatetxnmgr/sequencer_dispatch.go
@@ -161,7 +161,7 @@ func (s *Sequencer) DispatchTransactions(ctx context.Context, dispatchableTransa
 				PublicTxInput: pldapi.PublicTxInput{
 					From:            resolvedAddrs[i],
 					To:              &s.contractAddress,
-					PublicTxOptions: pldapi.PublicTxOptions{}, // TODO: Consider propagation from paladin transaction input
+					PublicTxOptions: pt.PublicTxOptions,
 				},
 			}
 


peterbroadhurst:
@hosie - I'm not sure I understand this one yet.

If the assembly is being done with the wrong `StateContext` that is a coordinator issue.

I'm happy to review the code instead of the comment, but I'm worried about doing so without understanding the situation we're in and what we're trying to acheive.

peterbroadhurst:
Thank you - fully understand now.

> a) Do we want to be estimating the gas for every transaction? I propose the answer is no 

This seems like the immediately solution. Asking the base EVM to estimage gas for an un-executable transaction is never going to work.

> b) If we do decide to estimate gas, are we making the call correctly.

The only way I can see a possibility to do gas estimation on chained transaction, if we really decided we need it, would be to either:

1) spin up an ephemeral EVM to do it.
2) batch the transactions together at the smart contract layer, so they are _one transaction_

peterbroadhurst (approved):
Great to put this in, and the discussion was particularly helpful as we think about the next step beyond this PR.

PR #432, opened 2024-11-13
handle batch of events

hosie:
When submitting transactions to pente in quick succession, the blockchain events are being passed to the domain as a batch but it is only processing the first event in the batch.  This PR fixes the for loop logic so that all events in the batch are processed


peterbroadhurst (approved):
👍

PR #434, opened 2024-11-13
(as yet) unexplained workaround for buffer overflow

hosie:
Not sure exactly why, but this alleviates the issue reported in https://github.com/LF-Decentralized-Trust-labs/paladin/issues/433. leaving this PR in draft until we learn more about the true nature of this bug and/or it becomes a more persistent issue on main ( currently only seeing it in `dynamic-coorinator-selection` branch)

### DIFF
--- a/core/go/internal/plugins/plugin_base.go
+++ b/core/go/internal/plugins/plugin_base.go
@@ -17,6 +17,7 @@ package plugins
 import (
 	"context"
 	"fmt"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


peterbroadhurst (changes requested):
This is a fundamental large lock on an interface designed to support all different plugin types. Happy that you trying this out is help you isolate the problematic code, but I agree this is just exploratory and making a big change like single-threading request/reply of requests to a plugins would need significant thought, architecture doc'ing on the proto API, and some other things.

peterbroadhurst:
Think this has been diagnosed in #443

PR #436, opened 2024-11-14
Allow pre-resolved seeds

peterbroadhurst:
Currently the config-time loading of seed key materials, only supports key stores where the resolution of a `name`/`path`/`index` is a fully-deterministic loading instruction for a key material.

However, there are cases where the result of this must be stored back to the DB in key manager when resolving keys. Particularly when using "master key encryption" approaches, where the `keyHandle` will be an encrypted representation of a key that can only be decrypted by the HSM.

This PR proposes the `keyHandle` can be supplied directly in the configuration for seed loading, mimicking the post-resolution state - where the `keyHandle` has been persisted. This allows the configuration to:
- Bypass  `FindOrCreateLoadableKey`
- Jump straight to `LoadKeyMaterial`


dwertent (approved):
LGTM

PR #437, opened 2024-11-15
fix: Make copyright information consistent

SamMayWork:
Noticed a lot of files without the copyright information

### DIFF
--- a/operator/pkg/config/config.go
+++ b/operator/pkg/config/config.go
@@ -1,3 +1,19 @@
+/*
+	Copyright 2024.
+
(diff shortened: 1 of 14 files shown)


hosie (approved):
LGTM. Observe that we do have 3 different styles of comments in various copyright headers across the code base.

I think we prefer single line comments in files that have build tags ( otherwise the formatter tries to put the build tag comment at the top) but I have not figured out if there is consistency of where we apply one style vs the other.  It seems like this PR sticks with whatever style was in the respective directory.

PR #440, opened 2024-11-15
Support secrets for paladin

dwertent:


operator/api/v1alpha1/paladin_types.go
```diff
@@ -37,6 +37,10 @@ type PaladinSpec struct {
 	// (vs. configuring a connection to a production blockchain network)
 	BesuNode string `json:"besuNode,omitempty"`
 
+	// AuthConfig is used to provide authentication details for blockchain connections
+	// If this is set, it will override the auth details in the config
```
onelapahead, line 41: I imagine we'll eventually need the same for Postgres auth security as well ?

onelapahead (approved):
Overall looks good to me, just not sure about the spelling/structure for `authConfig` in the long-term. As I mentioned there will be Postgres auth eventually, and likely remote key signer auth...

maybe something like:
```yaml
spec:
  authConfig:
     evmJSONRPC:
        secretRef:
     postgres:
         secretRef:
     # so on ...
```

Another need for `Secrets` might be specifying the username/password (or TLS certs) that secures Paladin's own JSONRPC endpoints. So many "auths", the CRD API needs to make it clear what auth configs for the frontdoor i.e. Paladin endpoints vs backdoor i.e. downstream dependencies.

operator/api/v1alpha1/paladin_types.go
hosie: Agree with @onelapahead here that we should be explicit so that we leave room for future proofing.  That could be  a single`auth` struct with fields for the different clients as @onelapahead suggests.   An alternative spelling might be to extend (or wrap) the `besuNode` struct that we already have in the CRD.  Currently, it must be a string that is interpreted as a  reference to a `Besu` CR but maybe this could become a struct that could either contain a object reference or a set of creds (by reference to secret) and the URL to an actual blockchain node. e.g.

i.e. either of the following would be valid
```yaml
baseLedger:
  besuNode: objectRef
```
or

```yaml
baseLedger:
  endpoints:
    jsonRPCUrl: ...
    wsUrl: ...
    authConfig: 
      ...
```

Although, I am assuming here that the same auth would always be used for JSONRPC endpoint and WS endpoint.  Would need to confirm if that is a safe assumption.

hosie (approved):
LGTM

operator/api/v1alpha1/paladin_types.go
dwertent: I was considering how much we want to "replicate" the Paladin configuration here. It's starting to resemble the Paladin config itself, so perhaps it would be better to define the Paladin config as a typed structure directly in the CRD definition, rather than keeping it as a blob

PR #441, opened 2024-11-17
Refactor nonce assignment to post-submit of public transactions

peterbroadhurst:
## The problem

The "2 phase submit" refactor of the public transaction manager DB transaction model in #179, introduced a problem with nonce allocation. Specifically on restart when there is a public transaction in the DB that is not yet submitted to the mempool of the blockchain, the nonce allocation attempts to re-use a nonce already allocated in the DB of the node.

Digging into this showed that we had a significant issue to resolve in the database transaction model, between the `Transaction Manager`/`Private Transaction Manager` submitting the transaction, and the `Public Transaction Manager` attempting to allocate a nonce.

Inherited from previous generations, we were assigning the nonce in-line with the submission, but this is problematic when we have batch database transactions. We need to know in this moment what other nonces are in-flight for that signing `from` address, across the database, and the `eth_getTransactionCount` call against the blockchain.

This is further complicated by the fact we might be submitting batches of public transactions covering multiple signing addresses. Two threads racing could hit A-B, B-A deadlocks if we took fine grained locks on an individual signing address. Then course grained locks would be very problematic for transaction submission performance.

## The solution

This PR splits nonce assignment out from transaction submission.

- A new `localId` (internally referred to as `pub_tx_id` in the DB) is assigned to every `PublicTx`
    - This is a numeric sequence managed by the DB
    - If two threads are racing with active DB transactions, then a lower number might be committed before a higher number
    - However, within one thread/DB-TX the numbers are assured to be increasing
- Nonce allocation is deferred to the `orchestrator`
    - There is only one of these per signing address, so it is ideally positioned to do the work
    - This thread polls from the DB in `pub_txn_id` ascending sequence
    - It might rewind in the case of racing DB TXs described above, so nonces...
        - **Are assured** to be allocated in ascending order to txs committed in the _same_ DB transaction
        - **Are not assured** to be in strictly ascending order compared to `pub_txn_id`
- The previous `signer_nonce` string column is removed as the primary join between public TX tables
    - We can no longer count on having a `nonce` at creation time of the DB record
    - Instead we use the `pub_txn_id` as the join entry
    - This required some quite complex DB migrations

## DB Migration notes

Due to the early phase of the release cycle, I propose the following compromises to limit the complexity of the DB migrations:

- There are no `down` migrations to re-instate the old `signer_nonce` behavior, for either DB
- For PostgreSQL data is retained on the `up` migration
- For SQLite data is **not** retained on the `up` migration

## Other fixes

- Remove the 30s delay in unit tests introduced in #403 due to `ethclient` having an internal retry for availability of the node. This code was moved to `componentmgr`, and changed to be configurable, and use the default `retry` package.
- Set `--tx-pool-limit-by-account-percentage` to `1.0` in the default k8s Besu setup to avoid:
    ```
    [2024-11-19T18:30:13.832Z] ERROR RPC[000000061] <-- ERROR: Transaction nonce is too distant from current sender nonce
    ```
- Stop creating un-used database tables `transaction_delegations` and `transaction_delegation_acknowledgements`
- Remove syncpoint code associated with said DB tables
- Introduce a deterministic election of a coordinator for any endorsement set based on deterministic sorting of endorsement parties, and unique nodes, and simple numeric hash to determine coordinator.
- ~Resolve this situation where delegations cross in the post, by adding a `blockHeight` to delegations, and when an incoming delegation-request comes in at a higher block height than an outgoing delegation-request, override the delegation and become the coordinator.~ This fix is incomplete, as it resulted in follow-on issues, which @hosie will work on over coming sprint.
    ```
    /tmp/t1.txt:[2024-11-19T19:01:40.941Z] DEBUG SelectCoordinatorNode: selected coordinator node node2 using round robin algorithm for blockHeight: 6497 and rangeSize 100  pid=1 role=pctm-loop-0x89e639ba9426c6b1f12a4a52cc5d92f56bffbf02
    /tmp/t1.txt:[2024-11-19T19:01:40.944Z]  INFO transactionFlow:Action TransactionID='f352b82e-464c-4dc4-b112-f1a6f6150fb8' Status='delegating' LatestEvent='TransactionSubmittedEvent' LatestError='' : Delegating transaction to node2 pid=1 role=pctm-loop-0x89e639ba9426c6b1f12a4a52cc5d92f56bffbf02
    
    
    /tmp/t3.txt:[2024-11-19T19:01:43.063Z] DEBUG SelectCoordinatorNode: selected coordinator node node2 using round robin algorithm for blockHeight: 6499 and rangeSize 100  pid=1 role=pctm-loop-0x89e639ba9426c6b1f12a4a52cc5d92f56bffbf02
    /tmp/t3.txt:[2024-11-19T19:01:43.068Z]  INFO transactionFlow:Action TransactionID='f352b82e-464c-4dc4-b112-f1a6f6150fb8' Status='delegating' LatestEvent='TransactionSwappedInEvent' LatestError='' : Delegating transaction to node2 pid=1 role=pctm-loop-0x89e639ba9426c6b1f12a4a52cc5d92f56bffbf02
    
    /tmp/t2.txt:[2024-11-19T19:01:43.055Z] DEBUG SelectCoordinatorNode: selected coordinator node node3 using round robin algorithm for blockHeight: 6500 and rangeSize 100  pid=1 role=pctm-loop-0x89e639ba9426c6b1f12a4a52cc5d92f56bffbf02
    /tmp/t2.txt:[2024-11-19T19:01:43.073Z]  INFO transactionFlow:Action TransactionID='f352b82e-464c-4dc4-b112-f1a6f6150fb8' Status='delegating' LatestEvent='TransactionAssembledEvent' LatestError='' : Delegating transaction to node3 pid=1 role=pctm-loop-0x89e639ba9426c6b1f12a4a52cc5d92f56bffbf02
    ```


core/go/componenttest/component_test.go
```diff
@@ -41,56 +40,16 @@ import (
...
 	"gorm.io/gorm"
-	"sigs.k8s.io/yaml"
 )
 
 func TestRunSimpleStorageEthTransaction(t *testing.T) {
 	//TODO refactor this to be more black box by using JSONRPC interface to invoke the public contract
 	ctx := context.Background()
 	logrus.SetLevel(logrus.DebugLevel)
 
-	var testConfig pldconf.PaladinConfig
```
peterbroadhurst, line 52: This whole block wasn't actually used.

config/pkg/pldconf/startup.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"github.com/kaleido-io/paladin/config/pkg/confutil"
+)
+
+type StartupConfig struct {
+	BlockchainConnectRetry RetryConfigWithMax `json:"blockchainConnectRetry"`
+}
+
+var StartupConfigDefaults = StartupConfig{
```
peterbroadhurst, line 27: This is the replacement of the logic added to `ethclient` in https://github.com/LF-Decentralized-Trust-labs/paladin/commit/e81df2f743e5b685a0ed13fe5434c87524318b32 as part of #403

core/go/db/migrations/sqlite/000013_delayed_nonce_allocation_updates.up.sql
peterbroadhurst, line 1: The reasoning for the different behavior with SQLite, is that you cannot edit foreign key relations in SQLite. So deletion and recreation of the tables is the right approach.

The complexity of adding a data copy across for SQLite from the old to the new tables (given it is a non-production option) did not seem worthwhile.

core/go/internal/components/domainmgr.go
```diff
@@ -41,7 +41,7 @@ type DomainManager interface {
 	ConfiguredDomains() map[string]*pldconf.PluginConfig
 	DomainRegistered(name string, toDomain DomainManagerToDomain) (fromDomain plugintk.DomainCallbacks, err error)
 	GetDomainByName(ctx context.Context, name string) (Domain, error)
-	GetSmartContractByAddress(ctx context.Context, addr tktypes.EthAddress) (DomainSmartContract, error)
+	GetSmartContractByAddress(ctx context.Context, dbTX *gorm.DB, addr tktypes.EthAddress) (DomainSmartContract, error)
```
peterbroadhurst, line 44: Required to avoid stacking DB transactions

core/go/internal/components/privatetxmgr.go
```diff
@@ -67,7 +67,7 @@ type PrivateTxManager interface {
 	TransportClient
 
 	//Synchronous functions to submit a new private transaction
-	HandleNewTx(ctx context.Context, tx *ValidatedTransaction) error
+	HandleNewTx(ctx context.Context, dbTX *gorm.DB, tx *ValidatedTransaction) error
```
peterbroadhurst, line 70: Required to avoid stacking DB transactions

core/go/internal/components/publictxmgr.go
```diff
@@ -77,7 +60,14 @@ type PublicTxManager interface {
...
 	GetPublicTransactionForHash(ctx context.Context, dbTX *gorm.DB, hash tktypes.Bytes32) (*pldapi.PublicTxWithBinding, error)
-	PrepareSubmissionBatch(ctx context.Context, transactions []*PublicTxSubmission) (batch PublicTxBatch, err error)
+
+	// Perform (potentially expensive) transaction level validation, such as gas estimation. Call before starting a DB transaction
+	ValidateTransaction(ctx context.Context, dbTX *gorm.DB, transaction *PublicTxSubmission) error
+	// Write a set of validated transactions to the public TX mgr database, notifying the relevant orchestrator(s) to wake, assign nonces, and start the submission process
+	WriteNewTransactions(ctx context.Context, dbTX *gorm.DB, transactions []*PublicTxSubmission) (func(), []*pldapi.PublicTx, error)
+	// Convenience function that does ValidateTransaction+WriteNewTransactions for a single Tx
+	SingleTransactionSubmit(ctx context.Context, transaction *PublicTxSubmission) (*pldapi.PublicTx, error)
+
```
peterbroadhurst, line 70: This is the significant change. The API is massively simplified here, to just a validate call (which does the `eth_estimateGas` in simple scenarios where the `gas` limit is not provided by the caller), and a batch write.

Note the `func()` returned by the `WriteNewTransactions` - this **must** be called outside of the DB transaction on success (only) to trigger processing of the newly added public transactions.

core/go/internal/privatetxnmgr/private_txn_mgr_test.go
```diff
@@ -78,6 +77,32 @@ var testABI = abi.ABI{
 	},
 }
 
+func mockWritePublicTxsOk(mocks *dependencyMocks) chan struct{} {
```
peterbroadhurst, line 80: I think the tests in this file might benefit from a few more helpers like this, to minimize the boilerplate in tests.

core/go/internal/privatetxnmgr/syncpoints/dispatch.go
```diff
@@ -44,12 +41,12 @@ type DispatchPersisted struct {
 	ID                       string             `json:"id"`
 	PrivateTransactionID     string             `json:"privateTransactionID"`
```
peterbroadhurst, line 42: I couldn't explain to myself as I was working on this:
- Why this is not a `uuid`
- Why there is no foreign key relationship

... but I think @hosie you are working in this area, so I decided not to make any changes.

core/go/internal/publictxmgr/in_flight_transaction_stage_controller.go
```diff
@@ -315,7 +315,8 @@ func (it *inFlightTransactionStageController) ProduceLatestInFlightStageContext(
 									rsc.InMemoryTx.GetGasPriceObject()
 									gasPriceJSON, _ := json.Marshal(rsc.InMemoryTx.GetGasPriceObject())
 									rsc.StageOutputsToBePersisted.TxUpdates.NewSubmission = &DBPubTxnSubmission{
-										SignerNonce:     rsc.InMemoryTx.GetSignerNonce(),
+										from:            rsc.InMemoryTx.GetFrom().String(),
```
peterbroadhurst, line 318: This is just so we have a key to use to dispatch all submissions for the same `from` address, to the same writer (as the submission itself doesn't have a from address in it)

core/go/internal/publictxmgr/transaction_orchestrator.go
```diff
@@ -220,6 +229,102 @@ func (oc *orchestrator) getFirstInFlight() (ift *inFlightTransactionStageControl
 	return
 }
 
+func (oc *orchestrator) initNextNonceFromDBRetry(ctx context.Context) error {
+	return oc.retry.Do(ctx, func(attempt int) (retryable bool, err error) {
+		return true, oc.initNextNonceFromDB(ctx)
+	})
+}
+
+func (oc *orchestrator) initNextNonceFromDB(ctx context.Context) error {
```
peterbroadhurst, line 238: On startup of the `orchestrator` we read the highest nonce, as we aren't assured to actually page that one into memory and start processing it.

core/go/internal/publictxmgr/transaction_orchestrator.go
```diff
@@ -220,6 +229,102 @@ func (oc *orchestrator) getFirstInFlight() (ift *inFlightTransactionStageControl
...
+	// Update the txns themselves, and our nextNonce
+	for i, tx := range toAlloc {
+		nonce := newNonces[i]
+		tx.Nonce = &nonce
+	}
+	oc.lastNonceAlloc = time.Now()
+	oc.nextNonce = &newNextNonce
+
+	return nil
+}
```
peterbroadhurst, line 326: This is the most important part of the change - the synchronous code to ensure we assign nonces to all transactions, before we do any processing on them.

It took me quite a bit of work to determine how in PSQL/SQLite I could use a single SQL call to `UPATE` and set multiple values (as this is very hot code performance wise). Further optimization might be possible.

core/go/pkg/ethclient/client.go
```diff
@@ -193,33 +192,13 @@ func (ec *ethClient) ChainID() int64 {
 	return ec.chainID
 }
 
-// setupChainID is a helper to get the chain ID from the node
-// we assume the node could intermittently have issues, be starting up
-// in paralle, etc. and therefore implement an expotential backoff mechanism
-// in the event of initial failure.
 func (ec *ethClient) setupChainID(ctx context.Context) error {
```
peterbroadhurst, line 195: This is just backing out the change from #403

core/go/pkg/persistence/testdb_sqlite.go
```diff
@@ -46,19 +42,3 @@ func NewUnitTestPersistence(ctx context.Context, suite string) (Persistence, fun
 	})
 	return p, func() { p.Close() }, err
 }
-
-func TestMigrateUpDown(t *testing.T) {
```
peterbroadhurst, line 50: These tests weren't actually being run - and I found some issues with broken `down` migrations once I re-instated them.

testinfra/docker-compose-test.yml
```diff
@@ -14,7 +14,7 @@ services:
     command:
       - /var/besu
   besu:
-    image: hyperledger/besu:latest
+    image: hyperledger/besu:24.10.0
```
peterbroadhurst, line 17: I found `latest` on my laptop was an old release, that didn't have the empty blocks modification. So it seemed good for us all to be testing with a consistent release.

toolkit/go/pkg/log/log.go
```diff
@@ -53,10 +53,12 @@ func InitConfig(conf *pldconf.LogConfig) {
 	output := confutil.StringNotEmpty(conf.Output, *pldconf.LogDefaults.Output)
 	switch output {
 	case "file":
+		filename := confutil.StringNotEmpty(conf.File.Filename, *pldconf.LogDefaults.File.Filename)
+		rootLogger.Infof("Logs diverted to %s", filename)
```
peterbroadhurst, line 57: This confused me for a bit on the zeto tests (as they take a long time with no console output), so I thought this would be helpful to remind us where the logs are going.

core/go/internal/privatetxnmgr/private_txn_mgr_test.go
hosie: 100% agree.  Been trying to evolve towards more util and less boilerplate whilst trying not to overcomplicate the utils or else we will need tests for the utils.

I like this addition.

core/go/internal/privatetxnmgr/syncpoints/dispatch.go
hosie: I can't explain why this is not a UUID.  Don't remember what (if) I was thinking at the time of writing the code or maybe I just didn't have a good grasp on what things we were going to be settling on UUIDs for.  Eitherway, agree it should be uuid and should have cascading delete.

hosie (approved):
LGTM

PR #445, opened 2024-11-19
Update Paladin configurations and add VS Code launch configurations

dwertent:
#### 1. **Enhancements to `baseLedgerEndpoint` Configuration**
   - The `baseLedgerEndpoint` in the Paladin node configurations now supports specifying the endpoint type (`local` or `endpoint`) with the corresponding details:
     - **Local Configuration**:
       ```yaml
       spec:
         baseLedgerEndpoint:
           type: local
           local:
             nodeName: node2
       ```
     - **Network Configuration**:
       ```yaml
       spec:
         baseLedgerEndpoint:
           type: endpoint
           endpoint:
             jsonrpc: "https://besu.node"
             ws: "wss://besu.mode"
             auth:
               type: secret
               secret:
                 name: besu-secret
       ```

#### 2. **Build Enhancements**
   - The `build.gradle` file has been updated to include a new task for preparing artifacts.
   - The github action will release the artifacts as well, this will enable installing the CRs independently from the operator installation . [Here is an example](https://github.com/dwertent/paladin/releases/tag/v0.0.5).


operator/config/crd/bases/core.paladin.io_paladins.yaml
```diff
@@ -68,10 +68,68 @@ spec:
...
                 type: object
+              baseLedgerEndpoint:
+                description: BaseLedgerEndpoint specifies the base endpoint for the
+                  ledger
+                properties:
+                  endpoint:
+                    description: Network specifies the configuration when the type
+                      is 'network'.
+                    properties:
+                      authConfig:
```
hosie, line 79: the `Config` part of this field name seems redundant.  Should it just be `auth` ?

operator/config/crd/bases/core.paladin.io_paladins.yaml
```diff
@@ -68,10 +68,68 @@ spec:
...
+                        properties:
+                          authMethod:
```
hosie, line 81: should this just be `method` given that it is already a child of a field named `authConfig` ( or `auth`)

PR #459, opened 2024-12-02
Update BondTest to offer better counterparty protection

awrichar:
Rather than authorizing a `PentePrivacyGroup` contract to execute a bundle of operations directly (which would allow any private contract in the group to trigger any of the operations), direct the `PentePrivacyGroup` to deploy a separate `Atom` on the base ledger, locked to only the desired set of operations. Delegation to this `Atom` will therefore _only_ allow the intended atomic transaction to take place all together.


domains/pente/src/test/java/io/kaleido/paladin/pente/domain/BondTest.java
```diff
@@ -221,15 +233,23 @@ void testBond() throws Exception {
...
+            // Create the atom factory on the base ledger
+            String atomFactoryAddress = testbed.getRpcClient().request("testbed_deployBytecode",
+                    "issuer",
+                    atomFactoryABI,
+                    atomFactoryBytecode,
+                    new HashMap<String, String>());
 
             // Alice deploys BondSubscription to the alice/custodian privacy group, to request subscription
             // TODO: if Alice deploys, how can custodian trust it's the correct logic?
             var bondSubscription = BondSubscriptionHelper.deploy(aliceCustodianInstance, alice, new HashMap<>() {{
```
peterbroadhurst, line 248: I found this a good reminder for me to google the initializer syntax, and I found some very good guidance that this anonymous-innerclass+initializer syntax is actually pretty nasty.

Fine for test, but not great for production code... and we're currently using it in both.

Instead we should move over to `Map.of(...` syntax: https://stackoverflow.com/a/59366753/21677265

peterbroadhurst (approved):
This looks great @awrichar 

One thought for a follow-up - would it make sense to provide a little more diagram information in the sample to help people understand the steps as you go through. It's a lot of steps, and it's hard to follow.

I could imagine like a presentation where the steps build up over time until the end, we could maybe put those charts in as visual cues alongside the code snippets.

awrichar:
Thanks - I agree we need some good diagrams. I've started scribbling a few, but probably needs more work before they can go in the docs. @jimthematrix was also trying to follow this tutorial and sketch as he went, so between us we can come up with something.

PR #463, opened 2024-12-03
distributed sequencer algorithm

hosie:
This PR introduces the detail documentation of the distributed sequencer protocol.

In domains ( such as Pente) where the spending rules for states allow any one of a group of parties to spend the state, then we need to coordinate the assembly of transactions across multiple nodes so that we can maximize the throughput by speculative spending new states and avoid transactions being reverted due to double concurrent spending / state contention. The distributed sequencer protocol described in this PR is the formal specification of how that operates and the responsibilities and expectations of each node involved.

This goes beyond describing what happens to be implemented in code at this moment in time and aims to be comprehensive in the specification of the algorithm. Further code changes PRs will be opened to bring the code inline with this architecture once it has been agreed.

Includes explicit (but concise) write up of the key architectural decisions made and the consequences of those compared to the alternatives.

Remaining TODOs before moving out of `draft`
Tweaks to protocol

 - [x] remove the `dispatchNotification` protocol point and replace with heartbeat messages from submitter
    - [x] as a consequence of this, `dispatchConfirmation` has little value being a persistence point on the sender node so update the diagrams to not assume persistence here
 - [x] heartbeat messages should go to everyone in the group to help with detection of active groups in e.g. in cases of coordinator revival
 - [ ] Delegation requests should include a field that states whether this is the first choice or not to help the proposed delegate decide if it agrees.
 - [x] Remove any mention of `DelegationReturned` message or possibly include it in a "potential future optimizations" section.  The protocol should not rely on this.
   
Fill in gaps / corrections in writeup
    
 - [x] add Transaction lifecycle state diagram including persistence points
 - [ ] add details on message exchange for domain context and how it helps with a smooth natural (happy path) coordiantor switchover and how the protocol allows the system to operate in certain error cases where that is not passed
 - [ ] add timeout cases to `TransactionDelegation` and `Endorsement` message exchanges
 - [ ] add section on configuration ( timeouts, range size etc...) and how they can be changed
 - [ ] fill out more details and fix diagram for `REQUEST DISPATCH CONFIRMATION`
 - [ ] complete details on `Check availability` sub process
 - [ ] `Endorsement request`  add flowchart and include where the startup message fits in
 - [ ] `Gather endorsement` add flowchart
 - [ ] `Select available coordinator` fix diagram and complete description
 - [ ] add section for `message-exchange-transaction-confirmation`
 - [ ] complete Terminology section

Formatting

 - [ ] some of the sequence diagrams have so many lanes, they get scaled down so far that they are difficult to read without zoom.  Either break them down to smaller sequencers, or combine lanes ( e.g. fold the database layers in) and/or pre-build the svg and include it as and image so that the reader can at least open the image in a separate tab
 - [ ] dark mode need to tweak the css so that the images look good in dark model.  This will require all images to be inline SVG or to have 2 copies of each image.  I think I prefer 2 copies because inline SVG has the downside that chrome does not give the option to "open image in new tab" and sometimes that is useful.
 - [ ] In various high level discussions (e.g. scenario walkthroughs) we make reference to message types and message exchange handshakes but do not currently include hyperlinks.
 - [ ] could possibly break down the "Senders Responsibility" section.  Maybe pull out failover to a separate section so that it shows up in the right nav.
 - [ ] the actual protocol points ( message exchanges and sub-processes) do not appear in the right nav.  Restructure the hierarchy levels so that they do


doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+ - there is no need for election `term`s in this algorithm.
+ - when coordinator responsibility is switched to another node, each inflight transaction is either re-assigned to the new coordinator or flushed through to confirmation on the base ledger
+ - the sender node for each transaction is responsible for ensuring that the transaction always has one coordinator actively coordinating it by detecting and responding to situations where the current coordinator becomes available, a more preferred coordinator comes back online or preference changes due to new block height.  (unless the sender deems the transaction to be no longer valid in which case, it marks it as reverted).
+ - The sender node continues to monitor and control the delegation of its transaction until it has received receipt of the transactions' confirmations on the base ledger. This provides an "at least once" quality of service for every transaction.
```
peterbroadhurst, line 28:
```suggestion
 - The sender node continues to monitor and control the delegation of its transaction until it has received receipt of the transactions' confirmations on the base ledger. This provides an "at least once" quality of service for every transaction at the distributed sequencer layer. As described earlier the blockchain enforces "at most once" semantics, so there is no possibility of duplicate transactions.
```

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+ - The handshake between the sender node and the coordinator node(s) attempts to minimize the likelihood of the same transaction intent resulting in 2 valid base ledger transactions but cannot eliminate that possibility completely so there is protection against duplicate intent fulfillment in the base ledger contract
+
+## Detailed Walkthrough
+To describe the algorithm in detail, we break it down to sub problems
+
+### Composition of committee
```
peterbroadhurst, line 34:
```suggestion
### Composition of committee

An smart contract instance of a domain opts into use of the distributed sequencer algorithm, and initializes it with the composition of the committee during the `InitContract` stage. This occurs individual on each node in the committee based on the on-chain configuration information that is written to the base ledger.
```

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+When creating a new Pente domain contract instance ( i.e. privacy group) the list of nodes that are candidates for coordinator selection is declared.  For Pente this happens to be exactly equal to the privacy group but potentially could theoretically be a subset of the privacy group.
```
peterbroadhurst, line 35:
```suggestion
The Pente privacy group smart contract, is an example of domain that uses the dynamic sequencer algorithm. The list of nodes that are candidates for coordinator selection is declared to be exactly equal to the privacy group (future versions might choose to this to be a subset of the privacy group).
```

peterbroadhurst: I believe it is significant to the specification whether it is the node or the identity that is the member of the committee. I have asserted in this comment (aligned with the code in Pente) that is is the fully qualified identity.

This means that you could have a committee of 4, even with a single node.

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+
+Similarly, any other future domain that opts in to the dynamic coordinator selection policy,  must provide this information as part of the constructor handshake between the domain and the core paladin engine.
+
+### Coordinator selection
+
+This is a form of leader election but is achieved through deterministic calculation for a given point in time ( block height), given predefined members of the group, and a common (majority) awareness of the current availability of each member. So there is no actual "election" term needed.
```
peterbroadhurst, line 41: It's a pedantic words thing, but leader election is the _outcome of the distributed algorithm_, rather than the thingn that is performed directly by Coordinator selection on each given node.

e.g. leader election is not performed using coordinator selection on a single node in isolation. Rather leader election is a/the property that emerges from every node following the full specification in this document, and _one component_ is a deterministic decision made independently by each node to come to the same conclusion.

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+
+Each node is allocated positions on a hash ring.  These positions are a function of the node name.  Given that the entire committee agrees on the names of all nodes, all nodes independently arrive at the same decision about the positions of all nodes.
```
peterbroadhurst, line 43: See note above about needing to be really specific about if it is `node name` or `committee member identifier`

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+
+For the current block height, a position on the ring is calculated from a deterministic function.  The node that is closest to that position on the ring is selected as the coordinator.  If that node is not available, then the next closest node is selected. If that node is not available, then the next closest again is selected, and so on.
+
+For each Pente domain instance ( smart contract API) the coordinator is selected by choosing one of the nodes in the configured privacy group of that contract
```
peterbroadhurst, line 47:
```suggestion
For each smart contract instance in the domain the coordinator is selected by choosing one of the nodes in the configured privacy group of that contract
```

(note I'm not calling out the ❓ on `node` being the right term here, as I've already mentioned that a couple of times)

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+
+ - the selector is a pure function where the inputs are the node names + the current block number and the output is the name of one of the nodes
+ - the function will return the same output for a range of `n` consecutive blocks ( where `n` is a configuration parameter of the policy)
```
peterbroadhurst, line 50: I don't think we have to solve for it in the first version of this specification, but I note that the ability to change `n` over time is something that might be necessary for tuning the algorithm on long-running privacy groups.

It probably fits into a wider note on "changing configuration"

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+
+
+### Node availability detection
+
+When a node starts acting as the coordinator, then it  periodically sends a heartbeat message to all nodes that for which it has an active delegation and will accept delegation requests from any other node in the group. 
+
+All sender nodes keep track of the current selected coordinator ( either by reevaluating the deterministic function periodically or caching the result until block height exceeds the current range boundary).  If they fail to receive a heartbeat message from the current coordinator, then they will chose the next closest node in the hashring (e.g. by reevaluating the hashring function with the unavailable nodes removed).
```
peterbroadhurst, line 64: What is the intended behavior of the algorithm in this scenario?

- Highest ranked coordinator is B
- D delegates transaction to B
- B stops for a period
    - Longer than the heartbeat interval
    - Shorter than it takes for the block-range to change
- A is the second-highest scored coordinator
- D delegates transaction to A (because it gives up on B)
- B becomes available again
- C delegates transaction to B

At this point D has delegated a TX to A, but C has delegated to B.
I haven't (yet at this point in the spec) understood what would cause A and B to notice that they are both active (because at the time A became the coordinator, B was not available to ask).

peterbroadhurst: Ah, ok found this statement below:

> If Node A comes back online, then it will begin transmitting heartbeat messages again because it has selected itself as coordinator given the current block height.

This seems complex to implement - what would cause A to evaluate that it selected itself as the coordinator? It does not have any transactions after the the restart. Why would this contract (of the infinite potential contracts) be one after restart it would choose to select itself as a coordinator?

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+
+`A-1` , `B-1`, `C-1` and `D-1` have already been submitted to base ledger and confirmed.  
+`A-2` , `B-2`, `C-2` and `D-2` have been prepared for submission, assigned a signing address and a nonce but have not yet been mined into a confirmed block.  NOTE it is irrelevant whether these transactions have been the subject of an `eth_sendTransaction` yet or not.  i.e. whether they only exist in the memory of the paladin node's transaction manager or whether they exist in the mempool of node A's blockchain node.  What is important is that
+
+ - they have been persisted to node A's database as being "dispatched" and
+ - are not included in block 3 or earlier
+
+![image](../images/distributed_sequencer_switchover_frame1.svg){ width="500" }
+
+As all nodes in the group ( including node A) receive receipts of block 3, they will recognize node B as the new coordinator
```
peterbroadhurst, line 147: I did not understand this sentence.

This implies some future inference model and/or state tracking that seems very complex to achieve, and it's unclear to me the justification for this.

I note I believe this different statement is true, but it's currently the only thing I understand to be true with respect to evaluation of block height:

> When each nodes in the group is notified that the block height of the chain has reached 4 (happens at different times on each node) they will recognize node B as the highest scoring coordinator when they have reason to run the scoring algorithm

peterbroadhurst: Ah, ok - I think I understand.
You are saying that the the combination of these two things will cause the scoring algorithm to occur:
- Detection of block height moving to 4
- The fact that transactions are unresolved for this contract address

This implies the algorithm re-evaluates the local "sender state" on detection of each block.

However, I'm worried I'm missing some detail as you say:

> receive **receipts** of block **3**

I'm not clear what receiving receipts means in this case, how it relates to block confirmations, or why it's block 3 (rather than 4) that's relevant here.

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+![image](../images/distributed_sequencer_switchover_frame3.svg){ width="500" }
+
+Node A sends a `DelegationReturned` message to the respective sender nodes for each non dispatched transaction that it had in flight.
+
+
+![image](../images/distributed_sequencer_switchover_frame4.svg){ width="500" }
+
+The final transmission from node A, as coordinator, is a message to node B with details of the speculative domain context predicting confirmation of transactions `A-2` , `B-2`, `C-2` and `D-2` .  This includes the state IDs for any new states created by those transactions and any states that are spent by those transaction.  This message also includes the transaction hash for the final transaction that is submitted on the sequence coordinated by node A.  Node B needs to receive this message so that it can proceed with maximum efficiency. Therefore this message is sent with an assured delivery quality of service (similar to state distribution).  
+
+Meanwhile, all nodes delegate transactions  `A-3` , `B-3`, `C-3` and `D-3` to node B.
```
peterbroadhurst, line 163: I'm struggling to understand the assumptions for this "meanwhile".

What makes it safe to send anything to B, if you are not absolutely certain that A has transferred your previous transactions to B?

The message from C->B that is sent by C because `DelegationReturned ` was received by C could (in fact in the above sequence is reasonably likely) to reach B, before the message from A->B that provides information about B's previous transactions.

I wonder, given these facts:
- `DelegationReturned` will be sent by A->C after any message (sorry couldn't see the name) for `I've taken ownership of submission` for `C-2` is sent from A->C
- The sender is responsible for their transactions
- The sender thus will know at the point it sends C-3 to B, that C-2 has reached the point of confirmed delegation on A

... whether C should just package up the information about which confirmed delegations it has on A, up to B.

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+Eventually transactions `A-3` , `B-3`, `C-3` and `D-3` are mined to a block.
+
+![image](../images/distributed_sequencer_switchover_frame8.svg){ width="500" }
+
+Note that there is a brief dip in throughput as node B waits for node A to flush through the pending dispatched transactions and there is also some additional processing for the inflight transactions that haven't been dispatched yet.  So a range size of 4 is unreasonable and it would be more likely for range sizes to be much larger so that these dips in throughput become negligible.
+
+TODO
+
+ - need more detail on precisely _how_  nodes B, C and D know that transactions `B-2``A-2` , `B-2`, `C-2` and `D-2` are past the point of no return and that transactions `A-3` , `B-3`, `C-3` and `D-3` do need to be re-delegated.
+ - need more detail on precisely _how_ node B can continue to coordinate the assembly of transactions `A-3` , `B-3`, `C-3` and `D-3` in a domain context that is aware of the speculative states locks from transactions `B-2``A-2` , `B-2`, `C-2` and `D-2`
```
peterbroadhurst, line 184: If we start to use `domain context` as a specification level thing (vs. just an implementation detail to meet the requirements of the spec) we will need to define what it is and how it works.

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+
+![image](../images/distributed_sequencer_switchover_frame8.svg){ width="500" }
+
+Note that there is a brief dip in throughput as node B waits for node A to flush through the pending dispatched transactions and there is also some additional processing for the inflight transactions that haven't been dispatched yet.  So a range size of 4 is unreasonable and it would be more likely for range sizes to be much larger so that these dips in throughput become negligible.
+
+TODO
+
+ - need more detail on precisely _how_  nodes B, C and D know that transactions `B-2``A-2` , `B-2`, `C-2` and `D-2` are past the point of no return and that transactions `A-3` , `B-3`, `C-3` and `D-3` do need to be re-delegated.
+ - need more detail on precisely _how_ node B can continue to coordinate the assembly of transactions `A-3` , `B-3`, `C-3` and `D-3` in a domain context that is aware of the speculative states locks from transactions `B-2``A-2` , `B-2`, `C-2` and `D-2` 
+ - It might be more productive for the next level of detail on these points to come in the form of a proposal in code.
```
peterbroadhurst, line 185: Some code to experiment with behavior is great... we will need to come back and formally address the behavior of the specification afterwards.

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+
+### Variation in block height
+It is likely that different nodes will become aware of new block heights at different times so the algorithm must accommodate that inconsistency. 
+
+ - given that different nodes index the blockchain events with varying latency, it is not assumed that all nodes have the same awareness of "current block number" at any one time. This is accommodated by the following
+ - Each node delegates all transactions to which ever node it determines as the current coordinator based on its latest knowledge of "current block"
```
peterbroadhurst, line 191:
```suggestion
 - Each node delegates the transactions in its submission list transactions to which ever node it determines as the current coordinator based on its latest knowledge of "current block"
```

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+ - The delegate node will accept the delegation if its awareness of current block also results in it being chosen by the selector function.  Otherwise, the delegate node rejects the delegation and includes its view of current block number in the response
+ - On receiving the delegation rejection, the sender node can determine if it is ahead or behind (in terms of block indexing) the node it had chosen as delegate.  
+ - If the sender node is ahead, it continues to retry the delegation until the delegate node finally catches up and accepts the delegation
+ - If the sender node is behind, it waits until its block indexer catches up and then selects the coordinator for the new range
+ - Coordinator node will continue to coordinate ( send endorsement requests and submit endorsed transactions to base ledger) until its block indexer has reached a block number that causes the coordinator selector to select a different node.
+ - at that time, it waits until all dispatched transactions are confirmed on chain, then delegates all current inflight transactions to the new coordinator.
```
peterbroadhurst, line 197: Some nesting of sub-bullets would help readability here

peterbroadhurst: Also, see comments above on my questions around "delegates all current inflight transactions to the new coordinator" being between coordinators. e.g. in the algorithm as described sometimes its the sender that delegates, and sometimes its another coordinator, and I'm not sure we've worked through that fully yet.

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+ - if the new coordinator is not up to date with block indexing, then it will reject and the delegation will be retried until it catches up. 
+ - while a node is the current selected coordinator, it sends endorsement requests to every other node for every transaction that it is coordinating
```
peterbroadhurst, line 199: This seems like an important part of the algorithm for sure, but I'm not sure why the spec for it is under `variation in block height`

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+ - coordinator becomes unavailable:  If the sender stops receiving heartbeat messages from the coordinator, then it assumes that the coordinator has become unavailable.  However it may actually be the case that the coordinator is still running and just that the network between the coordinator and the sender has a fault.  The action of the sender for each inflight transaction depends on exactly which stage of the flow the sender believes that transaction to be.  The consequence of the sender's action depend on whether the previous coordinator is still operating and where in the flow it believes the transaction to be.  Scenarios to explore are:
+     -  sender detects that the heartbeats have stopped before it has sent a response to a dispatch confirmation request (or before it has received the dispatch confirmation request)
+     -  sender detects that the heartbeats have stopped after it has received  
+  
+
+#### Simple happy path
+
+```mermaid
```
peterbroadhurst, line 248: Very cool 🚀

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+ - if not, then it rejects the endorsement and includes its view of the current block number in the rejection message
+ - when the coordinator receives the rejection message, it can determine if it is ahead or behind the requested endorser
+ - if the coordinator is ahead, it retries the endorsement request until the endorser catches up and eventually endorses the transaction
+ - if the coordinator is behind, then it waits until its block indexer reaches the next range boundary and delegates all inflight transactions to the new coordinator
+
+
+
+### Sender's responsibility
+
+The sender node for any given transaction remains ultimately responsible for ensuring that transaction is successfully confirmed on chain or finalized as failed if it is not possible to complete the processing for any reason.  While the coordination of assembly and endorsement is delegated to another node, the sender continues to monitor the progress and is responsible for initiating retries or re-delegation to other coordinator nodes as appropriate.
```
peterbroadhurst, line 211: > or re-delegation

Again we have the inconsistency here. This paragraph states that re-delegation is the sender's responsibility, but as discussed in comments above we seem to have split this responsibility between sender and coordinator.

As I've mentioned above, I think it would be a simpler to understand spec if it was always the sender's responsibility as this paragraph indicates.

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+
+Feedback available to the sender node that can be monitored to track the progress or otherwise of the transaction submission:
+
+ - when the sender node is choosing the coordinator, it may have recently received a heartbeat message from the preferred coordinator or an alternative coordinator
+ - when sending the delegation request to the coordinator, the sender node expects to receive an acknowledgement that the request has been received.  This is not a guarantee that the transaction will be completed.  At this point, the coordinator has only an in-memory record of that delegated transaction
```
peterbroadhurst, line 216: I'm not clear so far whether if a node has A-1 and A-2, if it is responsible for waiting for the ack of A-1 being delegated, before delegating A-2. This is important in edge cases such as failover, as discussed above

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+ - coordinator heartbeat messages.  The payload of these messages contains a list of transaction IDs that the coordinator is actively coordinating
+ - transaction confirmation request.  Once the coordinator has fulfilled the attestation plan, it sends a message to the transaction sender requesting permission to dispatch. If, for any reason, the sender has already re-delegated to another coordinator, then it will reject this request otherwise, it will accept.
```
peterbroadhurst, line 218: This seems to answer some of my earlier questions, but it also seems inconsistent again. This implies it's the sender that re-delegates (whereas the above diagrams implied some direct old-coordinator <-> new-coordinator comms)

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+ - blockchain event.  The base ledger contract for each domain emits a custom event that included the transaction ID.  The sender node will detect when such an event corresponds to one of its transactions and will record that transaction as confirmed.
+ - transaction reverted message.  When the submitter fails to submit the transaction, for any reason, and gives up trying , then a `TransactionReverted` message is sent to the sender of that transaction.  There are some cases where the submitter retries certain failures and does *not* send this message.
+
+Decisions and actions that need to be taken by the sender node
+
+ - When a user sends a transaction intent (`ptx_sendTransaction` or `ptx_prepareTransaction`), the sender node needs to chose which coordinator to delegate to.
```
peterbroadhurst, line 225: A detail point - but no transaction can be sent to a coordinator, until any referenced dependencies have been confirmed.

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+ - If the block height changes and there is a new preferred coordinator as per the selection algorithm then the sender node needs to decide whether to delegate the transaction to it. This will be dependent on whether the transaction has been dispatched or not.
+ - If the coordinator node seems to have forgotten about the transaction, then the sender node needs to decide to re-delegate it
+ - If the preferred coordinator node becomes unavailable then the sender node needs to decide which alternative coordinator to delegate to
+ - If a transaction has been delegated to an alternative coordinator and the preferred coordinator becomes available again, then the sender needs to decide to re-delegate to the preferred coordinator
```
peterbroadhurst, line 229: Covered in a detail question earlier this exact scenario. It was unclear to me what would cause a sender to find out "the preferred coordinator becomes available again"

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+ - Given that there could have been a number of attempts to delegate and re-delegate the transaction to different coordinators, when any coordinator reaches the point of dispatching the transaction, the sender needs to decide whether or not it is valid to dispatch at the time, by that coordinator
+ - If the base ledger transaction is reverted for any reason, then the sender decides to retry
+
+To illustrate, lets consider the following scenarios
+
+ - simple happy path: sender delegates transaction to a coordinator and that coordinator eventually successfully submits it to the base ledger contract
```
peterbroadhurst, line 235: I think it would be better to provide the textual descriptions along with each diagram.

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+Reply messages also contain
+
+  - `CorrelationID` this is a copy of the `MessageID` of the message that this reply corresponds to.
+
+These can be used by the transport layer plugins to exploit qualities of service capabilities of the chosen transport. 
+
+
+In addition to these, the distributed sequencer protocol defines some common message exchange patterns that provide delivery assurances and correlation in the context of the handshakes described in the protocol walkthrough above.
+
+<!-- TBD: `CorrelationID` is not currently used in the implementation.  Should we remove this or redefine its behavior (i.e. to be a copy of the `RequestID`) or leave it as an option for the transport plugin to use? -->
```
peterbroadhurst, line 733: I think it is important in the spec to define if any messages are "replies" in any way. If they are, then the `CorrelationID` would be used. If not, then it's unnecessary as each message is handled independently.

I note if we can make the specification _not need_ replies, that will be simpler to implement. As soon has you have the concept of an "in-flight" request, you have to have a mechanism to track that in memory and clean it up.

My read of our spec is we're trying to avoid this kind of situation.

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+
+#### Assured delivery
+
+In some cases, it important for one party to have some assurance that the other party has received a message and has triggered an action based on that message.  The action may be to store the message in a persistent store or may be to trigger some in memory process.  
+The approach to achieving assured delivery is to define message exchange patterns where the messages are expected to be reciprocated with some form of acknowledgment.
```
peterbroadhurst, line 738: Think this is the specific case where we need to work through whether there is a `CorrelationID` or not.

peterbroadhurst: Overall I think this section needs some discussion - as I've not been able to correlate it with the rest of the doc.

Our spec above seemed to be of the form:
- Coordinator->Sender: I am ready to do X if you want me to
- Sender->Coordinator: Do X

If either side just forgets state it's fine, it just can't handle the message. But that state isn't "I have message X in flight", it's "I'm ready to do this thing because transaction X is at the right state in my coordinator in-memory state".

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+message DelegationCommand {
+    string transaction_id = 1;
+    string delegate_node_id = 2;
+    string delegation_id = 3; //this is used to correlate the acknowledgement back to the delegation. unlike the transport message id / correlation id, this is not unique across retries
+    bytes private_transaction = 4; //json serialized copy of the in-memory private transaction object
+    int64 block_height = 5; // the block height upon which this delegation was calculated (the highest delegation wins when crossing in the post)
+}
+```
+
+As a member of the coordinator committee, a node that receives a `DelegationCommand` must either respond by sending `DelegationAccepted` or `DelegationRejected` message.
```
peterbroadhurst, line 907: `must either` seems incorrect here. `should either` seems more correct, as it's perfectly expected in our protocol (in my understanding) that a coordinator might just restart and forget in this situation.

peterbroadhurst: This might be wrong though as I think about it - maybe I'm looking for a definition of what we mean by `must`

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+    E@{ shape: diam, label: "this Is<br/>Coordinator" }
+    I@{ shape: lean-r, label: "Reject<br/>(wrong coordinator)" }
+    G@{ shape: fr-rect, label: "<a href="#subprocess-coordinate-transaction">coordinate transaction</a>" }
+    J@{ shape: fr-circ, label: " " }
+
+
+    A-->B
+    B --> |Behind| C
+    B --> |Ahead| C
+    B --> |Same| D
```
peterbroadhurst, line 933: I don't understand yet how this works. I understand the "I've not indexed that block yet" problem, but this algorithm seems to require an **exact** block match, and that mean a lot of thrashing on block boundaries. For example our dev environments come pre-configured with a 100ms (non-empty) block time, so the likelihood under even minimal  load of all nodes being on an identical block is low. With a more traditional 2sec block time, there's definitely a slightly wider window, but I still don't think we should optimize for exact block matching the whole time.

Isn't the _range_ more important?

e.g. the three paths are:
- Agreement: I've indexed up to that block, and I agree on your assessment of who the coordinator is
- Behind: I've not indexed that block yet
- Too far ahead: I've indexed so far ahead, that I think we're in the next range and I'm not the coordinator anymore.

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+When a node receives a heartbeat message from a coordinator node, it should compare that notification with its record of inflight transactions and active delegations
+ - if the receiving node has any in flight transactions that have been delegated to coordinators other than the source of the received heartbeat, then the receiving node should initiate a re-evaluation of its choice of coordinator
+
+```proto
+message HeartbeatNotification {
+    string heartbeat_id = 1;
+    google.protobuf.Timestamp timestamp = 2;
+    string idempotency_key = 3;
+    string contract_address = 4;
+    repeated string transaction_ids = 5;
```
peterbroadhurst, line 1187: This is great. I assume this is only the transaction IDs that have been delegated _by the sender this heartbeat is targeted to_?

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+    D --> E
+    E --> |Yes| G
+    E --> |No| I
+    G --> J
+```
+
+##### <a name="message-transaction-delegation-rejected"></a>Transaction delegation rejected
+
+The handling of a delegation rejected message depends on the reason for rejection.
+ - if the reason is `MismatchedBlockHeight` and the target delegate is ahead then a new delegation request is sent once the sender has reached a compatible block height.  Compatible block height is defined as a block height in the same block range.
```
peterbroadhurst, line 943: Ok - think this is confirmation of my above point - the algorithm above is working to "is compatible block height?" (not "same") semantics.

doc-site/docs/architecture/distributed_sequencer.md
```diff
@@ -0,0 +1,1437 @@
...
+  string preferred_coordinator = 1;
+  google.protobuf.Timestamp latest_heartbeat_time = 2;
+  string latest_heartbeat_id = 3;
+}
+
+```
+
+##### <a name="message-transaction-delegation-accepted"></a>Transaction delegation accepted
+
+If a node receives a `DelegationAccepted` message then it should start to monitor the continued acceptance of that delegation.  It can expect to receive `HeartbeatNotification` messages from the delegate node and for those messages to include the id of the delegated transaction. The `sender` node cannot assume that the `coordinator` node will persist the delegation request.  If the heartbeat messages stop or if the received heartbeat messages do not contain the expected transaction ids, then the sender should retrigger the `HandleTransaction` process to cause the transaction to be re-delegated either to the same delegate, or new delegate or to be coordinated by the sender node itself. Whichever is appropriate for the current point in time.
```
peterbroadhurst, line 978: Digressing to an implementation point here... how do we think code could reasonably be constructed for this?

It seems like a sender node will have:
- A database with zero-or-more (maybe 1million) transactions it wishes to get submitted
- Each transaction will be on the same of different contract address (maybe 1million contract addresses)
- Each contract address has a different coordinator committee
- Each transaction might be in one of a number of states (a state machine diagram would be awesome for this)
    - Initial state - candidate to be delegated to someone
    - Delegated to self - I'm the coordinator so I don't need to healthcheck myself
    - Delegated to remote, but not yet requested submission
    - Delegated to remote, submission requested, but submission not occurred
    - Delegated to remote, submission requested, submission approved, but not yet confirmed
    - Delegated to remote, submission confirmed
         - While in any of these states, we might detect a the block range changing means this is the "wrong" coordinator
    - Done. No need to continue tracking
- The decision on whether an action should be performed for any transaction in this DB could be triggered by:
    - A message from the active coordinator for that txn (we have specs for these starting to finalize in this doc)
    - The _lack of_ a message from the active coordinator fro that txn (the subject of this section)
    - The arrival of a new confirmed block
    - The arrival of an on-chain event (from a confirmed block) confirming a transaction
    - The completion of an earlier transaction that directly (through dependency) or indirectly (through memory-management of "in-fight" transactions for that contract)  unlocks processing

peterbroadhurst: In our code in Paladin we've chosen to have evaluation loops against the DB and in-memory state, working at various levels.  I guess I'm writing this question, because I'm not 100% sure how to correlate those code modules to the state machine(s) implied by this specification that I can see how it fully scales to "mempool management" of a large sea of transactions with efficient DB query and in-memory management.

Potentially this whole question should be moved out of this issue, into a separate one that is about updating our code to be an implementation of the spec.

peterbroadhurst (commented):
Going to pause my review here @hosie because I think there's a lot of outstanding conversations in what I've done so far that is affecting my ability to understand some of the more detailed items like the flow charts.

doc-site/docs/architecture/distributed_sequencer.md
hosie: yes.  I need to add some details generally to explain how explicit dependencies are handled.

doc-site/docs/architecture/distributed_sequencer.md
hosie: I am proposing that it may include other transactions and those should be ignored by the receiver of the message.

doc-site/docs/architecture/distributed_sequencer.md
hosie: The block numbers in the diagram start at 0 therefore block 3 is the last block in the first range.  Early versions of these diagrams had `x+0`,`x+1` etc... where `x` was the block height when the privacy group was created but I felt that was adding noise to the diagrams with no added value so I simplified it to be an absolute numbering and it happened to start at 0.

So when we see that block 3 has been mined, we can assume that any further transactions should come into the second range.  Given that block 0 is the genesis block, it is probably not very helpful for me to be including it in the first range so I'll change it the numbering system to start at 1.

peterbroadhurst: The indexing wasn't the problem. Starting from zero was fine for me - we were referring to the same block.

I think you just spelt "Detection of block height moving to 4" as "receive receipts of block 3".

If instead you had said "confirmation of block 3 being mined, so the next block is now block 4" I'd have been all square.

PR #469, opened 2024-12-05
Zeto support for deposit and withdraw with an ERC20 contract

### LINKED ISSUE #419: Add support for zeto deposit() and withdraw() calls

jimthematrix:
fixes #419 

The Zeto token implementations support interaction with an ERC20 token, to control the value supply publicly. With this paradigm, the token issuer, such as a central bank for digital currencies, can control the total supply in the ERC20 contract. This makes the supply of the tokens public.

The Zeto token contract can be configured to allow balances from a designated ERC20 contract to be "swapped" for Zeto tokens, by calling the `deposit` API. This allows any accounts that have a balance in the ERC20 contract to swap them for Zeto tokens. The exchange rate between the ERC20 and Zeto tokens is 1:1. On successful deposit, the ERC20 balance is transferred to the Zeto contract.

For the `deposit` call the sender must send along a proof that the UTXOs in the output array represent a value that equals the `amount` which will be taken from the sender's eth account in the ERC20 token and moved to the Zeto contract.

In the opposite direction, a user can call `withdraw` to exchange UTXOs they own back to the ERC20 account. For this call the sender must present a proof that the input UTXOs' total value is equal or greater than the `amount`, and the output UTXOs' total value represents the remainder.

Typically in this paradigm, the `mint` API on the Zeto domain should be locked down (disabled) so that the only way to mint Zeto tokens is by depositing.


domains/zeto/build.gradle
```diff
@@ -74,7 +74,7 @@ dependencies {
     coreGo project(path: ":core:go", configuration: "goSource")
 }
 
-task downloadZetoProver {
+task downloadZetoProver(dependsOn: "protoc") {
```
awrichar, line 77: Why do these download tasks depend on protoc?

sdk/typescript/src/domains/zeto.ts
```diff
@@ -1,60 +1,225 @@
...
-            type: "uint256",
-            internalType: "uint256",
-          },
-        ],
-      },
-    ],
-    outputs: [],
-  },
-];
+const zetoPrivateAbi = [{
```
awrichar, line 10: Might consider moving this out to a separate json file.

sdk/typescript/src/domains/zeto.ts
```diff
@@ -156,4 +336,85 @@ export class ZetoInstance {
...
+    this.erc20 = data._erc20;
+    return result;
+  }
+
+  async deposit(from: string, data: ZetoDepositParams) {
+    // first approve the Zeto contract to draw the amount from our balance in the ERC20
+    const txID1 = await this.paladin.sendTransaction({
+      type: TransactionType.PUBLIC,
+      abi: erc20Abi,
+      function: "approve",
```
awrichar, line 362: It feels weird for the Paladin SDK to make an "approve" call to the ERC20. I think this should be handled separately by the user, as we shouldn't make assumptions about their ERC20 contract (only about Paladin contracts).

sdk/typescript/src/domains/zeto.ts
```diff
@@ -156,4 +336,85 @@ export class ZetoInstance {
...
+      to: this.erc20,
+      from,
+      data: { value: data.amount, spender: this.address },
+    });
+    const result1 = await this.paladin.pollForReceipt(txID1, POLL_TIMEOUT_MS);
+    if (result1 === undefined) {
+      throw new Error("Failed to approve transfer");
+    }
+
+    const depositID = await this.paladin.prepareTransaction({
```
awrichar, line 372: Why do we manually prepare and then send, instead of just sending?

domains/zeto/build.gradle
jimthematrix: must be a result of me experimenting with "outputs" but forgot to remove them after deciding to not use it

domains/zeto/integration-test/e2e_test.go
```diff
@@ -314,3 +460,27 @@ func newTestbed(t *testing.T, hdWalletSeed *testbed.UTInitFunction, domains map[
 	rpc := rpcbackend.NewRPCClient(resty.New().SetBaseURL(url))
 	return done, tb, rpc
 }
+
+func getERC20Spec() (*solutils.SolidityBuild, error) {
+	bytes, err := os.ReadFile("./abis/SampleERC20.json")
+	if err != nil {
+		return nil, err
+	}
+	build := solutils.MustLoadBuildResolveLinks(bytes, map[string]*tktypes.EthAddress{})
```
awrichar, line 469: you don't technically need the `ResolveLinks` variant for this contract

domains/zeto/internal/zeto/zeto.go
```diff
@@ -345,6 +371,8 @@ func (z *Zeto) registerEventSignatures(eventAbis abi.ABI) {
 			z.transferSignature = event.SolString()
 		case "UTXOTransferWithEncryptedValues":
 			z.transferWithEncSignature = event.SolString()
+		case "UTXOWithdraw":
+			z.withdrawSignature = event.SolString()
```
awrichar, line 375: Just noting because I had to go look at the Zeto source code... `deposit` just emits the existing `UTXOMint` event, so it doesn't get a new entry here.

sdk/typescript/src/domains/zeto.ts
jimthematrix: at the moment Paladin has a domain-wide setting to either use a designated signer for all the private transactions, or use a hd wallet backed per-tx signing key. Here we need to use the `From` identity to actually sign this particular private transaction, while still allowing the rest of the `transfer` calls to be signed with a one-time key.

jimthematrix: talking with @awrichar further, there a `RequiredSigner` field in the response of the Prepare() domain function, which sounds like could be used to enforce the signer to be the `From` field identity.

domains/zeto/pkg/types/abi.go
```diff
@@ -66,6 +66,27 @@ var ZetoABI = abi.ABI{
 			{Name: "call", Type: "bytes"}, // assumed to be an encoded "transfer"
 		},
 	},
+	{
+		Name: "setERC20",
+		Type: abi.Function,
+		Inputs: abi.ParameterArray{
+			{Name: "_erc20", Type: "address"},
```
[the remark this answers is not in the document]
jimthematrix: making a sweeping change in the Zeto solidity code to name function parameters without the leading `_` but the internal parameters with it. So the assignment statement changes from `erc20 = _erc20` to `_erc20 = erc20`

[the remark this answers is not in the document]
jimthematrix: This will be picked up once we upgrade Paladin to the next Zeto release.

sdk/typescript/src/domains/zeto.ts
jimthematrix: moved it out of the SDK and into the example code

domains/zeto/integration-test/e2e_test.go
```diff
@@ -314,3 +460,27 @@ func newTestbed(t *testing.T, hdWalletSeed *testbed.UTInitFunction, domains map[
 	rpc := rpcbackend.NewRPCClient(resty.New().SetBaseURL(url))
 	return done, tb, rpc
 }
+
+func getERC20Spec() (*solutils.SolidityBuild, error) {
+	bytes, err := os.ReadFile("./abis/SampleERC20.json")
```
awrichar, line 465: Any reason we've done a runtime load here instead of "embed"?

awrichar (approved):

PR #471, opened 2024-12-09
ABI Upload

gabriel-indik:
New ABI upload feature introduced a dialog box where the user can submit ABIs through files or pasting their content:

image


gabriel-indik:
Successful upload confirmation:

image


When there are issues with the ABI:

image

gabriel-indik:
Temporary location in the header until a more complete navigation solution e.g. left nav, is introduced.

image

peterbroadhurst:
I think a drop-down/hamburger menu might be clearer for these things than adding buttons in

gabriel-indik:
@peterbroadhurst thank you for the feedback. I did a number of experiments and the most user-friendly experience resulted in the following design:


https://github.com/user-attachments/assets/50f101a9-1886-4181-a607-213755c46878

Note that in practice the cursor turns into a pointer when hovering the buttons. In the recording above that doesn't show but that's only because of the way the recorder works.

I updated the PR with the code changes to implement the design shown above.

ui/client/src/components/Header.tsx
```diff
@@ -14,24 +14,24 @@
...
+  const { refreshRequired, refresh } = useContext(ApplicationContext);
   const { t } = useTranslation();
   const navigate = useNavigate();
   const pathname = useLocation().pathname.toLowerCase();
   const theme = useTheme();
   const lessThanMedium = useMediaQuery(theme.breakpoints.down("md"));
+  const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
 
   const getTabFromPath = (path: string) => {
     if (path.startsWith('/ui/indexer')) {
```
dechdev, line 37: Might be a good idea to put the routes in a map to prevent typos, and replace the strings in places such as https://github.com/LF-Decentralized-Trust-labs/paladin/blob/main/ui/client/src/App.tsx#L97-L99.

```typescript
export const AppRoutes = {
  Indexer: '/ui/indexer',
  Submissions: 'ui/submissions',
  Submission: (submissionId: string) => `ui/submissions/${submissionId}`,
  ...
}
```

ui/client/src/components/PaladinTransaction.tsx
```diff
@@ -45,10 +45,12 @@ export const PaladinTransaction: React.FC<Props> = ({ paladinTransaction }) => {
   }
 
   const formatProperty = (value: any) => {
```
dechdev, line 47: Is this used anywhere else? Can be a util function if so

ui/client/src/menus/Settings.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  const handleAutoRefreshChange = (value: 'play' | 'pause' | null) => {
+    switch (value) {
+      case 'play': setAutoRefreshEnabled(true); break;
+      case 'pause': setAutoRefreshEnabled(false); break;
+    }
+  };
+
+  const handleColorChange = (mode: 'light' | 'dark' | null) => {
+    console.log(mode !== theme.palette.mode)
```
dechdev, line 50: Remove

ui/client/src/dialogs/ABIUpload.tsx
```diff
@@ -0,0 +1,203 @@
...
+  useEffect(() => {
+    if (dialogOpen) {
+      reset();
+      setRadioSelection('file');
+      setFileSelected(null);
+      setAbiText('');
+    }
+  }, [dialogOpen]);
+
+  const handleSubmit = async () => {
```
dechdev, line 69: See final review comment for more on this section

dechdev (changes requested):
This looks great overall. Since this PR is introducing the first "upload" of the UI, I think we should go ahead and setup Zod to validate all payloads being sent to the backend. 

[Zod Docs](https://zod.dev/?id=basic-usage)
`npm i zod`

Specifically, we can create an ABI zod schema to validate all ABIs that are uploaded to Paladin.

```typescript
const ABIParameterSchema = z.object({
  name: z.string(),
  type: z.string(),
  components: z.array(z.lazy(() => ABIParameterSchema)).optional(),
  internalType: z.string().optional(),
  indexed: z.boolean().optional(),
  baseType: z.string().optional(),
  arrayLength: z.number().optional(),
});

const ABIEntrySchema = z.object({
  type: z.enum(['function', 'constructor', 'event', 'fallback', 'receive']),
  name: z.string().optional(),
  inputs: z.array(ABIParameterSchema).optional(),
  outputs: z.array(ABIParameterSchema).optional(),
  stateMutability: z.enum(['pure', 'view', 'nonpayable', 'payable']).optional(),
  anonymous: z.boolean().optional(),
  constant: z.boolean().optional(),
  payable: z.boolean().optional(),
});

const ABISchema = z.array(ABIEntrySchema);
```

Then, you can validate ABIs:

```typescript
const validABI = ABISchema.parse(abiObject);
if (!validABI.success) {
  throw new Error('Invalid ABI');
}
// This will have the same abi object as the original, but with the types validated
mutate(validABI.data)
```

ui/client/src/components/Header.tsx
gabriel-indik: Great idea, will move those to constants defined in a separate file.

ui/client/src/components/PaladinTransaction.tsx
gabriel-indik: Good point, this is a good candidate to be moved into an utilities file. I also took the opportunity to rename it and adjust the formatting, so now JSON structures are easier to read:

image

ui/client/src/menus/Settings.tsx
gabriel-indik: Good catch!

ui/client/src/dialogs/ABIUpload.tsx
gabriel-indik: This is a very interesting idea. I like being able to validate the ABI on the client before it is submitted to the Paladin node. @dechdev provided another interesting source for getting a comprehensive Zod based validator for ABI (https://github.com/wevm/abitype/tree/main/packages/abitype) which we could use. I started experimenting with it and built a prototype. It works exactly as expected: valid ABIs pass, and invalid throw errors, but exploring the error details can lead to long complex structures. Here is an example of a small ABI (simple storage) to which I introduced 1 small typo (input -> nput)

```json
ZodError: [
  {
    "code": "invalid_union",
    "unionErrors": [
      {
        "issues": [
          {
            "received": "constructor",
            "code": "invalid_literal",
            "expected": "error",
            "path": [
              0,
              "type"
            ],
            "message": "Invalid literal value, expected \"error\""
          },
          {
            "code": "invalid_type",
            "expected": "array",
            "received": "undefined",
            "path": [
              0,
              "inputs"
            ],
            "message": "Required"
          },
          {
            "code": "invalid_type",
            "expected": "string",
            "received": "undefined",
            "path": [
              0,
              "name"
            ],
            "message": "Required"
          }
        ],
        "name": "ZodError"
      },
      {
        "issues": [
          {
            "received": "constructor",
            "code": "invalid_literal",
            "expected": "event",
            "path": [
              0,
              "type"
            ],
            "message": "Invalid literal value, expected \"event\""
          },
          {
            "code": "invalid_type",
            "expected": "array",
            "received": "undefined",
            "path": [
              0,
              "inputs"
            ],
            "message": "Required"
          },
          {
            "code": "invalid_type",
            "expected": "string",
            "received": "undefined",
            "path": [
              0,
              "name"
            ],
            "message": "Required"
          }
        ],
        "name": "ZodError"
      },
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "array",
            "received": "undefined",
            "path": [
              0,
              "inputs"
            ],
            "message": "Required"
          }
        ],
        "name": "ZodError"
      }
    ],
    "path": [
      0
    ],
    "message": "Invalid input"
  }
]
```

It looks like if there is an error present, and its location is relatively hight in the object hierarchy within the ABI, that can cascase into many secondary errors.

I think this functionaltiy merits its own separate issue where we can investigate more in detail how to get the most out of Zod in order to improve user experience when it comes to raising problems in ABIs.

dechdev: For error validation, this is the recommended package from the Zod team: https://github.com/causaly/zod-validation-error

Reference: https://zod.dev/?id=error-handling

dechdev (approved):

ui/client/src/dialogs/ABIUpload.tsx
gabriel-indik: Right, that's the approach I was using. In the examples that are shown in the Zod documentation, validation is clear for the simple flat-objects used. However, When using an Abi combined with the schema from AbiType, the errors that Zod list for a single type can be difficult to understand:

image

dwertent (approved):
Relying on @dechdev approval

Issue #476, opened 2024-12-10
bug: when a Pente transaction is reverted during assembly, Pente stops working


awrichar:
### What happened?

I submitted an invalid private transaction to Pente. It was reverted (as expected) during the assembly phase. Now I cannot submit the transaction again - I never receive a receipt. Transaction `94ac5cb8-d906-4860-9b4b-0029a399fa86` in these logs is an example of a transaction submitted _after_ a revert (so this transaction was frozen in limbo).

### What did you expect to happen?

My transaction should always succeed or fail - it should not get stuck.

### How can we reproduce it (as minimally and precisely as possible)?

[paladin1.log](https://github.com/user-attachments/files/18086534/paladin1.log)

### OS version

Mac / Sequoia

awrichar:
I believe this may be a pointer to the issue:

```
[2024-12-10T22:08:30.927Z] DEBUG AssembleCoordinator:waitForDone aa0c46bc-2899-4452-b40a-b84bddf96af8 pid=1
[2024-12-10T22:08:31.928Z] ERROR AssembleCoordinator:waitForDone request timeout for request aa0c46bc-2899-4452-b40a-b84bddf96af8 pid=1
```

This request ID `aa0c46bc-2899-4452-b40a-b84bddf96af8` seems to be related to transaction `3e5b3e01-d17e-4b14-a131-e52cbfaedad3`, which is the previous transaction that reverted in the private contract.

```
[2024-12-10T22:08:32.257Z]  INFO Inserting receipt txId=3e5b3e01-d17e-4b14-a131-e52cbfaedad3 success=false failure=PD012214: Unable to decode revert data (no revert data available) txHash=0x07573d823c7cb4015f9dbaaa7e59ec796a11a85d5f14f18995b1c7cdf2998205 pid=1 role=block_indexer
```

From what I can tell, this revert:
1. Does not directly trigger the waiting code to know that assembly has failed
2. In fact the enclosing code seems to time out _before_ the revert happens, indicating possibly some mismatched timeouts here
3. The code seems to hang forever in `waitForDone` [in this case](https://github.com/LF-Decentralized-Trust-labs/paladin/blob/f4a838b8f709a39fd59d34e35661aa6e81402de9/core/go/internal/privatetxnmgr/assemble_coordinator.go#L135-L137) because nothing causes it to break out of the infinite `for`

awrichar:
[...]

PR #479, opened 2024-12-13
Update examples and run during e2e build

awrichar:
Add a `PaladinVerifier` type to the SDK for storing and resolving Paladin identities.

Update the examples to use this new type, and run all of the examples as part of `gradle e2e` (to help ensure the examples stay functional).

PR #481, opened 2024-12-16
free up assembly coordinator after revert

hosie:
The assembly coordinator will not allow the next transaction to be assembled until it receives a `Complete` call from the sequencer. Normally this happens once it has received the completion event for a previous transaction  but this was missing from the code branch that handles `revert` result.


peterbroadhurst (approved):
I'll go ahead and merge this, then pull it into #472 (which pulls in #479) to run the E2E.

peterbroadhurst:
Fixes #476

PR #485, opened 2024-12-19
Exploration on contracts for NotoDepositWithdraw

awrichar:
This is a proposal for what a smart contract would look like to support a withdraw/deposit pattern on Noto. Specifically, the pattern implemented here allows withdrawing some Noto value as an ERC20 and later re-depositing it as Noto.

- Each withdrawal is a separate ERC20 token
- Each withdrawal is also tracked as a UTXO state, owned by the withdrawer
- ERC20 tokens may be moved around as normal without affecting UTXO states
- ERC20 token owners may deposit ERC20 tokens back to Noto, which will spend+confirm UTXO states
- While ERC20 tokens are in the clear, parties may use randomly chosen keys for pseudonymity

Pros:
- Allows an infinite range of programmability by temporarily converting Noto to ERC20

Cons:
- Some loss of confidentiality
- Some nuance in representing ownership of withdrawn tokens until they are deposited again

This is not yet a formal proposal for inclusion in Paladin, but more of a discussion-via-code to determine if this is a useful pattern to pursue, or if it breaks too many features around privacy.

Also, because Noto already knows how to talk to Pente, it's possible that a _private_ ERC20 could be used instead of one on the shared ledger - although I haven't mentally worked through the details of this. It would require a different set of methods on the Noto contract.

### DIFF
--- a/solidity/contracts/domains/noto/NotoDepositWithdraw.sol
+++ b/solidity/contracts/domains/noto/NotoDepositWithdraw.sol
@@ -1,0 +1,141 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.20;
+
+import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import {Noto} from "./Noto.sol";
+import {NotoERC20, NotoERC20Factory} from "./NotoERC20.sol";
+
+/**
+ * @dev Noto variant that allows withdrawing Noto into an ephemeral ERC20
+ *      token, then later depositing back as Noto.
+ */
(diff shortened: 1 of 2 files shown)


awrichar:
After having some discussion on where this landed, I've decided it's not actually deposit/withdraw. It's similar, but that pattern is a bit more about controlling supply and providing an alternative for mint/burn operations. 

This has actually ended up being a form of locking and unlocking, using ERC20 for programmable delegation of locked amounts. It feels like it actually needs to end up as an extension of locking (possibly the "hooks" on that feature should support this pattern as a way that you could handle custom locks).

PR #490, opened 2024-12-20
Add new domain callbacks

awrichar:
New callbacks that can be triggered by domains:
- **SendTransaction** - directly trigger a new transaction, such as in response to an event.
- **LocalNodeName** - access the name of the local node. Noto uses this to determine if the current node houses the notary identity.
- **GetStates** - look up states by ID, regardless of whether they are available.


core/go/internal/statemgr/domain_context.go
```diff
@@ -249,6 +249,15 @@ func (dc *domainContext) mergeInMemoryMatches(schema components.Schema, states [
 
 }
 
+func (dc *domainContext) GetStates(dbTX *gorm.DB, schemaID tktypes.Bytes32, ids []string) (components.Schema, []*pldapi.State, error) {
+	idsAny := make([]any, len(ids))
+	for i, id := range ids {
+		idsAny[i] = id
+	}
+	query := query.NewQueryBuilder().In(".id", idsAny).Query()
+	return dc.ss.findStates(dc, dbTX, dc.domainName, &dc.contractAddress, schemaID, query, pldapi.StateStatusAll)
```
awrichar, line 258: Note: I have not done any merging of unflushed states, so this will only find states that have been flushed to the database. Might need to add "confirmed and available but unflushed" states into the results for completeness.

core/go/internal/domainmgr/domain.go
```diff
@@ -771,3 +775,55 @@ func (d *domain) BuildDomainReceipt(ctx context.Context, dbTX *gorm.DB, txID uui
...
+	contractAddress, err := tktypes.ParseEthAddress(tx.Transaction.ContractAddress)
+	if err != nil {
+		return nil, err
+	}
+	var functionABI abi.Entry
+	if err = json.Unmarshal([]byte(tx.Transaction.FunctionAbiJson), &functionABI); err != nil {
+		return nil, err
+	}
+
+	id, err := d.dm.txManager.SendTransaction(ctx, &pldapi.TransactionInput{
```
peterbroadhurst, line 793: I think this needs to propagate the DB transaction from the DomainContext.

peterbroadhurst: I note that will require the `SendTransaction` code to be updated, as currently it opens its own DB TX here:
https://github.com/LF-Decentralized-Trust-labs/paladin/blob/ec6eedceda5b58b1d582e3d7d5ab932e4ea8f542/core/go/internal/txmgr/transaction_submission.go#L381-L386

core/go/internal/domainmgr/domain.go
```diff
@@ -771,3 +775,55 @@ func (d *domain) BuildDomainReceipt(ctx context.Context, dbTX *gorm.DB, txID uui
...
+	return &prototk.SendTransactionResponse{Id: id.String()}, nil
+}
+
+func (d *domain) LocalNodeName(ctx context.Context, req *prototk.LocalNodeNameRequest) (*prototk.LocalNodeNameResponse, error) {
+	return &prototk.LocalNodeNameResponse{
+		Name: d.dm.transportMgr.LocalNodeName(),
+	}, nil
+}
+
+func (d *domain) GetStates(ctx context.Context, req *prototk.GetStatesRequest) (*prototk.GetStatesResponse, error) {
```
peterbroadhurst, line 814: I believe this (as implemented) is `GetConfirmedStatesByID`.

For some reason we pass it to the state context, but it doesn't use the state context. Would it make more sense to just call `statestore.FindStates(ctx, ss.p.DB(), domain, schema, &query, status)` directly?

peterbroadhurst (commented):
Couple of questions @awrichar to discuss

core/go/internal/domainmgr/domain.go
awrichar: Yea, agreed. So the current behavior for `SendTransaction` is that it will always create 1) a database transaction, and 2) a key resolution context. Obviously we now need a version that allows _passing in_ a database transaction. I think the simplest change to the interface would therefore be:
- add `dbTX *gorm.DB` to all of the send/prepare method signatures, so that the caller is _always_ responsible for opening the database transaction
- still cause txmgr to create and dispose of a key resolution context, using the passed in database transaction

I could pick lots of other solutions - such as having multiple signatures (versions that take in a database transaction, and versions that do not). I could require the caller to pass in a key resolution context, or provide multiple signatures where it may be passed in or created on demand.

If you have an immediate steer based on knowing this code, let me know... otherwise I'll dive in and start testing the various options.

peterbroadhurst: Having the `txmgr` external methods always take in a `*gorm.DB` _and a_ `components. KeyResolver` if needed, seems good to me.

core/go/internal/domainmgr/domain.go
awrichar: I believe what I want is `GetStatesByID`. I want to be able to look up the private data for a state, and depending on timing, the state may be proposed, confirmed, or spent by the time I perform the lookup.

core/go/internal/domainmgr/domain.go
peterbroadhurst: I've helped work through this one.

core/go/internal/domainmgr/domain.go
peterbroadhurst: I've gone ahead and:
- Renamed to `GetStatesByID`
- Made it return un-flushed states from the Domain Context as well as anything in the DB

core/go/internal/statemgr/domain_context.go
peterbroadhurst: This is done now

core/go/internal/components/statemgr.go
```diff
@@ -96,6 +96,12 @@ type DomainContext interface {
 	// The dbTX is passed in to allow re-use of a connection during read operations.
 	FindAvailableStates(dbTX *gorm.DB, schemaID tktypes.Bytes32, query *query.QueryJSON) (Schema, []*pldapi.State, error)
 
+	// GetStates retrieves a set of states by ID - regardless of whether they are:
```
awrichar, line 99: nit: name is now GetStatesByID

core/go/internal/statemgr/domain_context.go
```diff
@@ -249,6 +258,26 @@ func (dc *domainContext) mergeInMemoryMatches(schema components.Schema, states [
...
+func (dc *domainContext) GetStatesByID(dbTX *gorm.DB, schemaID tktypes.Bytes32, ids []string) (components.Schema, []*pldapi.State, error) {
+	idsAny := make([]any, len(ids))
+	for i, id := range ids {
+		idsAny[i] = id
+	}
+	query := query.NewQueryBuilder().In(".id", idsAny).Sort(".created").Query()
+	schema, matches, err := dc.ss.findStates(dc, dbTX, dc.domainName, &dc.contractAddress, schemaID, query, pldapi.StateStatusAll)
+	if err == nil {
+		var memMatches []*components.StateWithLabels
+		memMatches, err = dc.mergeUnFlushed(schema, matches, query, false)
```
awrichar, line 270: It looks like this function still filters out spent states - so there's a small chance we'll miss states that were created _and_ spent in memory before being flushed to the database.

awrichar (commented):
@peterbroadhurst thanks for cleaning up the remaining items on the domain interface. I left a couple of comments back, but overall it looks great to me.

core/go/internal/statemgr/domain_context.go
peterbroadhurst: Good spot - I've updated so it does not exclude those locked-for-spend states

peterbroadhurst (approved):

peterbroadhurst:
See this in the docker build failure:

```
#38 11.44 * What went wrong:
#38 11.44 Could not determine the dependencies of task ':core:java:jar'.
#38 11.44 > Could not resolve all files for configuration ':core:java:runtimeClasspath'.
#38 11.44    > Could not find any version that matches com.google.protobuf:protobuf-java-util:3.25+.
#38 11.44      Versions that do not match:
#38 11.44        - 4.29.3
#38 11.44        - 4.29.2
#38 11.44        - 4.29.1
#38 11.44        - 4.29.0
#38 11.44        - 4.29.0-RC3
#38 11.44        - + 25 more
#38 11.44      Searched in the following locations:
#38 11.44        - https://repo.maven.apache.org/maven2/com/google/protobuf/protobuf-java-util/maven-metadata.xml
#38 11.44        - https://repo.maven.apache.org/maven2/com/google/protobuf/protobuf-java-util/3.25.6/protobuf-java-util-3.25.6.pom
#38 11.44      Required by:
#38 11.44          project :core:java
#38 11.44          project :core:java > project :toolkit_java
```

Issue #488, opened 2024-12-20
Peer management enhancements


peterbroadhurst:
### What would you like to be added?

- Better thread management of state distribution backlog and retries
- Status information to users for "active peers"
- Clear code responsibility for managing the node-to-node transport definitions

### Why is this needed?

See: https://github.com/LF-Decentralized-Trust-labs/paladin/blob/a06db216c29d9b238da26984fca491a446daf96a/core/go/internal/statedistribution/state_distributer.go#L115-L116

peterbroadhurst:
Initial architecture plan:

- The responsibility is combined into `transportmgr` from `statedistribution` (which gets removed)
- There is an in-memory status + go-routine for every "active peer"
- This go-routine is responsible for completing state distributions
- This go-routine also processes requests to send things from a dedicated go channel to that peer
- `transportmgr` takes ownership of the payload spec for messages
- Only spec-defined messages can be sent/received
- A short in-memory retry is done by transport manager to handle very intermittent errors to minimize complexity in transport plugins
- Messages are still one-way fire&forget, but these short retries manage very short reconnect-style error scenarios
- Status (bytes sent, lifetime of active, etc.) on failures is available over JSON/RPC on `transport_peers` / `transport_peerInfo`
![image](https://github.com/user-attachments/assets/743c8004-9bf6-4a9c-b58f-c42cd7fae59c)

peterbroadhurst:
> Initial architecture plan:
> 
> * The responsibility is combined into `transportmgr` from `statedistribution` (which gets removed)
> * There is an in-memory status + go-routine for every "active peer"
> * This go-routine is responsible for completing state distributions
> * This go-routine also processes requests to send things from a dedicated go channel to that peer
> * `transportmgr` takes ownership of the payload spec for messages
> * Only spec-defined messages can be sent/received
> * A short in-memory retry is done by transport manager to handle very intermittent errors to minimize complexity in transport plugins
> * Messages are still one-way fire&forget, but these short retries manage very short reconnect-style error scenarios
> * Status (bytes sent, lifetime of active, etc.) on failures is available over JSON/RPC on `transport_peers` / `transport_peerInfo`
[...]

PR #491, opened 2024-12-23
Peer Management Enhancements (issue #488)

peterbroadhurst:
Implements #488 

- [x] Update transport manager with new responsibilities
    - [x] Reliable message DB tables
    - [x] Peer management / activation
    - [x] Peer inactivity reaping / deactivation
    - [x] Fire+forget send short retry with Go channels
    - [x] Reliable send from local DB tables, using `sequence` to track unsent messages
    - [x] Ack/Nack sending
    - [x] Ack/Nack storage
    - [x] Move `protox.Message` to `protox.PaladinMsg` to remove some confusion in the code with gRPC `Message`
- [x] Reconciling private TX manager data duplicate data structures and redundant code
    - [x] Reconcile `components.TransportMessage` and `protox.PaladinMsg`
    - [x] Move list of `Components` to `protox.PaladinMsg` `enum` called `Component`
    - [x] **Change** export/import for domain context to include un-written state data
       - [x] Remove state distribution from `Complete()` on assemble, as states should flow in `endorse` payload
    - [x] Update private TX manager to call new `TransportManager` APIs
- [x] Assembling / storing data types delivered via reliable message types
    - [x] `StateDistribution`
        - Payload contains the domain/schema/id/nullifier info
        - State data itself is **not duplicated** in the DB, and only stored in the `StateManager`
    - [x] `Receipt`
    - [x] `PreparedTransaction`
- [x] JSON/RPC interface
    - [x] `transport_peers`
    - [x] `transport_peerInfo`
    - [x] Add struct docs (and run `gradle generateDocs`)
    - [x] Including statistics in the `pldapi.PeerInfo` structure
https://github.com/LF-Decentralized-Trust-labs/paladin/blob/2b62d3999047e32cda68be624d2d66bd496f9557/toolkit/go/pkg/pldapi/peerinfo.go#L21-L38


toolkit/go/pkg/plugintk/plugin_type_transport.go
```diff
@@ -28,6 +28,8 @@ type TransportAPI interface {
 	ConfigureTransport(context.Context, *prototk.ConfigureTransportRequest) (*prototk.ConfigureTransportResponse, error)
 	SendMessage(context.Context, *prototk.SendMessageRequest) (*prototk.SendMessageResponse, error)
 	GetLocalDetails(context.Context, *prototk.GetLocalDetailsRequest) (*prototk.GetLocalDetailsResponse, error)
+	ActivateNode(context.Context, *prototk.ActivateNodeRequest) (*prototk.ActivateNodeResponse, error)
```
hosie, line 31: Minor spelling observation.  `ActivateNode` could be misleading esp. if seen out of context. `ActivatePeering` might be more accurate.

hosie (approved):
LGTM.  A couple of minor comments but on the whole, this is a great improvement.

toolkit/go/pkg/plugintk/plugin_type_transport.go
peterbroadhurst: Good point - went for `ActivatePeer`/`DeactivatePeer`

peterbroadhurst:
Merging is dependent on #498, because of a small tweak to the `prepare` behavior in this PR - it now only sends the prepared transaction to the address designated as the `from` in that prepared transaction.

The E2E test / tutorial in `example/bond` is affected by this, but not after the refactor that was already in plan.

@awrichar and I discussed whether we should remove prepared transaction entirely in this PR, as they no longer have an active role, but we decided to decouple that change.

peterbroadhurst:
Hit intermittent build failure https://github.com/LF-Decentralized-Trust-labs/paladin/issues/410 - re-running

peterbroadhurst:
Thanks @awrichar for addressing the E2E

PR #496, opened 2025-01-02
Reduce testbed logging to stdout during build

awrichar:
This should greatly clean up CI build output to make test failures easier to find.

- Dump docker compose logs to a file instead of stdout, and collect file in GitHub build artifacts
- Set testbed logging to file earlier (previously the logger was initialized to stdout and then later redirected, but could still result in thousands of lines going to stdout)
- Fix testbed package tests to log to a file, just like other component tests that are built on top of the testbed


peterbroadhurst (approved):
✨

PR #494, opened 2024-12-31
zeto: update to v0.0.11 and fix lock functionality

awrichar:
Caveats: does not yet work with nullifier variants, and needs additional unit test coverage. However, locking behavior is likely to change further based on pending discussions in https://github.com/hyperledger-labs/zeto/issues/119.

PR #498, opened 2025-01-04
Noto/Zeto atomic swap example

awrichar:
In a chain with #483 

This example shows how to perform an atomic swap between Zeto and Noto (with Pente notary hooks), with full counterparty protection. It leverages locking functionality on both tokens to set aside the value for the swap, then approve a single base ledger `Atom` contract to finalize the swap. This ensures neither party can receive the other's tokens unless the full swap is executed.

Some caveats:
- Noto locking functionality is still under review in #483 and could change further - but seeing the steps in this flow may also help to validate or suggest needed changes to that logic.
- Zeto lock is not yet supported for nullifier variants, as the signing module code isn't yet added to the domain (noted in #494).
- Zeto locks do not currently prevent Paladin from trying to spend the state - so additional transfers created while a lock is in effect may try to (unsuccessfully) spend states that are locked. Ongoing discussion about locking and difficulties with indexing is at https://github.com/hyperledger-labs/zeto/issues/119.

This example may eventually need to be accompanied by a tutorial in the docs - but until the above caveats are addressed, it feels premature. Also, the fact that Zeto still relies on `prepareTransaction` is potentially problematic (as I really think we should remove this piece of functionality).

---

This also updates domain receipts with additional information, including a parsed list of transfers.


awrichar:
Test failure is https://github.com/LF-Decentralized-Trust-labs/paladin/issues/185

PR #499, opened 2025-01-06
Reconcile race

dwertent:


Dockerfile
```diff
@@ -185,8 +185,12 @@ ENV PATH=$PATH:/usr/local/java/bin
 # Define the entry point for running the application
 ENTRYPOINT [                         \
     "java",                          \
+    "--add-opens", "java.base/jdk.internal.misc=ALL-UNNAMED", \
```
dwertent, line 188: This is to prevent the warning logs

PR #501, opened 2025-01-06
500: provide feedback for invalid ABI

gabriel-indik:
Ensure feedback is provided when the ABI provided is invalid even though the JSON is (syntactically) valid.


gabriel-indik:
Feedback provided in the UI when the JSON is syntactically correct but is not a valid ABI:

image

dwertent (approved):

PR #505, opened 2025-01-07
Fix bugs when processing transactions that have info states and no output states

### LINKED ISSUE #502: bug: transaction with no input/output states was not delegated properly
### What happened?

I was testing some changes to the Noto domain, part of which involved building a transaction that had only "read" and "info" states, and no "input" or "output" states. The transaction flow seemed to get stuck when it should have delegated to the notary. While I saw the first few steps of the transaction flow happen on node 2 (the sender node), I saw no record at all of the transaction on node 1 (the notary node). When I changed the flow to have "inputs" and "outputs" after assembly, it seemed to work.

Transactions should always be delegated and submitted properly, regardless of what types of states they contain.

See attached logs. Transaction was `046a84d1-bf4e-4ab2-8fa7-74844f4a3306`.

[paladin1.log](https://github.com/user-attachments/files/18325669/paladin1.log)
[paladin2.log](https://github.com/user-attachments/files/18325670/paladin2.log)
[paladin3.log](https://github.com/user-attachments/files/18325671/paladin3.log)

I'm not positive that this is related to read/info states, but that seems to be the only obvious difference in the flows that worked and those that did not.

### OS version

_No response_

awrichar:
Fixes #502

### DIFF
--- a/core/go/internal/domainmgr/event_indexer.go
+++ b/core/go/internal/domainmgr/event_indexer.go
@@ -357,7 +357,7 @@ func (d *domain) handleEventBatchForContract(ctx context.Context, dbTX *gorm.DB,
 	}
 
 	// Then any finalizations of those states
-	if len(stateSpends) > 0 || len(stateReads) > 0 || len(stateConfirms) > 0 {
+	if len(stateSpends) > 0 || len(stateReads) > 0 || len(stateConfirms) > 0 || len(stateInfoRecords) > 0 {
 		if err := d.dm.stateStore.WriteStateFinalizations(ctx, dbTX, stateSpends, stateReads, stateConfirms, stateInfoRecords); err != nil {
 			return nil, err
 		}
--- a/core/go/internal/privatetxnmgr/transaction_flow_actions.go
+++ b/core/go/internal/privatetxnmgr/transaction_flow_actions.go
@@ -370,7 +370,8 @@ func (tf *transactionFlow) writeAndLockStates(ctx context.Context) {
 	// we are accessing the transactionFlow's PrivateTransaction object which is only safe to do on the sequencer thread
 	// but we need to make sure that these writes/locks are complete before the assemble requester thread can proceed to assemble
 	// the next transaction
-	if tf.transaction.PostAssembly.OutputStatesPotential != nil && tf.transaction.PostAssembly.OutputStates == nil {
+	if (tf.transaction.PostAssembly.OutputStatesPotential != nil && tf.transaction.PostAssembly.OutputStates == nil) ||
+		(tf.transaction.PostAssembly.InfoStatesPotential != nil && tf.transaction.PostAssembly.InfoStates == nil) {
 		// We need to write the potential states to the domain before we can sign or endorse the transaction
 		// but there is no point in doing that until we are sure that the transaction is going to be coordinated locally
 		// so this is the earliest, and latest, point in the flow that we can do this


core/go/internal/domainmgr/event_indexer.go
```diff
@@ -357,7 +357,7 @@ func (d *domain) handleEventBatchForContract(ctx context.Context, dbTX *gorm.DB,
 	}
 
 	// Then any finalizations of those states
-	if len(stateSpends) > 0 || len(stateReads) > 0 || len(stateConfirms) > 0 {
+	if len(stateSpends) > 0 || len(stateReads) > 0 || len(stateConfirms) > 0 || len(stateInfoRecords) > 0 {
```
peterbroadhurst, line 360: ok - this makes sense

So @awrichar this is saying that you can have a transaction that only includes `info`, and still results in a state receipt, as long as the info states are still emitted from the on-chain smart contract.

Does that fit your use case ok?

core/go/internal/privatetxnmgr/transaction_flow_actions.go
```diff
@@ -370,7 +370,8 @@ func (tf *transactionFlow) writeAndLockStates(ctx context.Context) {
 	// we are accessing the transactionFlow's PrivateTransaction object which is only safe to do on the sequencer thread
 	// but we need to make sure that these writes/locks are complete before the assemble requester thread can proceed to assemble
 	// the next transaction
-	if tf.transaction.PostAssembly.OutputStatesPotential != nil && tf.transaction.PostAssembly.OutputStates == nil {
+	if (tf.transaction.PostAssembly.OutputStatesPotential != nil && tf.transaction.PostAssembly.OutputStates == nil) ||
```
peterbroadhurst, line 373: Spending a bit of time working out where this function fits into the transaction flow...

peterbroadhurst: Ok. This is adding the existence of states into the domain context. But these `info` states should never become available. So I'm not yet sure we're right here...

peterbroadhurst: No - it's all good. That critical difference between the two types of state (output states that become available, vs. info states that just need storing/distributing) is covered in the called function:
https://github.com/LF-Decentralized-Trust-labs/paladin/blob/5290e466f55c29848d2b28ec5165693c5ed77cfe/core/go/internal/domainmgr/private_smart_contract.go#L288

peterbroadhurst (approved):
Thanks @awrichar - great work on diagnosing these.

core/go/internal/domainmgr/event_indexer.go
awrichar: Yes, exactly.

FWIW I don't think it's where I've actually ended up in my use case at the moment - I believe my transactions ended up having at least one other type of state, and not only info states. But fundamentally I don't see why we would _ignore_ the domain telling us to confirm info states, simply because other states weren't present.

core/go/internal/privatetxnmgr/transaction_flow_actions.go
awrichar: Yes, this block is storing the state data for new states proposed in this assembled transaction. That should include both output states and info states.

Overall I'm a bit uncomfortable with the transaction flow here, where we iteratively look for `nil` things that we think we can populate. This fix has resolved a later panic as detailed in #502 because of some assumptions about what will and won't be `nil` at various points in the flow - and it's very difficult to follow how this check leads to that one.

It feels like there may be many more edge cases like this lurking around. I've not taken on a larger refactor at the moment, but it feels like one may be needed at some point.

PR #507, opened 2025-01-08
Update docs

gabriel-indik:
The QueryJSON section in the docs contained examples that are not valid. Specifically, nested value objects. This PR addresses that issue and in addition it incorporates the available fields and their types to various queries in the APIs section.
These are included in the form of a table. Example:

image


dwertent (changes requested):
It seems the documentation here is [generated](https://github.com/LF-Decentralized-Trust-labs/paladin/blob/7d67e2812ddc4cdd2ce019e126bb78b3228d17ba/toolkit/go/pkg/reference/reference.go#L380), so I’m wondering if it’s necessary to add anything manually. 

Regarding the fields, it looks like the, for example, the `bidx_queryIndexedBlocks` [function receives the query object](https://github.com/LF-Decentralized-Trust-labs/paladin/blob/7d67e2812ddc4cdd2ce019e126bb78b3228d17ba/toolkit/go/pkg/pldclient/blockindex.go#L112) generated by the script. Could you help me understand why specific fields are being written for the query? Are there any restrictions from the API that require this?

For the changes in `queryjson.md`, it seems there might be a small issue with how the docs are being generated. Would it make sense to address this through the docs generator script instead of making manual updates?

gabriel-indik:
@dwertent I'll look into the generation portion you pointed out. With regards to the fields and their types, that's exactly where the documentation was falling short. The query it accepts will support only a specific set of fields with a specific set of types per API call.

gabriel-indik:
@dwertent thank you for the heads up, you are absolutely right, that portion of the docs is generated so making the changes to the markdown is not the right approach. I modified the Go code that generates the file that contained the main issues (wrong nesting of the values for the queries). When it comes to adding the tables, that will likely require some additional work. I was looking at how reflection is being used to extract all the information into the docs (which is brilliant) but for the case of the specific fields available in each query (and their types) those are defined as variables in other packages and not directly referenced by the methods. So this will require a bit more investigation. I will be creating a separate issue for those.

PR #517, opened 2025-01-16
Add resource configuration to Paladin, Postgres, and Besu controllers

dwertent:
Introduce resource configuration options for Paladin, Postgres, and Besu controllers to enhance resource management and allocation.

### DIFF
--- a/operator/charts/paladin-operator/assets/config.json
+++ b/operator/charts/paladin-operator/assets/config.json
@@ -8,7 +8,8 @@
     "tolerations": {{ toJson .Values.paladin.tolerations }},
     "affinity": {{ toJson .Values.paladin.affinity }},
     "nodeSelector": {{ toJson .Values.paladin.nodeSelector }},
-    "securityContext": {{ toJson .Values.paladin.securityContext }}
+    "securityContext": {{ toJson .Values.paladin.securityContext }},
+    "resources": {{ toJson .Values.paladin.resources }}
   },
   "besu": {
     "image": "{{ .Values.besu.image.repository }}:{{ .Values.besu.image.tag }}",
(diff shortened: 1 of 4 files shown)


peterbroadhurst (commented):
I think I need to understand this better @dwertent 

We're putting resource specifications (like CPU/Memory requests/limits) into the config file of the operator.

- Are these just the defaults?
- Are there a default set of defaults? (I don't see any checked in with this change)
- Can they be overridden on a per-CR basis still?

dwertent:
> Are there a default set of defaults? (I don't see any checked in with this change) 

Currently, we don’t have default CPU/memory requests and limits. I believe Sarah was testing them, but they aren’t included in this change.  

> Can they be overridden on a per-CR basis still?

The Paladin CR focuses on Paladin’s application-level configuration, whereas this ConfigMap handles operator-specific configurations. For example, to set Paladin’s image, you configure it in the operator’s ConfigMap, not the Paladin CR. That’s why these changes are made in the ConfigMap. When the ConfigMap is updated, the operator restarts (hence #561) and applies the new configuration to the Paladin, Besu, and Postgres nodes accordingly.

onelapahead:
> The Paladin CR focuses on Paladin’s application-level configuration, whereas this ConfigMap handles operator-specific configurations. For example, to set Paladin’s image, you configure it in the operator’s ConfigMap, not the Paladin CR. That’s why these changes are made in the ConfigMap. When the ConfigMap is updated, the operator restarts (hence https://github.com/LF-Decentralized-Trust-labs/paladin/pull/561) and applies the new configuration to the Paladin, Besu, and Postgres nodes accordingly.

So this is tricky - I could definitely see the case for wanting Paladin's of different sizes depending on their usage/load. Images are a bit less unique per CR instance, but I do think as a general design principal for the operator(s), we want:
1. Defaults for "dynamic" configurations like images, resources, etc. should live in a config file for the operator. We basically have this today, and its great that the operator restarts/re-reconciles those defaults changing whenever the configmap is updated.
2. The ability to override the defaults on a per CR basis. Application-level configuration and resources in particular go hand-in-hand. If you add more domains, worker threads, etc. you need more CPU/memory to accommodate it.

I think this PR solves 1., but it should be followed up with the ability to change the K8s/workload settings at the CR level as needed.

We could split these dynamic configs into two categories: admin and user. Admins are likely concerned with security so the image and securityContext, etc. should be only definable by the admins via the operator configmap, and the user just gets whatever has been chosen by the admin. User are concerned with the application-level config and performance, and so mainly the resources I think fall into that category ?

dwertent:
@onelapahead 

I see your point, but in Kubernetes, this kind of strict separation between "admin" and "user" controls doesn’t always make sense. For example, in many cases, admins may want to control resource allocations (due to cost constraints, quota enforcement, etc.), while in other cases, they might care more about controlling the image (for security or versioning reasons). Kubernetes itself does not make this distinction - a single Pod specification includes both the image and the resource requests/limits. 

This brings us to two key questions:  
1. Do we want to allow specific deployment configurations for each Paladin node?  
2. What is the intended scope of the Paladin CRD?  

Regarding the first question, based on our previous discussions (referencing issue #582 ), we believe that in production environments, there will typically be a single Paladin node running per namespace. Given this, there is no strong need to configure individual node-level settings. Instead, having the operator manage settings such as labels, images, and resource allocations at the namespace level should be sufficient for all Paladin instances deployed within that namespace.  

For the second question, up until now, the Paladin CRD has served purely as an application-level configuration file rather than a deployment configuration. If we want to extend its scope to include deployment settings (e.g., resources, images), then we should consider either:  
- Introducing a separate CRD - for deployment-specific configurations.  
- Restructuring the existing Paladin CRD - to clearly separate application settings from deployment settings.

PR #521, opened 2025-01-21
Adding e2e to the CI pipeline

dwertent:


core/go/pkg/bootstrap/instance.go
```diff
@@ -78,6 +80,16 @@ func (i *instance) signalHandler() {
 	}
 }
 
+func (i *instance) startProfiler() {
+	go func() {
+		address := "localhost:6060"
+		log.L(i.ctx).Infof("Starting profiler on '%s'", address)
```
peterbroadhurst, line 86: Just want to confirm this isn't duplicative of https://github.com/LF-Decentralized-Trust-labs/paladin/blob/7b55a1d40e58895baf7757e6de97ff57a852c019/operator/internal/controller/paladin_controller.go#L680-L685

peterbroadhurst (commented):
Great work @dwertent!

One final question before merge

core/go/pkg/bootstrap/instance.go
dwertent: It is, I'll remove it :)

dwertent: Taking a second look, I see the debug server is configurable, but it's not the pprofiler server.
I will add the import for the pprofiler so it will be supported on the same port.

peterbroadhurst (approved):

PR #526, opened 2025-01-23
Adding NFT Support to the Zeto Domain

dwertent:
## Overview  

This PR introduces core support for **Non-Fungible Tokens (NFTs)** in the **Zeto domain**. This lays the foundation for handling NFT transactions, including `minting` and `transfers`.

### Roadmap 
- [x] Core NFT support to the Zeto domain  
- [x] Integration tests   
- [ ] Add support for NFT handling in **operator installation**  
- [ ] Add support for NFT token implementations using **nullifiers** for NFT transactions  
- [ ] Include an **NFT example in the tutorial** to showcase usage


domains/zeto/internal/zeto/nonfungible/handler_transfer.go
```diff
@@ -0,0 +1,380 @@
...
+		protoExtras, err := proto.Marshal(extrasObj)
+		if err != nil {
+			return nil, i18n.NewError(ctx, msgs.MsgErrorMarshalExtraObj, err)
+		}
+		extras = protoExtras
+	}
+	*/
+
+	provingRequestExtras := &corepb.ProvingRequestExtras_NonFungible{
+		TokenIds:  tokenIDs,
```
jimthematrix, line 277: to be consistent, should move the `InputValues` and `OutputValues` to a new protobuf `ProvingRequestExtras_Fungible`. or I think a better approach is doing the following:
- add a new property in the `ProvingRequestCommon` definition called "tokenType", it can be "fungible" or "non-fungible"
- add a property of type `bytes` and call it "tokenSecrets". it can be unmarshal'ed into either of the following protobuf definitions, depending on the value of the token type above:
  - `TokenSecrets_Fungible`, which has `InputValues` and `OutputValues`
  - `TokenSecrets_NonFungible`, which has `TokenIds` and `TokenUris`

domains/zeto/internal/zeto/nonfungible/utils.go
```diff
@@ -0,0 +1,45 @@
...
+	findVerifierFunc          func(string, string, string, []*pb.ResolvedVerifier) *pb.ResolvedVerifier                                = domain.FindVerifier
+	findAttestationFunc       func(string, []*pb.AttestationResult) *pb.AttestationResult                                              = domain.FindAttestation
+)
+
+func getAlgoZetoSnarkBJJ(name string) string {
+	return zetosignerapi.AlgoDomainZetoSnarkBJJ(name)
+}
+
+// Generate a random 256-bit integer
+func cryptoRand256() (*big.Int, error) {
```
jimthematrix, line 32: this doesn't look specific to the non-fungible implementation (salt should be the same), shouldn't re-implement. should be moved to the `common` package

domains/zeto/internal/zeto/signer/inputs_assembler.go
```diff
@@ -39,7 +40,36 @@ func assembleInputs_anon(inputs *commonWitnessInputs, keyEntry *core.KeyEntry) m
...
 		"outputOwnerPublicKeys": inputs.outputOwnerPublicKeys,
 	}
-	return witnessInputs
+}
+func assembleInputs_nfanon(ctx context.Context, inputs *commonWitnessInputs, tokenData *pb.ProvingRequestExtras_NonFungible, keyEntry *core.KeyEntry) map[string]interface{} {
+	tokenIds := make([]*big.Int, len(tokenData.TokenIds))
+	for i, id := range tokenData.TokenIds {
+		t, k := new(big.Int).SetString(id, 0)
+		if !k {
+			panic(i18n.NewError(ctx, msgs.MsgErrorTokenIDToString, id))
```
jimthematrix, line 49: shouldn't crash the process with errors related to a single transaction payload

solidity/contracts/domains/interfaces/IZetoPrivate.sol
```diff
@@ -2,9 +2,12 @@
 pragma solidity ^0.8.20;
 
 /**
- * @title IZetoPrivate
+ * @title IZetoPrivate (Deprecated)
```
jimthematrix, line 5: no need to have this to complicate things as we haven't had any official releases.

jimthematrix (changes requested):
overall looks very good! I like the direction with the restructuring. pointed out a few places for some further optimizations to better share code b/w the fungible and non-fungible packages

jimthematrix (approved):
approved with small updates requested. go ahead to address them if you are making updates anyway, or otherwise leave them to a future PR

PR #529, opened 2025-01-24
Feedback from #515

peterbroadhurst:
Address feedback from #515 

@hosie - please let me know if there are additional items you want me to cover immediately (vs. thoughts for future work).


hosie (approved):
That's great.  Thanks @peterbroadhurst

PR #532, opened 2025-01-24
Query keys RPC method

gabriel-indik:
Thought process around the design of the code in this PR: initially this was being modeled to provide the necessary API for the graphical user interface to allow the user to explore keys in the form of a tree. There would be a root node which the user could expand and see what segments are present. Each segment may be just a segment, a key, or a segment and a key.
After Looking into it further and discussing with @peterbroadhurst a different approach was evaluated. Instead of using a tree structure, use a combination of a table and a breadcrumb. For this the design of the API had to be adjusted in order to optimize performance. Specifically, to minimize the number of API calls, rather than providing a way to explore nodes, we introduced a key query method. The query can contain filters such as "parent=a.b.c" which would allow us to combine a breadcrumb with a table as described earlier in order to provide the same level of navigation as a tree, but with the advantage that information will be displayed in detail at every level explored. Note that this also opens more doors for adding filters and sorting options.


core/go/internal/keymanager/dbtypes.go
```diff
@@ -45,3 +47,15 @@ type DBKeyVerifier struct {
...
+var KeyEntryFilters filters.FieldSet = filters.FieldMap{
+	"isKey":       filters.BooleanField("isKey"),
+	"hasChildren": filters.BooleanField("hasChildren"),
+	"parent":      filters.StringField("parent"),
+	"index":       filters.Int64Field("index"),
+	"path":        filters.StringField("path"),
+	"type":        filters.StringField("type"),
+	"verifier":    filters.StringField("verifier"),
+	"wallet":      filters.StringField("wallet"),
+	"keyHandle":   filters.StringField("keyKandle"),
```
peterbroadhurst, line 60: Typo here

core/go/internal/keymanager/keymanager_rpc_test.go
```diff
@@ -50,6 +51,78 @@ func TestRPCLocalDetails(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, "m/44'/60'/1'/0/0", resolvedKey.KeyHandle)
 
+	/*
+			       Keys used for testing
+		    |---------------------------------|
```
peterbroadhurst, line 56: Tiny nit visible in Github UI - looks like one of the tabs ended up as spaces on just this line

core/go/internal/keymanager/manager.go
```diff
@@ -242,3 +244,34 @@ func (km *keyManager) ReverseKeyLookup(ctx context.Context, dbTX *gorm.DB, algor
...
+		`key_verifiers.type AS "type",` +
+		`key_verifiers.algorithm AS "algorithm",` +
+		`key_verifiers.verifier AS "verifier",` +
+		`key_mappings.wallet AS "wallet",` +
+		`key_mappings.key_handle AS "key_handle"`,
+	)
+
+	q.Joins("LEFT OUTER JOIN key_verifiers ON key_paths.path = key_verifiers.identifier")
+	q.Joins("LEFT OUTER JOIN key_mappings ON key_paths.path = key_mappings.identifier")
+	q.Joins(`LEFT OUTER JOIN (SELECT parent AS "p" from key_paths AS p) AS k ON key_paths.path = k.p`)
```
peterbroadhurst, line 268: Just a question on how this behaves when there are lots of children

gabriel-indik: Excellent point, we need to introduce distinct to address that scenario. I'll update the code.

core/go/internal/keymanager/dbtypes.go
gabriel-indik: Great catch, addressed.

core/go/internal/keymanager/keymanager_rpc_test.go
gabriel-indik: Interestingly in my local environment this shows fine. I too noticed it does not appear correctly in the PR on GitHub. I re-wrote that comment, hopefully that will address the formatting issue.

peterbroadhurst (approved):
This is great, thanks @gabriel-indik

peterbroadhurst:
Looks like before merge you just need to check the SQL with SQLite as well as PSQL.
`TestRPCLocalDetails` is failing in the build.

PR #534, opened 2025-01-26
Privacy group management & off-chain messaging

peterbroadhurst:
## Summary

This PR adds two big ticket features to Paladin:

1. Making a `PrivacyGroup` a first class object in Paladin
   - Previously it was something implemented entirely within a domain (like `Pente`)
   - Distribution of data about the privacy group is now automatic to all members of that group
   - EVM transactions can be submitted to the group, without needing to do complex wrapping
   - The Domain implements new lifecycle actions to perform the wrapping/validation on behalf of users
2. Adding off-chain messaging as an additional feature within a privacy group
   - For use where the strict ordering and non-repudiation provided by the blockchain is not required
   - Same mechanism used to distribute states and privacy group information
   - Provides WebSocket based reliable notification system (like receipt listeners)

## Scope

- [x] Lifecycle management APIs for privacy groups
- [x] Distribution of privacy groups as objects between Paladin nodes
- [x] EVM transaction submission directly to privacy groups, as similar to `eth_sendTransaction` as possible
- [x] Private un-pinned (no blockchain) messaging within a privacy group, additional to domain based smart contracts
- [x] Update Pente to implement the necessary Domain APIs
- [x] Update Samples + E2E tests to validate and use this new function
- [x] Docs updates

This is not the full scope of what privacy groups could evolve to being in Paladin, but it represents a very significant simplifying step for app developers. I believe it also takes us beyond the feature set for privacy group management provided in any previous generation of Privacy-group/Channel tech (particularly for managing huge numbers of privacy groups).

## New JSON/RPC APIs

- `pgroup_createGroup`
    - Create a new privacy group - returns object containing the `transaction` to track as well as the `id`
- `pgroup_getGroupById`
    - Get a privacy group by id (which is the genesis state ID - see data model below)
- `pgroup_getGroupByAddress`
    - Get a privacy group by its deployment address - once it's been confirmed via a blockchain receipt
- `pgroup_queryGroups`
    - Query groups by their basic properties (id, created time etc.)
- `pgroup_queryGroupsWithMember`
    - Query groups by their basic properties, scoped to a particular member
- `pgroup_sendTransaction`
    - Send an EVM transaction into the privacy group, to be processed by the domain, then turned into a private txn
    - Note the resulting private TX contained the mapped inputs, as derived by the domain (see architecture below)
- `pgroup_call`
    - Send an EVM call into the privacy group, and return the result
- `pgroup_createMessageListener`
    - Create a new message listener (near identical in function to `ptx_createReceiptListener`)
- `pgroup_queryMessageListeners`
    - Query messages listener (near identical in function to `ptx_queryReceiptListeners`)
- `pgroup_getMessageListener`
    - Get a message listener by name (near identical in function to `ptx_getReceiptListener`)
- `pgroup_startMessageListener`
    - Start a message listener by name (near identical in function to `ptx_startReceiptListener`)
- `pgroup_stopMessageListener`
    - Stop a message listener by name (near identical in function to `ptx_stopReceiptListener`)
- `pgroup_deleteMessageListener`
    - Delete a message listener by name (near identical in function to `ptx_deleteReceiptListener`)
- `pgroup_sendMessage`
    - Send a message to all members of a privacy group, identified by ID
- `pgroup_getMessageById`
    - Get an individual privacy group message by ID
- `pgroup_queryMessages`
    - Query messages in a privacy group
- `pgroup_subscribe`/`pgroup_unsubscribe`/`pgroup_ack`/`pgroup_nack`
    - WebSocket subscriptions (same as equivalent `ptx_` calls)
- `pstate_getSchemaById`
    - Useful to get a schema directly by its ID
- `transport_queryReliableMessages`
    - Query _outgoing_ transport-level reliable messages
    - Note these are not the same as privacy group messages, but they are the envelope for transferring them
    - These are used to transport privacy groups as well, as well as state distributions
- `transport_queryReliableMessageAcks`
    - Query transport-level acks received for sent messages

## New Domain lifecycle gRPC APIs

- `InitPrivacyGroup`
    - Takes the user-supplied properties and merges them with required properties, to generate:
        1. The genesis state
        2. The genesis _private_ transaction (which is the same as can manually be created)
- `ValidatePrivacyGroup`
    - Called to validate a privacy group received from another party, before it is stored for local use
    - Extracts the `members` from the privacy group as that is Paladin-level queryable information about every group
- `WrapPrivacyGroupEVMTX`
    - Takes a `pgroup_sendTransaction`/`pgroup_call` request from a user, as close to a "normal" EVM TX as possible
    - Maps to a private transaction to execute that EVM transaction within the privacy group

## Architecture

The biggest architectural decision in this PR is to make everything **additive**.

The way privacy groups work in Pente, has not changed. All previous ones work.

The privacy group management system is a layer on top, distributing information about privacy groups using the data model described below, and allowing messaging within those groups.

There is an implication to this in terms of having both off-chain (privately distributed) and on-chain (transparently distributed via event) configuration for the genesis of the chain, and I've covered the detailed implications of that in this comment:

https://github.com/LF-Decentralized-Trust-labs/paladin/blob/dfe9956a1358345187c6706e5406e7550c37e8fb/domains/pente/src/main/java/io/kaleido/paladin/pente/domain/PenteDomain.java#L731-L742

### Data model for Privacy Groups

I am proposing that each Privacy Group is identified by the ID of a `state`.

This means it has:
- A domain
- A schema ID
- Members
- A name (not guaranteed to be unique)
- A genesis salt (used inside the state)
- Domain specified configuration
    - Constrained to string<->string properties
- Arbitrary numbers of custom properties
    - Constrained to string<->string properties

Due to the distributed nature of privacy groups, where they are generated by one party and shared with another, this has lots of really important characteristics around clash-prevention. We can't do something as simplistic as assuming a single common property called `name` can be made globally unique. We also can't assume any group of parties will have a single privacy group - there could be thousands/millions, if a new privacy group is used for each bi-lateral transaction for example.

~It also means the ABI schema for a privacy group will need to be distributed with that privacy group, and we'll need a new domain API integration point to validate that a schema and privacy group is parsable by it.~

> We decided to have a single flexible Paladin ABI schema, which was possible when we decided that the `name`, `id` and `members` (and `created` time) are the only queryable fields of a privacy group.

### Private off-chain messaging

Privacy group messaging is a completely optional feature that can be used in any privacy group that has been created using the privacy group management APIs.

> NOTE: In this PR is necessary to specify a domain when creating a privacy group, and thus a blockchain factory-deploy transaction will occur on-chain when creating any privacy group (even if the primary purpose is to exchange off-chain private messages).

Messages have the following characteristics:
- Delivery is **exactly once** to each node
    - Each message is stored with a UUID
    - An individual `reliable message` (in the transport) is created to distribute it to each member of the privacy group
    - Only one copy is sent to each node (regardless of how many members of the privacy group are hosted on that node)
    - Reliable messages retry indefinitely, until they receive an `ack` or a `nack`
    - If a `nack` is received, then the message will be abandoned
    - The receiver is responsible for distinguishing between silence (retry), `ack` and `nack`
- Order is **not guaranteed**
    - Identically to state distribution, the delivery of messages is not strictly ordered
    - Messages can overtake each other in retry scenarios
    - If strict message ordering is required then _use a private blockchain transaction_ (that is the point of the blockchain)

### Deferred scope

In this scope I deliberately defer some key things that have been discussed as vision:
- Support for `eth_sendRawTransaction` semantics
    - Requires synchronous assemble - pushes on big picture items around externally managed keys / signing / proofs
- Support for `eth_getTranasctionReceipt` semantics
    - Requires nonce to be known at send TX time, so pulls in same challenges as `eth_sendRawTransaction`
- Support for `eth_getBlockByNumber` / `eth_getLogs` etc. semantics
    - Some bigger decisions to be made before we can treat these privacy groups like full EVM chains

However, this PR is a step towards all of the above.


ryjones:
Using stalebot is an antipattern :)

doc-site/docs/index.md
```diff
@@ -2,10 +2,12 @@
...
 Paladin provides programmable privacy for EVM.
 
-![Paladin overview](./images/paladin_overview.svg)
+[![Paladin overview](./images/paladin_overview.svg)](./images/paladin_overview.svg)
 
 Paladin gives enterprises control of sensitive information without compromising transparency or scalability.
 
+[READ THE LAUNCH BLOG](https://www.lfdecentralizedtrust.org/blog/announcing-paladin-an-lf-decentralized-trust-lab-for-programmable-privacy-on-evm)
+
 The Paladin is Apache 2.0 open source, with open governance through [Linux Foundation Decentralized Trust](https://www.lfdecentralizedtrust.org).
```
awrichar, line 11: "The Paladin"

sdk/typescript/src/interfaces/transaction.ts
```diff
@@ -1,6 +1,7 @@
+import { HexString } from "ethers/lib.commonjs/utils/data";
```
awrichar, line 1: Is this used?

sdk/typescript/src/domains/pente.ts
```diff
@@ -69,7 +71,7 @@ export const privateDeployABI = (
   ],
 });
 
-const privateInvokeABI = (
+export const privateInvokeABI = (
```
awrichar, line 74: Is this needed anymore?

sdk/typescript/src/domains/pente.ts
```diff
@@ -82,7 +84,7 @@ const privateInvokeABI = (
   ],
 });
 
-const privateCallABI = (
+export const privateCallABI = (
```
awrichar, line 87: Is this needed anymore?

example/bond/src/index.ts
```diff
@@ -54,14 +54,11 @@ async function main(): Promise<boolean> {
...
-    group: {
-      salt: newGroupSalt(),
-      members: [bondIssuer, bondCustodian],
-    },
+  const issuerCustodianGroup = await penteFactory.newPrivacyGroup({
+    members: [bondIssuer, bondCustodian],
     evmVersion: "shanghai",
     endorsementType: "group_scoped_identities",
-    externalCallsEnabled: true,
+    externalCallsEnabled: true,
```
awrichar, line 61: nit: trailing whitespace here and on 165

doc-site/docs/architecture/pente.md
```diff
@@ -99,97 +99,59 @@ The guiding principals are:
...
-        {"name": "externalCallsEnabled", "type": "boolean"},
-    ]
-}
-```
+1. Create a privacy group using the [pgroup_createGroup](../../reference/apis/pgroup/#pgroup_creategroup) JSON/RPC API
+    - This distributes information about the privacy group off-chain to all members
+    - You can add properties to your privacy group to allow all members to find it (a business transaction identifier for example)
+2. Wait for a receipt to the transaction returned in the [PrivacyGroup](../../reference/types/privacygroup/)
+    - This means that the blockchain has confirmed creation of the privacy group
+3. Submit transactions using the [pgroup_sendtransaction](../../reference/apis/pgroup/#pgroup_sendtransaction) JSON/RPC API
```
awrichar, line 116: Should it be capitalized as `pgroup_sendTransaction`?

doc-site/docs/architecture/pente.md
```diff
@@ -99,97 +99,59 @@ The guiding principals are:
...
-* **externalCallsEnabled** - true to allow this privacy group to trigger external calls to other contracts on the base ledger
+Privacy groups also allow you to send private messages that are completely off-chain, to distribute information
+without the latency (and gas) overhead of mining a blockchain transaction.
 
-### deploy
+> There are no ordering or non-repudiation guarantees when doing this, as the shared ordering context
+> and transaction signature (+execution) verification of a full EVM blockchain transaction are bypassed.
 
-Deploys a new contract into the private EVM group.
+See [pgroup_sendmessage](../../reference/apis/pgroup/#pgroup_sendmessage) for more details of sending messages.
```
awrichar, line 136: Is this `pgroup_sendMessage`?

toolkit/go/pkg/tktypes/abi_inference.go
```diff
@@ -0,0 +1,130 @@
...
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
+)
+
+// Takes arbitrary JSON data and attempts to infer an ABI from it.
+// Is not a complete system for all JSON.
+// Current capability set:
+// - Requires top-level JSON to be an object
+// - Does NOT preserve order of properties in a JSON object
+//   - This would require much more complex json.Token based code, or a swap-in library like
```
awrichar, line 39: Feels like this comment got garbled

toolkit/go/pkg/pldapi/states.go
```diff
@@ -91,21 +91,21 @@ type Schema struct {
...
 type StateEncoded struct {
-	ID              tktypes.HexBytes   `json:"id"`
-	DomainName      string             `json:"domain"`
-	Schema          tktypes.Bytes32    `json:"schema"`
-	ContractAddress tktypes.EthAddress `json:"contractAddress"`
-	Data            tktypes.HexBytes   `json:"data"`
+	ID              tktypes.HexBytes    `json:"id"`
+	DomainName      string              `json:"domain"`
+	Schema          tktypes.Bytes32     `json:"schema"`
+	ContractAddress *tktypes.EthAddress `json:"contractAddress"` // nil used for states like privacy group genesis that exists before state creation
```
awrichar, line 107: Should we have the "nil" comment on 98 as well?

core/go/internal/groupmgr/manager.go
```diff
@@ -0,0 +1,646 @@
...
+
+func (gm *groupManager) Stop() {
+	gm.rpcEventStreams.stop()
+	gm.stopMessageListeners()
+	gm.cancelCtx()
+}
+
+func (gm *groupManager) validateProperties(ctx context.Context, spec *pldapi.PrivacyGroupInput) (err error) {
+
+	// If no specific ABI has been provided for the input properties, we infer one from the properties
```
awrichar, line 150: This is so far the thing that has surprised me the most. I worry a little bit about inferring an ABI from JSON values. In that case why use ABI? I may need to keep reading some more to understand why we need an ABI and also why it's safe to infer it (when we've required it explicitly everywhere else in Paladin so far).

core/go/internal/groupmgr/manager.go
```diff
@@ -0,0 +1,646 @@
...
+}
+
+func (gm *groupManager) validateMembers(ctx context.Context, members []string) (remoteMembers map[string][]string, err error) {
+	localNode := gm.transportManager.LocalNodeName()
+	remoteMembers = make(map[string][]string)
+	if len(members) == 0 {
+		return nil, i18n.NewError(ctx, msgs.MsgPGroupsNoMembers)
+	}
+	for _, m := range members {
+		_, node, err := tktypes.PrivateIdentityLocator(m).Validate(ctx, "", false)
```
awrichar, line 191: Does this indicate that all member names must be fully qualified with node names?

domains/pente/src/main/java/io/kaleido/paladin/pente/domain/PenteDomain.java
```diff
@@ -571,25 +575,292 @@ protected CompletableFuture<ToDomain.BuildReceiptResponse> buildReceipt(ToDomain
...
+             var finalABI = JsonABI.newTuple("PentePrivacyGroup", "PentePrivacyGroup", JsonABI.newParameters(
+                     JsonABI.newIndexedParameter("salt", "bytes32"),
+                     JsonABI.newTuple("pente", "PentePrivacyGroupSettings", JsonABI.newParameters(
+                             JsonABI.newParameter("members", "string[]"),
+                             JsonABI.newParameter("evmVersion", "string"),
+                             JsonABI.newParameter("endorsementType", "string"),
+                             JsonABI.newParameter("externalCallsEnabled", "bool")
+                     ))
+             ));
+             var propertiesABI = mapper.readValue(request.getPropertiesAbiJson(), JsonABI.Parameters.class);
```
awrichar, line 619: So the implication is that the `propertiesABI` I pass in when initializing the privacy group must be a superset (or at least an overlapping set) with the `finalABI` defined here. At first glance this feels a bit confusing. Should there be two separate ABIs for "well-known properties set by the domain" and "user-defined properties set on group creation"?

awrichar: Also why is "salt" randomly outside the "pente" section?

core/go/internal/groupmgr/manager.go
```diff
@@ -0,0 +1,646 @@
...
+		GenesisTX:       txIDs[0],
+	}
+	if err = gm.insertGroup(ctx, dbTX, dbPG, spec.Members); err != nil {
+		return nil, err
+	}
+	group = dbPG.mapToAPI()
+	group.Genesis = states[0].Data
+	group.Members = spec.Members
+
+	// We also need to create a reliable send the state to all the remote members
```
awrichar, line 310: nit: feels like a word missing here

core/go/internal/groupmgr/manager.go
```diff
@@ -0,0 +1,646 @@
...
+	}
+	groupIDs := make([]tktypes.HexBytes, len(pgs))
+	for i, pg := range pgs {
+		groupIDs[i] = pg.ID
+	}
+	var dbMembers []*persistedGroupMember
+	err := dbTX.DB().WithContext(ctx).
+		Where(`"group" IN ( ? )`, groupIDs).
+		Order("domain").
+		Order(`"group"`).
```
awrichar, line 377: Why is this quoted like it is?

core/go/internal/groupmgr/message_listener.go
```diff
@@ -0,0 +1,705 @@
...
+	log.L(l.ctx).Infof("Added message %d/%s (domain='%s') to batch %d", pm.LocalSeq, pm.ID, pm.Domain, b.ID)
+	b.Messages = append(b.Messages, pm.mapToAPI())
+}
+
+func (l *messageListener) nextReceiver(b *messageDeliveryBatch) (r components.PrivacyGroupMessageReceiver, err error) {
+
+	for {
+		l.receiverLock.Lock()
+		if len(l.receivers) > 0 {
+			r = l.receivers[int(b.ID)%len(l.receivers)]
```
awrichar, line 575: Is this round-robin delivery?

core/go/internal/groupmgr/messages.go
```diff
@@ -0,0 +1,226 @@
...
+		return nil, err
+	}
+
+	// Create the reliable message delivery to the other parties
+	remoteMembers, err := gm.validateMembers(ctx, pg.Members)
+	if err != nil {
+		return nil, err
+	}
+
+	// We also need to create a reliable send the state to all the remote members
```
awrichar, line 116: nit: feels like a missing word here

core/go/internal/groupmgr/rpc_eventstreams.go
```diff
@@ -0,0 +1,207 @@
...
+	closed    chan struct{}
+}
+
+func (es *rpcEventStreams) HandleStart(ctx context.Context, req *rpcclient.RPCRequest, ctrl rpcserver.RPCAsyncControl) (rpcserver.RPCAsyncInstance, *rpcclient.RPCResponse) {
+	es.subLock.Lock()
+	defer es.subLock.Unlock()
+
+	var eventType tktypes.Enum[pldapi.PGroupEventType]
+	if len(req.Params) >= 1 {
+		eventType = tktypes.Enum[pldapi.PGroupEventType](req.Params[0].StringValue())
```
awrichar, line 72: Could maybe be clearer about these params... I'm seeing that:
- `req.Params[0]` is the event type (currently only one type "messages" supported)
- for type "messages", there should be exactly one additional parameter, and that is the listener name

But it took a few minutes of searching around to infer that.

core/go/internal/groupmgr/manager.go
peterbroadhurst: I also found this the most awkward thing to work out. How to have a field-queryable state, while encompassing everything in the privacy group.

Been having a deep think about this, and I think I have a different idea (would address this and the salt thing).

High level is to compress all domain options into an “extra data” style by testing. Then members and salt become mandatory fields at the paladin (not domain level). This means the user can be in direct control of the schema, and the state hash can become the salt of the privacy group.

Going to work on getting that to the point we could talk it through. There’s still a bit of chicken and egg in there, and it means some rework.

sdk/typescript/src/domains/pente.ts
```diff
@@ -281,20 +287,22 @@ export class PentePrivacyGroup {
...
-  ) {}
+  ) {
+  }
 
   abstract using(
     paladin: PaladinClient
   ): PentePrivateContract<ConstructorParams>;
 
-  async sendTransaction(transaction: PenteContractTransactionInput){ 
+  async sendTransaction(transaction: PenteContractTransactionInput, txOptions?: Partial<IPrivacyGroupEVMTXInput>){
```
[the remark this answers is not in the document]
peterbroadhurst: So the intent here is to provide overrides. It's not that the options are optional or not in the core structure.
So I could define a new structure, but to follow the pattern I'm proposing here, it would be identical to `Partial<IPrivacyGroupEVMTXInput>` (e.g. every field is `?`)

core/go/internal/groupmgr/manager.go
peterbroadhurst: Ok - doing this thought exercise I think I came to a proposal.

Note the constraints I'm designing to with:

- User supplies some custom properties like `color: "blue"`
    - These need to be indexed to be useful
    - This **must** change the ABI for these to be `indexed: true` (cannot use a generic name/value set)
- Paladin has some fundamental requirements
    - `members` must be recoverable from the state (to extract on receipt on a remote node)
    - `salt` must be present in all states (for them to be secure)
- Privacy groups need to allow customization of their settings
    - Think of `evmVersion` or `externalCallsEnabled` for Pente
    - I had believed we wanted these encoded into the genesis state as well

Now I had really wanted the Pente `salt` at group creation time to be the `id` of the state - *not the `salt` inside the state*.

I couldn't make that happen, because it's not known until the state is finalized, and the state cannot be finalized until Pente is invoked to add in the settings.

So I think I need to have _two gRPC_ calls on state creation:

1. To get Pente to finalize the options - so they can be placed in the state
2. To generate the private transaction - now the state ID is known

This would allow me to move the State Schema supply/generation out of the Domain layer, into the Paladin layer - putting the user in full control of it.

awrichar: I believe I agree with nearly everything above (including splitting the gRPC calls as proposed). The only thing I'm not sure about is this:

> - User supplies some custom properties like `color: "blue"`
    - These need to be indexed to be useful
    - This **must** change the ABI for these to be `indexed: true` (cannot use a generic name/value set)

It feels like we've added a lot of complexity in order to support privacy groups having a user-customizable schema with queryable fields. This implies that I'd regularly want to look up a _set_ of related privacy groups (i.e. all the "blue" groups) and I'm just not coming up with a good reason for that. In general I want to locate _the specific_ privacy group that I need to use for a particular transaction.

Is it possible that we just need a `name` for the group (which the _user_ must make unique), and then a set of _non-queryable_ custom metadata properties?

peterbroadhurst:
Popping back into review while I work on the great discussion from @awrichar that will let me simplify the state system.

core/go/internal/groupmgr/manager.go
peterbroadhurst: Yes. I've updated the field docs to make that clear.

domains/pente/src/main/java/io/kaleido/paladin/pente/domain/PenteDomain.java
peterbroadhurst: This has all be reconciled into a much simpler system.

Now the `salt` of the genesis state is always generated, and is called `genesisSalt` - to distinguish it from the `salt` that is the previously named (and unchanged) input into each Pente transaction (which is now the `id` of the genesis state / Privacy Group)

core/go/internal/groupmgr/manager.go
peterbroadhurst: Because `group` is a special SQL word that has to be escaped

core/go/internal/groupmgr/message_listener.go
peterbroadhurst: Yes, of batches (and associated acknowledgement), and only when you have multiple receivers attached to a single listener (multiple active subscribed websockets).

core/go/internal/groupmgr/rpc_eventstreams.go
peterbroadhurst: Understood. I've put together some docs, but event docs could be more comprehensive:

https://lf-decentralized-trust-labs.github.io/paladin/head/reference/types/transactionreceiptlisteneroptions/#create-receipt-listener (which is moving to `CreateReceiptListener` in this PR)

core/go/internal/groupmgr/manager.go
peterbroadhurst: I did some renaming as I went through the implementation and hope the above is simpler.

toolkit/go/pkg/tktypes/abi_inference.go
awrichar, line 1: Is this file still needed, or did we remove the need with the latest changes?

sdk/typescript/src/domains/pente.ts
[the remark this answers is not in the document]
awrichar: OK, I think it's fine for now

domains/pente/src/main/java/io/kaleido/paladin/pente/domain/PenteDomain.java
```diff
@@ -571,25 +575,268 @@ protected CompletableFuture<ToDomain.BuildReceiptResponse> buildReceipt(ToDomain
...
+             return CompletableFuture.completedFuture(BuildReceiptResponse.newBuilder().
                      setReceiptJson(new ObjectMapper().writerWithDefaultPrettyPrinter().writeValueAsString(jsonReceipt)).
                      build());
          } catch (Exception e) {
              return CompletableFuture.failedFuture(e);
          }
      }
- 
+
+     void validatePropType(String subSection, JsonABI.Parameter prop, String requiredType) throws IllegalArgumentException {
```
awrichar, line 597: I think this is now unused?

toolkit/go/pkg/tktypes/abi_inference.go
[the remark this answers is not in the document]
awrichar: Personally I would vote to delete it, although I know it represents some interesting dev work. It still feels like an odd pattern to me that we would infer an ABI schema based on the typing of JSON values.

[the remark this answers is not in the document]
peterbroadhurst: I might do that in a follow-up - so I have more hope of finding it again if we need it

domains/pente/src/main/java/io/kaleido/paladin/pente/domain/PenteDomain.java
```diff
@@ -571,25 +575,268 @@ protected CompletableFuture<ToDomain.BuildReceiptResponse> buildReceipt(ToDomain
...
+                 endorsementType = PenteConfiguration.DEFAULT_ENDORSEMENT_TYPE;
+             }
+             resBuilder.putConfiguration("endorsementType", endorsementType);
+
+             var externalCallsEnabled = inputConf.get("externalCallsEnabled");
+             if (externalCallsEnabled == null || externalCallsEnabled.isEmpty()) {
+                 externalCallsEnabled = "false";
+             }
+             resBuilder.putConfiguration("externalCallsEnabled", externalCallsEnabled);
+
```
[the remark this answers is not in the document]
peterbroadhurst: I made this a full error

PR #537, opened 2025-01-27
Key explorer UI

gabriel-indik:
User experience:
Uses the combination of a breadcrumb and a paginated table in order to provide efficient navigation through a (potentially very big) set of keys that are hierarchically organized.

image

Clicking on the folder icons (with tooltip "Open folder") takes the user to different sections of the key hierarchy:

image

Reverse lookup:

image

Features to note:
 - Makes use of new key query API
 - Provides sorting by name | index, ascending | descending
 - Pagination is based on referenced key entries depending on sort field and order
 - Preference saved in local storage for:
     - Sort field
     - Sort order
     - Rows per page
 - Query string captures location within navigation


ui/client/src/components/Hash.tsx
```diff
@@ -1,4 +1,4 @@
-// Copyright © 2024 Kaleido, Inc.
+// Copyright © 20245 Kaleido, Inc.
```
peterbroadhurst, line 1: 👨‍🚀

ui/client/src/views/Keys.tsx
```diff
@@ -0,0 +1,368 @@
...
+import { constants } from "../components/config";
+import SearchIcon from '@mui/icons-material/Search';
+import { ReverseKeyLookupDialog } from "../dialogs/ReverseKeyLookup";
+import RemoveIcon from '@mui/icons-material/Remove';
+import { VerifiersDialog } from "../dialogs/Verifiers";
+
+export const Keys: React.FC = () => {
+
+  const getDefaultRowsPerPage = () => {
+    const valueFromStorage = window.localStorage.getItem(constants.KEYS_ROWS_PER_PAGE);
```
peterbroadhurst, line 37: ok - so maybe now I understand why we have separate constants.
We're using separate sections in the storage, to keep the status of each table - so i you nav somewhere else, and then back to the table - we'll remember where you were (not just on the back button, as would be the case with the query strings in the location HREF).

ui/client/src/components/config.ts
```diff
@@ -16,6 +16,9 @@
 
 export const constants = {
   COLOR_MODE_STORAGE_KEY: 'color-mode',
+  KEYS_SORT_BY_STORAGE_KEY: 'keys-sort-by',
```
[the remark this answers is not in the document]
peterbroadhurst: (responded to myself on this one a bit later)

ui/client/src/views/Keys.tsx
```diff
@@ -0,0 +1,368 @@
...
+  };
+
+  const handleChangePage = (
+    _event: React.MouseEvent<HTMLButtonElement> | null,
+    newPage: number
+  ) => {
+    if (newPage === 0) {
+      setRefEntries([]);
+    } else if (newPage > page) {
+      refEntries.push(keys[keys.length - 1]);
```
peterbroadhurst, line 214: So the depth of the stack here will be the number of pages forwards you've gone

peterbroadhurst (changes requested):
Looks great.

Couple of questions/observations - but none are explicit requests to change. Just to have a think about (maybe mostly wrt to what we do in future PRs that re-use this code).

However, there is something odd with the build due to the case change, which I don't know if it's just local to me.

ui/client/src/components/config.ts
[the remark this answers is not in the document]
gabriel-indik: It's a good point that I too thought about, in fact originally I wrote the code exactly how you are describing. The reason I later switched to a prefixed key is because we may have multiple tables and each table may have a different set of columns. So the sort-by in one table may not apply to another.

ui/client/src/dialogs/ReverseKeyLookup.tsx
```diff
@@ -0,0 +1,192 @@
...
+        setIsEthereum(true);
+        setOtherType('');
+        setOtherAlgorithm('');
+        setNotFound(false);
+      }, 200);
+    }
+  }, [dialogOpen]);
+
+  useEffect(() => {
+    setType(isEthereum ? 'eth_address' : otherType);
```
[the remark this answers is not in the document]
gabriel-indik: Couldn't agree more, code updated.

ui/client/src/App.tsx
```diff
@@ -26,13 +26,14 @@ import { BrowserRouter, Navigate, Route, Routes } from "react-router-dom";
 import { Header } from "./components/Header";
 import { ApplicationContextProvider } from "./contexts/ApplicationContext";
 import { darkThemeOptions, lightThemeOptions } from "./themes/default";
-import { Indexer } from "./views/indexer";
+import { Indexer } from "./views/Indexer";
```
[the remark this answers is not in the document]
gabriel-indik: This was caused by the case change in the file name. Included a commit changing the file name, the back to the original with the correct casing. Hopefully that will address the issue.

ui/client/src/views/Keys.tsx
gabriel-indik: That's exactly right. Even if you close the web browser and open it again, we'll "remember" what the user preference was when it comes to what column they want to sort by and what order (asc/desc).

ui/client/src/views/Keys.tsx
gabriel-indik: That is exactly right. This auxiliary stack structure is kept in memory so that we can go back 1 page at a time. Going to the first page simply consists of clearing the stack.

peterbroadhurst (approved):
Thanks for working through all the questions 👍

PR #542, opened 2025-01-29
noto: docs on hooks, locking

awrichar:
In a chain with https://github.com/LF-Decentralized-Trust-labs/paladin/pull/525

Adds documentation for notary mode (basic/hooks) and for the various lock/unlock methods.


doc-site/docs/architecture/noto.md
```diff
@@ -272,6 +376,171 @@ Inputs:
...
+### Notary mode: basic
+
+When a Noto contract is constructed with notary mode `basic`, the following notary behaviors can be configured:
+
+| Option         | Default | Description |
+| -------------- | ------- | ----------- |
+| restrictMint   | true    | _True:_ only the notary may mint<br>_False:_ any party may mint |
+| allowBurn      | true    | _True:_ token owners may burn their tokens<br>_False:_ tokens cannot be burned |
+| allowLock      | true    | _True:_ token owners may lock tokens (for purposes such as preparing or delegating transfers)<br>_False:_ tokens cannot be locked (not recommended, as it restricts the ability to incorporate tokens into swaps and other workflows) |
+| restrictUnlock | true    | _True:_ only the creator of a lock may unlock it<br>_False:_ any party may unlock (not recommended, as it allows anyone to claim locked tokens)
```
jimthematrix, line 515: may need to add some text on why _False_ is allowed at all, since it sounds unsafe.

jimthematrix (approved):
lgtm with a small comment that can be addressed in a future PR

doc-site/docs/architecture/noto.md
awrichar: Yes, we can circle back on this. The main reason it is configurable in "basic" mode is because it is always off in "hooks" mode (so you must implement restrictions there if you want only the lock owner to unlock... but you can also do complex things to allow other parties to unlock if they provide some sort of other credentials as "data" to the unlock operation). I chose that anything which is "sometimes on and sometimes off" had a config flag.

We could make some of these things non-negotiable (always on) in "basic" mode. It probably just needs more explaining in terms of what things are always true in "basic" mode, what things are configurable in "basic" mode, and what things you must consider enforcing in "hooks" yourself because they're not enforced in the domain in "hooks" mode.

awrichar: Addressed in https://github.com/LF-Decentralized-Trust-labs/paladin/pull/596

PR #545, opened 2025-01-31
Zeto lock support

jimthematrix:


domains/zeto/internal/zeto/fungible/utils.go
```diff
@@ -40,13 +42,24 @@ import (
...
 }
 
 func (h *baseHandler) getAlgoZetoSnarkBJJ() string {
 	return getAlgoZetoSnarkBJJ(h.name)
 }
 
+// the Zeto implementations support two input/output sizes for the circuits: 2 and 10,
+// if the input or output size is larger than 2, then the batch circuit is used with
+// input/output size 10
+func getInputSize(sizeOfEndorsableStates int) int {
```
dwertent, line 56: This function is implemented at `common.GetInputSize`

domains/zeto/internal/zeto/fungible/utils.go
```diff
@@ -78,8 +91,44 @@ func validateAmountParam(ctx context.Context, amount *tktypes.HexUint256, i int)
...
-func generateMerkleProofs(ctx context.Context, callbacks plugintk.DomainCallbacks, merkleTreeRootSchema *prototk.StateSchema, merkleTreeNodeSchema *prototk.StateSchema, tokenName string, stateQueryContext string, contractAddress *tktypes.EthAddress, inputCoins []*types.ZetoCoin) ([]core.Proof, *corepb.ProvingRequestExtras_Nullifiers, error) {
+func utxosFromInputStates(ctx context.Context, states []*prototk.EndorsableState, desiredSize int) ([]string, error) {
+	return _utxosFromStates(ctx, states, desiredSize, true)
+}
+
+func utxosFromOutputStates(ctx context.Context, states []*prototk.EndorsableState, desiredSize int) ([]string, error) {
+	return _utxosFromStates(ctx, states, desiredSize, false)
+}
+
+func _utxosFromStates(ctx context.Context, states []*prototk.EndorsableState, desiredSize int, isInputs bool) ([]string, error) {
```
dwertent, line 102:
```suggestion
func utxosFromStates(ctx context.Context, states []*prototk.EndorsableState, desiredSize int, isInputs bool) ([]string, error) {
```

dwertent (commented):
I left some comments, not blockers

domains/zeto/integration-test/e2e_test.go
awrichar, line 1: Unrelated to this particular change, but do you think this test suite should move up a level (into `domains/integration-test` instead of `domains/zeto/integration-test`)?

awrichar: The Noto E2E test already moved up there.

domains/integration-test/pvp_test.go
```diff
@@ -428,8 +428,15 @@ func (s *pvpTestSuite) TestNotoForZeto() {
...
-	zeto.Lock(ctx, tktypes.MustEthAddress(bobKey.Verifier.Verifier), transferZeto.EncodedCall).SignAndSend(bob).Wait()
+	zeto.Lock(ctx, tktypes.MustEthAddress(bobKey.Verifier.Verifier), 1).SignAndSend(bob).Wait()
+
+	jq := query.NewQueryBuilder().Limit(100).Equal("locked", true).Query()
+	lockedZetoCoins := findAvailableCoins(t, ctx, rpc, zetoDomain.Name(), zetoDomain.CoinSchemaID(), zeto.Address, jq, func(coins []*zetotypes.ZetoCoinState) bool {
+		return len(coins) >= 1
+	})
+	lockedZeto, _ := lockedZetoCoins[0].Data.Hash(ctx)
+
+	transferZeto := zeto.TransferLocked(ctx, lockedZeto, bobKey.Verifier.Verifier, alice, 1).Prepare(bob)
```
awrichar, line 439: Note that we're still using `Prepare` - I'd like to continue thinking about whether we can do a simpler thing (specific to the Zeto domain and a single Paladin node) rather than needing to fire up the full coordinated transaction engine.

domains/zeto/integration-test/e2e_fungible_test.go
```diff
@@ -208,14 +210,92 @@ func (s *fungibleTestSuiteHelper) testZeto(t *testing.T, tokenName string, useBa
...
+	coins = findAvailableCoins(t, ctx, s.rpc, s.domain, zetoAddress, isNullifiersToken)
 	require.Len(t, coins, expectedCoins)
 
 	log.L(ctx).Info("*************************************")
 	log.L(ctx).Infof("Withdraw back to ERC20 balance from Zeto")
 	log.L(ctx).Info("*************************************")
 	_, err = s.withdraw(ctx, zetoAddress, controllerName, 100)
 	require.NoError(t, err)
+
+	if tokenName != constants.TOKEN_ANON {
```
awrichar, line 222: Might be worth a comment to indicate why this test is different for different tokens.

domains/zeto/integration-test/e2e_fungible_test.go
```diff
@@ -208,14 +210,92 @@ func (s *fungibleTestSuiteHelper) testZeto(t *testing.T, tokenName string, useBa
...
+	require.NoError(t, err)
+	_, err = s.lock(ctx, zetoAddress, controllerName, 1, recipient1EthAddr)
+	require.NoError(t, err)
+
+	coins = findAvailableCoins(t, ctx, s.rpc, s.domain, zetoAddress, isNullifiersToken, true /* locked */)
+	require.Len(t, coins, 2)
+	locked1, _ := coins[0].Data.Hash(ctx)
+	locked2, _ := coins[1].Data.Hash(ctx)
+
+	// for now unlock only works properly for the ANON token
```
awrichar, line 245: Maybe this comment just needs to be moved up to the circuit breaker at line 222

domains/zeto/integration-test/e2e_fungible_test.go
```diff
@@ -208,14 +210,92 @@ func (s *fungibleTestSuiteHelper) testZeto(t *testing.T, tokenName string, useBa
...
+		"utxos":    []string{locked1.String()},
+		"delegate": recipient2EthAddrStr,
+		"data":     "0x",
+	}
+	txInputJson, err := json.Marshal(txInput)
+	require.NoError(t, err)
+	// submit a public transaction to delegate the lock
+	build, err := getZetoAnonSpec()
+	require.NoError(t, err)
+	_, err = s.tb.ExecTransactionSync(ctx, &pldapi.TransactionInput{
```
awrichar, line 269: Can we use the Zeto helpers here?

domains/zeto/internal/zeto/fungible/handler_lock.go
```diff
@@ -233,70 +208,92 @@ func (h *lockHandler) Prepare(ctx context.Context, tx *types.ParsedTransaction,
...
-		"delegate": params.Delegate,
-		"data":     data,
+
+	lockedInfo, err := parseInfoStatesJson(req.InfoStates[0].StateDataJson)
+	if err != nil {
+		return nil, i18n.NewError(ctx, msgs.MsgErrorParseInfoStates, err)
 	}
-	paramsJSON, err := json.Marshal(LockParams)
+	unlockedOutputStates := req.OutputStates[:len(req.OutputStates)-int(lockedInfo.Size.Uint64())]
+	lockedOutputStates := req.OutputStates[len(req.OutputStates)-int(lockedInfo.Size.Uint64()):]
```
awrichar, line 222: Tracking the size of these slices feels a little bit fragile. Can we just look at the state data to separate them (either they are two different schemas, or they have a "locked" flag that tells whether they are a locked/unlocked state)?

domains/zeto/internal/zeto/fungible/handler_lock.go
```diff
@@ -233,70 +208,92 @@ func (h *lockHandler) Prepare(ctx context.Context, tx *types.ParsedTransaction,
...
+		"inputs":        inputs,
+		"outputs":       outputs,
+		"lockedOutputs": lockedOutputs,
+		"proof":         common.EncodeProof(proofRes.Proof),
+		"delegate":      tx.Params.(*types.LockParams).Delegate.String(),
+		"data":          data,
+	}
+	transferFunction := getLockABI(tx.DomainConfig.TokenName)
+	if common.IsNullifiersToken(tx.DomainConfig.TokenName) {
+		delete(params, "inputs")
```
awrichar, line 250: Rather than setting "inputs" and then deleting it, maybe we should only set it in an `else` case.

example/swap/src/index.ts
```diff
@@ -155,15 +155,33 @@ async function main(): Promise<boolean> {
...
+  });
+  if (lockedStateId === undefined) {
+    logger.error("No locked state found in state receipt");
+    return false;
+  } else {
+    logger.log(`Locked state ID: ${lockedStateId}`);
+  }
+
+  const txID = await zetoCash.using(paladin3).prepareTransferLocked(investor2, {
+    lockedInputs: [lockedStateId],
```
awrichar, line 180: What are the assumptions around locked states? It seems to be:
- Zeto will _always_ represent locked value as a single state
- The state ID of that locked state must be extracted by the user in order to spend it
- The user may choose to spend all of that state or some of it - what happens to the remaining value?

awrichar: If we do require the user to extract state IDs or other identifiers, I'd highly recommend that Zeto implements a "domain receipt" that makes it very clear to retrieve the necessary info, rather than forcing the user to parse the raw "states" list. Both Pente and Noto have now implemented domain receipts, but I think Zeto does not yet.

domains/zeto/integration-test/e2e_test.go
jimthematrix: my current thinking is that the `domains/zeto/integration-test` tests full permutation of the domain capabilities, while `domains/integration-test` tests cross-domain flows. But I don't have any objection to moving all testbed based integration-tests into the domain/integration-test as was done for Noto

example/swap/src/index.ts
jimthematrix: makes sense, opened https://github.com/LF-Decentralized-Trust-labs/paladin/issues/600 to add this in a future PR

jimthematrix: > Zeto will always represent locked value as a single state (we will never allow customizing, for instance restricting states to have a min/max value and therefore having multiple states created for a lock)

Zeto itself allows multiple locked outputs from the `lock()` function. but the current Paladin private API outputs only one locked state. can certainly enhance it to be more flexible if I can better understand the use cases

> The state ID of that locked state must be extracted by the user in order to spend it

The locked state ID are broadcast in the event. but I think having a more direct association of the resulting state ID with the submitted transaction via a domain receipt would be a good idea. so agreed with the suggestion to add that to Zeto

> The user may choose to spend all of that state or some of it - what happens to the remaining value?

If there are remaining value that are not spent by the `transferLocked()` call, they become unlocked. we could update the function signature to add `lockedOutputs` alongside `outputs`, but before making the update, I'd like to better understand the use cases

domains/integration-test/build.gradle
```diff
@@ -98,7 +109,7 @@ task test(type: Exec, dependsOn: [":core:go:makeMocks"]) {
     args '-v'
     args '-cover'
     args '-covermode=atomic'
-    args '-timeout=120s'
+    args '-timeout=480s'
```
awrichar, line 112: 😱

domains/integration-test/util.go
```diff
@@ -101,3 +159,33 @@ func newZetoDomain(t *testing.T, config *zetotypes.DomainFactoryConfig, factoryA
 	}
 	return waitForDomain, tbd
 }
+
+func findAvailableCoins[T any](t *testing.T, ctx context.Context, rpc rpcclient.Client, domainName, coinSchemaID, methodName string, address *tktypes.EthAddress, jq *query.QueryJSON, readiness ...func(coins []*T) bool) []*T {
```
awrichar, line 163: Rather than requiring an RPC method name to be passed in, should we consider one of these options:
- provide two helpers (one that queries states and another that queries nullifiers)
- use a boolean param to choose between states and nullifiers

domains/zeto/internal/zeto/fungible/handler_transferLocked.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+	params := map[string]any{
+		"inputs":  inputs,
+		"outputs": outputs,
+		"proof":   common.EncodeProof(proofRes.Proof),
+		"data":    data,
+	}
+	transferFunction := getTransferLockedABI(tx.DomainConfig.TokenName)
+	if common.IsNullifiersToken(tx.DomainConfig.TokenName) {
+		delete(params, "inputs")
```
awrichar, line 258: I think cleaner to add this only in an "else" block (rather than adding it and then removing it)

domains/zeto/internal/zeto/fungible/handler_transferLocked.go
```diff
@@ -0,0 +1,340 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	signer := &tx.Params.(*types.FungibleTransferLockedParams).Delegate
+	if req.Transaction.Intent == prototk.TransactionSpecification_PREPARE_TRANSACTION {
+		// if called as prepare, it means the delegate might not exist locally
+		// and the signer doesn't matter because the caller is interested in the
+		// prepared transaction, not the signature
+		signer = &req.Transaction.From
```
awrichar, line 276: Just another thing that feels very wonky about "prepare"

awrichar (approved):

domains/integration-test/util.go
jimthematrix: will make the update in my outstanding branch for adding locking support to the nullifier tokens

domains/zeto/internal/zeto/fungible/handler_transferLocked.go
jimthematrix: made the update in my outstanding branch for adding locking support to the nullifier tokens

domains/zeto/internal/zeto/fungible/handler_transferLocked.go
jimthematrix: "wonky" is the right word.

PR #552, opened 2025-02-05
Port i18n into toolkit

EnriqueL8:
So far Paladin was importing FireFly Common's i18n package to handle translation for error messages and descriptions, this would cause conflicts if an application imported both FireFly and Paladin.

This PR ports the i18n pkg into the Paladin toolkit. 

There is a discussion item on how strongly people feel that the registering a prefix should be done from each different pkg or we should move some core prefixes into the toolkit.  Happy to hear suggestions!


hosie (approved):
LGTM

EnriqueL8:
Can't seem to figure out why the build is failing, doesn't seem related to my changes

EnriqueL8:
Ah it's linting `:toolkit:go:lint`

PR #556, opened 2025-02-12
fixing concurrency issues in auto-fueling

Chengxuan:
the current logic is missing a mutex in the tracking map, so when there are multiple concurrent threads tracking fueling txs for different destination addresses, it will result in concurrent write issue to the tracking map.


annamcallister (approved):

hosie (approved):
LGTM

PR #557, opened 2025-02-12
Filters for UI

gabriel-indik:
As part of the development for this feature we took a look at how filters work in other projects that use the same type of QueryJSON API. While these filters work perfectly well, one issue that was observed is that they allow any field-operator combination. In practice, only some of those combinations are valid and the rest are not. This leads to situations where the user may inadvertently select an operator that is invalid for a given field, resulting in an API error. This error causes the page not to update any more. The following example shows the issue:

https://github.com/user-attachments/assets/a7cf52f8-8795-4aef-b295-103b14938a18

To solve the problem, a new design is introduced where the client side logic is data-type aware. This makes it possible to ensure that only valid field-operator combinations are selected. In addition, rather than having a filter condition "rule matches/does not match", the operator list has been expanded to include all permutations in an easy-to-understand list. The following example shows the changes suggested:

https://github.com/user-attachments/assets/9f5900c3-34f3-46b6-af75-c8986a416e5f

As it can be seen in the example, depending on which field is selected, the list of operators dynamically changes. When the first field "name" is selected, all string operators are included (contains, starts with, etc) in addition to other operators that are valid for both numbers and strings (greater than, greater than or equal, etc). When selecting a field that corresponds to a number however, the list changes to only show operators that are valid for numbers. Finally, when the "Is Folder" field is selected, only options for boolean values are displayed (equals to either true or false).


ui/client/src/dialogs/ReverseKeyLookup.tsx
```diff
@@ -87,9 +89,26 @@ export const ReverseKeyLookupDialog: React.FC<Props> = ({
         const index = path.lastIndexOf('.');
         if(index !== -1) {
           setParent(path.substring(0, index));
-          setFilter(path.substring(index + 1))
+          // setFilter(path.substring(index + 1))
```
shorsher, line 92: still need this?

ui/client/src/queries/keys.ts
```diff
@@ -14,41 +14,53 @@
...
 
-  if (pathFilter !== undefined) {
-    requestPayload.params[0].eq.push({
-      field: 'path',
-      value: parent !== '' ? `${parent}.${pathFilter}` : pathFilter
-    });
-  }
+  // if (pathFilter !== undefined) {
+  //   requestPayload.params[0].eq.push({
+  //     field: 'path',
```
shorsher, line 58: a few commented lines we could remove here if you want to clean up

shorsher (approved):
Approved with some minor nits for your consideration 🚀

ui/client/src/dialogs/ReverseKeyLookup.tsx
gabriel-indik: Great point, this was part of a refactoring where the filter was changed for a completely different function. I forgot to remove the commented out invocation of the previous version. Will address it.

ui/client/src/queries/keys.ts
gabriel-indik: Great point again. There was quite a bit of experimentation on this file, I did not clean it up after that was complete. Will clean it up now.

gabriel-indik:
Also included in this PR: resolve verifier functionality


https://github.com/user-attachments/assets/c96a16d1-be20-42af-bc54-6421eaac8c12

dwertent (approved):
Relying on @shorsher review

core/go/internal/keymanager/dbtypes.go
```diff
@@ -49,11 +49,12 @@ func (t DBKeyVerifier) TableName() string {
...
 
 var KeyEntryFilters filters.FieldSet = filters.FieldMap{
-	"isKey":       filters.BooleanField("isKey"),
-	"hasChildren": filters.BooleanField("hasChildren"),
+	"isKey":       filters.BooleanField("key_mappings.identifier IS NOT NULL"),
+	"hasChildren": filters.BooleanField("k.p IS NOT NULL"),
 	"parent":      filters.StringField("parent"),
 	"index":       filters.Int64Field(`"index"`),
 	"path":        filters.StringField("path"),
+	"name":        filters.StringField("REGEXP_SUBSTR(key_paths.path, '[^\\.]+$')"),
```
dwertent, line 57: @gabriel-indik it's looks like the tests are failing because of this

gabriel-indik: You are absolutely right, addressed the issue.

PR #559, opened 2025-02-13
Fixing mysql/postgres units

dwertent:
## Fix Race Condition in Unit Tests  

This PR resolves a race condition that occurred during unit testing.  

#### Issue  
The race condition caused inconsistent behavior in test assertions. As shown in the [logs](https://productionresultssa15.blob.core.windows.net/actions-results/8275b8c6-0280-4918-959b-d33aea76e79e/workflow-job-run-ca395085-040a-526b-2ce8-bdc85f692774/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-02-13T21%3A20%3A34Z&sig=qBxE9ojenB%2FxHfLe6sNjZ%2BZ60kp8IRXE1d3kbd3hJjY%3D&ske=2025-02-14T07%3A07%3A02Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-02-13T19%3A07%3A02Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-01-05&sp=r&spr=https&sr=b&st=2025-02-13T21%3A10%3A29Z&sv=2025-01-05), the first error had `maxAttempts=0`, while the second had `maxAttempts=1`:  

```
[2025-02-13T21:07:48.791Z] ERROR PD011642: Events ABI is invalid: invalid character '!' looking for beginning of value (attempt=1, maxAttempts=0) domain=test1  
[2025-02-13T21:07:49.041Z] ERROR all expectations were already fulfilled, call to database transaction Begin was not expected (attempt=2, maxAttempts=1) domain=test1  
```  

This discrepancy occurred because the `Do` function ran in a separate goroutine from the test setting `UTSetMaxAttempts`. As a result, the test encountered an unexpected error message, leading to test failures.  

#### Fix  
The solution is to initialize the retry logic with `maxRetries` at the start, rather than modifying it later. This ensures consistent behavior across test runs and eliminates the race condition.

### DIFF
--- a/core/go/internal/domainmgr/domain.go
+++ b/core/go/internal/domainmgr/domain.go
@@ -85,8 +85,8 @@ func (dm *domainManager) newDomain(name string, conf *pldconf.DomainConfig, toDo
 	d := &domain{
 		dm:              dm,
 		conf:            conf,
-		defaultGasLimit: pldconf.DefaultDefaultGasLimit, // can be set by config below
-		initRetry:       retry.NewRetryLimited(&conf.Init.Retry),
+		defaultGasLimit: pldconf.DefaultDefaultGasLimit,             // can be set by config below
+		initRetry:       retry.NewRetryIndefinite(&conf.Init.Retry), // indefinite retry
 		name:            name,
 		api:             toDomain,
 		initDone:        make(chan struct{}),
--- a/core/go/internal/domainmgr/domain.go
+++ b/core/go/internal/domainmgr/domain.go
@@ -86,7 +86,7 @@ func (dm *domainManager) newDomain(name string, conf *pldconf.DomainConfig, toDo
 		dm:              dm,
 		conf:            conf,
 		defaultGasLimit: pldconf.DefaultDefaultGasLimit, // can be set by config below
-		initRetry:       retry.NewRetryIndefinite(&conf.Init.Retry),
+		initRetry:       retry.NewRetryLimited(&conf.Init.Retry),
 		name:            name,
 		api:             toDomain,
 		initDone:        make(chan struct{}),
--- a/config/pkg/pldconf/domainmgr.go
+++ b/config/pkg/pldconf/domainmgr.go
@@ -46,5 +46,5 @@ var ContractCacheDefaults = &CacheConfig{
 }
 
 type DomainInitConfig struct {
-	Retry RetryConfig `json:"retry"`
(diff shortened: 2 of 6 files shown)


core/go/internal/domainmgr/domain.go
```diff
@@ -86,7 +86,7 @@ func (dm *domainManager) newDomain(name string, conf *pldconf.DomainConfig, toDo
 		dm:              dm,
 		conf:            conf,
 		defaultGasLimit: pldconf.DefaultDefaultGasLimit, // can be set by config below
-		initRetry:       retry.NewRetryIndefinite(&conf.Init.Retry),
+		initRetry:       retry.NewRetryLimited(&conf.Init.Retry),
```
peterbroadhurst, line 89: This is a very significant functional change. Really need a clear description of everywhere this is used, and why a non-infinite retry is safe.

peterbroadhurst (changes requested):

core/go/internal/domainmgr/domain.go
dwertent: By default, the `Limited` parameter is set to `0` [when no explicit value is provided](https://github.com/LF-Decentralized-Trust-labs/paladin/blob/c1b5160704306e8529e1a59f261d42f3e7a35a74/toolkit/go/pkg/retry/retry.go#L55C2-L55C75), which is then [interpreted as infinite retries](https://github.com/LF-Decentralized-Trust-labs/paladin/blob/c1b5160704306e8529e1a59f261d42f3e7a35a74/toolkit/go/pkg/retry/retry.go#L70C31-L70C48). Therefore, this change does not affect the current behavior. As I explained in the PR, the reason...

peterbroadhurst: Going to take a look and understand the race condition in the test - we do need a test-only change here.

peterbroadhurst: I've proposed a commit directly to the branch, that:
- Updates some of the missing code comments that lead to confusion
- Splits the `DomainRegistered` function so the unit tests can call `UTSetMaxAttempts(1)` before the init() routine kicks off

Hope that addresses the timing condition in the test 🤞

jimthematrix (approved):

peterbroadhurst (approved):

PR #560, opened 2025-02-14
fix: Add write permissions to LOCALBIN in build process

dwertent:


hosie (approved):
LGTM and verified locally that it does fix https://github.com/LF-Decentralized-Trust-labs/paladin/issues/543

PR #561, opened 2025-02-14
feat: implement checksum for operator config

dwertent:
Adding a checksum to the operator annotations ensures that the operator restarts whenever the ConfigMap changes


onelapahead (approved):
#️⃣ ➕

PR #562, opened 2025-02-14
feat: Add action to update JSON file and modify SDK version in workflows

dwertent:
This will ensure the typescript-SDK is released with the same version as the helm-chart.


peterbroadhurst (approved):
So this means version control of the Typescript SDK is **not performed via source code checkin** and instead is performed at release time, by modifying the `package.json` on the fly.

Issue #555, opened 2025-02-07
Simplify Paladin Setup and Tutorials


dwertent:
# Overview  
Currently, users following the Paladin tutorials are required to clone and build the project locally, which has led to environment-related issues and inconsistencies. To streamline this process, we will **publish the Solidity artifacts as part of the GitHub release process**.  

This will allow users to **download the precompiled artifacts** instead of needing to build them manually, ensuring a smoother and more reliable tutorial experience.  

### Implementation Plan  
1. Publish Solidity artifacts as GitHub release assets
   - Every release will include the compiled Solidity artifacts.  
   - This ensures that users always work with the correct version of the contracts.  
2. Automate artifact publishing with GitHub Actions
   - Generate and upload Solidity artifacts as part of the release workflow.  
3. Update tutorial documentation  
   - Instruct users to **clone the latest stable tag** instead of `main`.  
   - Guide users to **download the Solidity artifacts** from the release page.  

This change will significantly simplify the tutorial setup while maintaining consistency across releases. 🚀

> This feature was proposed following discussions on [Discord](https://discord.com/channels/905194001349627914/1303371167020879903/1330166599864422513).

PR #568, opened 2025-02-18
Publishing of Solidity Artifacts in GitHub Releases

dwertent:
# Overview

As part of our effort to streamline the user experience with Paladin tutorials (#555 ), we want to publish Solidity artifacts as part of our GitHub release process. This PR implements the first step of our implementation plan:

## What this PR does

Automates the generation and publication of Solidity artifacts during the release process via GitHub Actions.
Ensures that every new release includes pre-compiled Solidity artifacts, making them readily available for users without the need for local compilation.

## Next Steps
Now, we need to update the tutorial documentation to reflect this change. 
We'll need to:
1. Instruct users to clone the latest stable tag rather than main.
2. Provide steps for downloading and using the Solidity artifacts from the release page.

# Release

Here are the released artifacts ([source](https://github.com/dwertent/paladin/releases/tag/v0.1.1)):
image

Here is the new flow ([source](https://github.com/dwertent/paladin/actions/runs/13396292675)):
image

### DIFF
--- a/.github/workflows/release-charts.yaml
+++ b/.github/workflows/release-charts.yaml
@@ -157,7 +157,7 @@ jobs:
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@v1.6.0
         with:
-          mark_as_latest: false # the release is marked as latest in the next step
+          mark_as_latest: false # the release job is marked as latest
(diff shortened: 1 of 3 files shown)


hosie (approved):
LGTM

PR #569, opened 2025-02-18
Infinite scrolling for registry

gabriel-indik:
The registry view was one of the first panels built. In its first implementation, it would display all entries of all registries. The query limit for each of the registries was set to 100. This worked fine for development / testing purposes where the number of entries in the registry tends to be small. However, if the UI is used in an environment where the number of entries in the registry is in the hundreds or thousands, the user won't be able to access the full list.

For this reason the following changes were made:
1. Instead of showing all entries from all registries, there is now a registry-selector control in the upper-left corner of the screen.
2. Once a registry is selected (if there is just one, it is auto-selected) the entries are shown using infinite scrolling
3. Toggle control has been added to view All | Active | Inactive registry entries
4. Filters have been added in order to allow the user to quickly locate entries based on id, name and owner
5. Filters are persisted using local storage
6. Registry entries are sorted by name on the server side

image


shorsher (approved):
📜 ♾️ 🚀

dwertent (approved):
Relying on @shorsher's approval

PR #578, opened 2025-02-20
Address fields in filters for tx

gabriel-indik:
It was observed that using JSON Query for transactions would result in either unexpected errors (unknown columns or invalid SQL statements). After examining the code it could be verified that the issue was simply due to the way the FieldMap was defined for the filters. Specifically reserved SQL words ("from", "to") need to be escaped. In addition, "domain" needs to be changed to "transactions.domain" to remove ambiguity. 
With the changes in place, the JSON queries run against transactions correctly return the expected results.

### DIFF
--- a/core/go/internal/txmgr/transaction_query.go
+++ b/core/go/internal/txmgr/transaction_query.go
@@ -37,9 +37,9 @@ var transactionFilters = filters.FieldMap{
 	"created":        filters.TimestampField("created"),
 	"abiReference":   filters.TimestampField("abi_ref"),
 	"functionName":   filters.StringField("fn_name"),
-	"domain":         filters.StringField("domain"),
-	"from":           filters.StringField("from"),
-	"to":             filters.HexBytesField("to"),
+	"domain":         filters.StringField(`transactions.domain`),
+	"from":           filters.StringField(`"from"`),
+	"to":             filters.HexBytesField(`"to"`),
 }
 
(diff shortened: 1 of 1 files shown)


peterbroadhurst:
I think this one is very significant architecturally based on your description.

We have a pattern for this @gabriel-indik - can you describe the alignment between the two areas of code?

https://github.com/LF-Decentralized-Trust-labs/paladin/blob/b9cfde1743d0fd59859bbb0bdce627d8320d9a7f/core/go/internal/statemgr/state.go#L205-L210

peterbroadhurst:
Or are these just SQL reserved words that need escaping - I might have misunderstood?

dwertent (dismissed):

gabriel-indik:
Details are included in https://github.com/LF-Decentralized-Trust-labs/paladin/issues/579.
With regards to the question above, this particular issue is caused by 2 items:
1. Columns that are SQL reserved words and need to be escaped (from, to)
2. A column that is ambiguous

For item 2, this is because the query joins the tables transactions and transaction_receipts, both of which have the `domain` column. Running the query works fine, but if there is a `WHERE` condition that involves the `domain` column, the database needs to know which column to do the comparison against. In other words, as long as there isn't a condition with that column, then the query runs fine, but if there is, the "ambiguous (SQLSTATE 42702)" error is returned.

One solution is to simply specify in the filters that when the user references the `domain` column, in the SQL query that should be `transactions.domain`.

peterbroadhurst (approved):
Thanks for the clarification - sorry I didn't get it first time round 🤦

PR #584, opened 2025-03-04
Fix postgres go migrations

alexwood:
- fix missing COMMIT in migrations file, caused error when running migrations.
- add `postgresql-client` to the runtime image so that it can be use to bootstrap the db and run the migrations for an external db using the paladin image


dwertent (dismissed):

alexwood:
So in the kaleido platform we manage the postgres database lifecycle separately to the runtime. 

Managing the database lifecycle. includes ensuring the database exists, ensuring we have run the migrations for version of the runtime. It also includes cleaning up the database  at the correct time in our lifecycle. (service deletion) 

To ensure the data bases exists (and to delete it) we use the PSQL client in combination with the specific credentials for the database and user, we need the same PSQL credentials to run the migrations. 

The pattern we use is running the runtime image (which contains the migrations) as a job to both ensure/migrate/cleanup the database in step with the lifecycle. 

All the existing images we use that also contain migrations have the PSQL client installed (inc firefly). However if this is an issue adding it to paladin oss dockerfile. We could do this differently for paladin. We will need to build paladin images in our pipeline anyway so we could override the base image or inject the client on top of the paladin image.

PR #585, opened 2025-03-04
allow custom resource labels for paladin cr

alexwood:
- so that we can plumb through our own labels to the generated resources for a paladin node
- which we need in order to make our custom network policies work


dwertent (changes requested):
I'm not sure if we want to include "deployment configurations" in the Paladin CRD definition. Please refer to #517 (this [comment](https://github.com/LF-Decentralized-Trust-labs/paladin/pull/517#issuecomment-2698049619) for more context).  

I understand this is blocking you. In the meantime, you can set the labels when installing the chart: [values.yaml#L97](https://github.com/LF-Decentralized-Trust-labs/paladin/blob/main/operator/charts/paladin-operator/values.yaml#L97).

PR #597, opened 2025-03-06
Reliable message peers must filter by node name when dispatching messages

awrichar:
Fixes https://github.com/LF-Decentralized-Trust-labs/paladin/issues/595


hosie (approved):
LGTM

PR #598, opened 2025-03-08
Add websocket client to SDK along with simple example

awrichar:


example/event-listener/src/index.ts
```diff
@@ -0,0 +1,147 @@
...
+  const wsClient = new PaladinWebSocketClient(
+    {
+      url: "ws://127.0.0.1:31549",
+      subscriptions: ["example-event-listener"],
+    },
+    async (sender, event) => {
+      if (event.method === "ptx_subscription") {
+        for (const receipt of event.params.result.receipts) {
+          await processReceipt(receipt);
+          sender.ack(event.params.subscription);
```
peterbroadhurst, line 101: Maybe worth a comment on the importance of this?

peterbroadhurst (approved):
This is great 🙌 

One small suggestion on calling out the `ack`.

example/event-listener/src/index.ts
awrichar: Yes, and I've actually put it in the wrong place... should be outside the loop.

PR #609, opened 2025-03-21
sdk: add methods for looking up privacy groups

awrichar:


sdk/typescript/src/paladin.ts
```diff
@@ -307,6 +307,22 @@ export default class PaladinClient {
     return res.data.result;
   }
 
+  async getPrivacyGroupById(domainName: string, id: string) {
```
peterbroadhurst, line 310: Looks like this is the non-lint change

peterbroadhurst (approved):

PR #611, opened 2025-03-31
Event listeners

annamcallister:
https://github.com/LF-Decentralized-Trust-labs/paladin/issues/581

Expose the ability to create an event listener externally. This is effectively wiring up the existing block chain internal event stream functionality (plus some lifecycle enhancements) with the existing websocket functionality.


core/go/internal/domainmgr/event_indexer.go
```diff
@@ -183,83 +183,84 @@ func (d *domain) batchEventsByAddress(ctx context.Context, dbTX persistence.DBTX
...
 
-func (d *domain) handleEventBatch(ctx context.Context, dbTX persistence.DBTX, batch *blockindexer.EventDeliveryBatch) error {
-
-	// First index any domain contract deployments
-	nonDeployEvents, txCompletions, err := d.dm.registrationIndexer(ctx, dbTX, batch)
-	if err != nil {
-		return err
-	}
+func (d *domain) handleEventBatch(ctx context.Context, batch *blockindexer.EventDeliveryBatch) (func(dbTX persistence.DBTX) error, error) {
+	return func(dbTX persistence.DBTX) error {
```
annamcallister, line 187: There's quite a few changes like this which due to alphabetical ordering appear at the top of the PR but the reason isn't apparent until much further down. This comment is probably the most helpful in explaining why https://github.com/LF-Decentralized-Trust-labs/paladin/blob/72e24d40b31042802afdc6dec229788cff6755e6/core/go/pkg/blockindexer/event_stream_types.go#L127

core/go/internal/txmgr/blockchain_event_listeners.go
```diff
@@ -0,0 +1,369 @@
...
+}
+
+// A blockchain event listener is a wrapper which externalises a blockchain internal event stream.
+// All persistence belongs to the internal event stream.
+// This layer manages
+//   - mapping between "ptx_<verb>BlockchainEventListener" RPC calls and internal event stream
+//     lifecycle functions
+//   - dispatching event batches to registered receivers
+
+type blockchainEventListener struct {
```
annamcallister, line 50: While we've been talking about `event listeners`, I went for `blockchainevent` rather than just `event` as it feels like there might be other types of event to come beyond on chain events for public transactions (e.g. events from domains?)

core/go/pkg/blockindexer/block_indexer.go
```diff
@@ -163,23 +167,6 @@ func (bi *blockIndexer) Start(internalStreams ...*InternalEventStream) error {
 	return nil
 }
 
-func (bi *blockIndexer) AddEventStream(ctx context.Context, dbTX persistence.DBTX, stream *InternalEventStream) (*EventStream, error) {
```
annamcallister, line 166: moved to the event streams file as it felt like this function belonged there with the other event stream lifecycle functions

core/go/internal/domainmgr/event_indexer.go
peterbroadhurst: Having had a bit more of an understanding of the problem you're trying to solve here now @annamcallister, I do not think this is the function you should be changing.

Instead, I'm currently thinking you should:
- Add a new option to the **internal code** version event streams to disable DB transactions
   - Maybe in here: https://github.com/LF-Decentralized-Trust-labs/paladin/blob/75dc7b79da059d4a6d567d0a8a7842d7f08b1022/core/go/pkg/blockindexer/event_stream_types.go#L138-L143
- Use this mode switch choose whether this function uses a DB transaction or not:
    - This `runBatch` function: https://github.com/LF-Decentralized-Trust-labs/paladin/blob/75dc7b79da059d4a6d567d0a8a7842d7f08b1022/core/go/pkg/blockindexer/event_streams.go#L489-L494
- Make it so the JSON/RPC attached event streams set the mode to non-transaction (but all the existing code that is internal to Paladin, and designed requiring the DB TX continues to use the TX)

core/go/internal/domainmgr/event_indexer.go
```diff
@@ -183,83 +183,84 @@ func (d *domain) batchEventsByAddress(ctx context.Context, dbTX persistence.DBTX
...
+		// First index any domain contract deployments
```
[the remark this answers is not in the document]
annamcallister: I only modified the white space around this line. This file is no longer part of the PR.

core/go/internal/domainmgr/event_indexer.go
peterbroadhurst: Closing this out thanks @annamcallister 

I see you addressed the feedback by splitting out the two types of callback on the interface here:
https://github.com/LF-Decentralized-Trust-labs/paladin/blob/6489d6330562c1c292f6358a956f1e7c49f08c60/core/go/pkg/blockindexer/event_stream_types.go#L127-L147

core/go/internal/txmgr/blockchain_event_listeners.go
peterbroadhurst: Yes - I think this is probably the hardest spelling conversation on this PR.

I'm feeling good about `blockchainevents` being the term for these, but in getting there I want to highlight some of the worries I had to work through (and would like @awrichar's validation on):

- These listener do **not** cover the events from EVM privacy groups (directly at least)
    - Logically of course these are blockchain events, but my feeling was we already use the prefix `domainreceipt`, and thus if we were to directly create a listener for these we'd put the word `domain` in there
- We've got the word `publicTx` in various places, and I thought for a minute if `public` should be here too
    - I just couldn't find a way to fit sensibly. Everything I tried seemed wrong. I didn't feel`publicTxEvents` for example made sense.
- I noticed we have the RPC event listener on `ptx`, and maybe it would be more obvious if it was on `blockindexer`
   - But to decode the events you need the ABI capability that's in the `ptx` realm, and practically it would mean a whole bunch more code to add a listener into the `blockindexer` section

core/go/internal/componentmgr/manager.go
```diff
@@ -315,6 +315,9 @@ func (cm *componentManager) startBlockIndexer() (err error) {
 		_, err = cm.blockIndexer.GetBlockListenerHeight(cm.bgCtx)
 		err = cm.wrapIfErr(err, msgs.MsgComponentBlockIndexerStartError)
 	}
+	if err == nil {
+		err = cm.txManager.LoadBlockchainEventListeners()
```
peterbroadhurst, line 319: Looking into the code I see why you made this a whole separate lifecycle (rather than just make these new returned `internalEventStreams`).

Overall this function is a bit clunky and ❄️ in the component initialization - but that was true before your addition.

core/go/pkg/blockindexer/event_streams.go
```diff
@@ -490,10 +618,18 @@ func (es *eventStream) runBatch(batch *eventBatch) error {
...
+		}
+		err = es.bi.persistence.Transaction(es.ctx, func(ctx context.Context, dbTX persistence.DBTX) (err error) {
+			if !es.useNOTXHandler {
+				err = es.handlerDBTX(ctx, dbTX, &batch.EventDeliveryBatch)
+				if err != nil {
+					return err
+				}
 			}
 			// commit the checkpoint
 			return dbTX.DB().
```
peterbroadhurst, line 635: I don't think there's any need to use a DB transaction here for the checkpoint in the else case.

I wonder if it makes sense to:
- Move the `event_stream_checkpoints` code to an `updateCheckpoint` function
- Have code like:

```go
func (es *eventStream) runBatch(batch *eventBatch) error {

	return es.bi.retry.Do(es.ctx, func(attempt int) (retryable bool, err error) {
		if es.useNOTXHandler {
			// Call the handler outside of any transaction, and update the checkpoint
			err = es.handlerNOTX(es.ctx, &batch.EventDeliveryBatch)
			if err == nil {
				err = es.updateCheckpoint(...)
			}
			return true, err
		}
		// Otherwise start a database transaction, run the callback function
		err = es.bi.persistence.Transaction(es.ctx, func(ctx context.Context, dbTX persistence.DBTX) (err error) {
			if !es.useNOTXHandler {
				err = es.handlerDBTX(ctx, dbTX, &batch.EventDeliveryBatch)
				if err != nil {
					return err
				}
			}
			// commit the checkpoint
			return es.updateCheckpoint(...)
		})
		return true, err
	})

}
```

core/go/internal/txmgr/blockchain_event_listeners.go
awrichar: I'm definitely not plugged in to the full context here, but the reasoning above seems good, and I think "blockchain events" sounds explainable/understandable.

core/go/pkg/blockindexer/event_streams.go
peterbroadhurst: Thanks for making the update there @annamcallister 👍

peterbroadhurst (approved):

PR #619, opened 2025-04-02
Package refactor

annamcallister:
_Description updated after PR review_

Split `toolkit` package into ~`sdk` and `toolkit`~`common`, `sdk`  and `toolkit` so that `sdk` is importable with go modules.

`common`: packages used all over the codebase, including by `sdk` but which we don't expect to get external usage
-  `i18n`
- `tkmsgs` - renamed to `pldmsgs` now it's not in `toolkit`
- `log`

`sdk`: packages a consumer needs to use `pldclient`
- `pldapi`
- `tktypes` - renamed to `pldtypes` now it's not in `toolkit`
- `query`
- `tlsconf`
- `rpcclient` (I wouldn't expect this to get externally consumed but we have `pldclient -> tlsconf` and `pldclient -> rpcclient -> tlsconf` so moving it elsewhere isn't going to work)

`toolkit`: everything else (still a bit of a mixed bag)

* ~`toolkit` contains `prototk` and everything that depends on it- i.e. it has become a package for developing domains/plugins~
* ~`sdk` contains everything else (this is quite a mixture of things, and not all things we would expect an external consumer to use, but I don't know that now is the time to engage people on how we could improve this)~

While declaring all dependencies fully so that go module imports work, I found a circular dependency between `toolkit`/`sdk` packages and `config`, so I have completely removed the `config` dependency on those packages.


annamcallister:
The reasoning was a combination of using "depends on prototk" as a decision filter for where packages went, plus `pldclient` using `rpcserver` for tests which also brings in indirect dependencies on `router`, `staticserver` and `httpserver` amongst other things.

I've been able to remove the `pldclient` dependency on `rpcserver`, so we now have

`common`: packages used all over the codebase, including by `sdk` but which we don't expect to get external usage
-  `i18n`
- `tkmsgs`
- `log`

`sdk`: packages a consumer needs to use `pldclient`
- `pldapi`
- `tktypes`
- `query`
- `tlsconf`
- `rpcclient` (I wouldn't expect this to get externally consumed but we have `pldclient -> tlsconf` and `pldclient -> rpcclient -> tlsconf` so moving it elsewhere isn't going to work)

`toolkit`: everything else (still a bit of a mixed bag)

I'll open a follow on PR to rename `tkmsgs` and `tktypes` to drop the `tk` now they're not in the toolkit but I think there's enough renaming here for one PR.

peterbroadhurst (approved):
Thanks for going the extra mile on that - this really is a big improvement in the structure

PR #625, opened 2025-04-14
always set last submit time after submission

annamcallister:


Chengxuan (approved):
As LastSubmit time is used for resubmission check, it should be updated for all submission attempts that resulted in categorized expected errors.

PR #626, opened 2025-04-15
Stop pulling all blocks to catch up with into memory

annamcallister:
The block indexer dispatcher loop is reading all blocks into an unbounded array in memory until it has caught up with the head of the chain. Pulling more blocks into memory at a time doesn't speed anything up as the enrichment is only being triggered for one block each iteration of the loop, and that process is only starting after all the blocks have been read into memory.

This PR changes it so that one block is read per loop iteration, and the enrichment started immediately for each block. Batches can be written as soon as max batch size is reached.

I've also included some trace logs that I added while debugging so they are there in future.

### DIFF
--- a/core/go/pkg/blockindexer/block_indexer.go
+++ b/core/go/pkg/blockindexer/block_indexer.go
@@ -338,9 +338,11 @@ func (bi *blockIndexer) notificationProcessor(ctx context.Context) {
 // Then we update the state the dispatcher uses to walk forwards from and see what
 // is confirmed and ready to dispatch
 func (bi *blockIndexer) processBlockNotification(ctx context.Context, block *BlockInfoJSONRPC) {
-
 	bi.stateLock.Lock()
 	defer bi.stateLock.Unlock()
+
+	log.L(ctx).Tracef("<processBlockNotification> block.Number:%d, bi.nextBlock:%d, len(bi.newHeadToAdd):%d, len(bi.blocksSinceCheckpoint):%d",
+		block.Number, *bi.nextBlock, len(bi.newHeadToAdd), len(bi.blocksSinceCheckpoint))
 
 	// If the block is before our checkpoint, we ignore it completely
 	if block.Number < *bi.nextBlock {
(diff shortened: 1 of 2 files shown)


core/go/pkg/blockindexer/block_indexer.go
```diff
@@ -438,32 +448,26 @@ func (bi *blockIndexer) dispatcher(ctx context.Context) {
...
-	for {
-		timeoutContext := ctx
 
-		if len(pendingDispatch) == 0 {
-			pendingDispatch = nil // ensure we clear the memory if we just looped through a set with pendingDispatch[1:] below
+	timeoutContext := ctx
 
-			// spin getting blocks until we it looks like we need to wait for a notification
-			lastFromNotification := false
-			for bi.readNextBlock(ctx, &lastFromNotification) {
```
peterbroadhurst, line 451: I agree I can't see as I read and re-read this code any reason for this inner loop.
The batching happens in the outer loop.

However, I'm not yet convinced I understand why the `lastFromNotification` boolean has been removed.

The purpose of this boolean was to track if we were popping things off of the front of the `bi.newHeadToAdd` queue, as there's no point in doing an expensive query out to see if there are more blocks (because we've got a list of blocks already).

I do realize this logic is quite complex, and it would take me significantly more time to propose structural simplications.

But for now, is it possible that removing this inner tight loop (while leaving the `&lastFromNotification` optimization in place) will solve the problem?

Or is there a reason both changes are needed?

annamcallister: Ah ok I now see the extra value that provides. I got caught up with tests which got stuck because they don’t simulate the notification that causes that bool to be set and it was all looking equivalent to me.

I've pushed a change which adds it back in. The value now needs to persist over the outer loop iterations but gets set back to false once no more transactions are found- which is the same behaviour as the old code, just in a different place now.

peterbroadhurst (approved):
Thanks for diving into his complex (possibly over complex 😬 ) code here to get to root cause.

PR #631, opened 2025-04-22
zeto: add user-facing "data" that can be passed through transfers

awrichar:
Allows a user/application to annotate a mint or transfer with extra data of their own.

Mimics the same function from Noto.

Also contains collapsed functionality from #633, #634, #646.


domains/zeto/pkg/types/states.go
```diff
@@ -94,6 +94,38 @@ var ZetoNFTokenABI = &abi.Parameter{
...
+	Name:         "TransactionData",
+	Type:         "tuple",
+	InternalType: "struct TransactionData",
+	Components: abi.ParameterArray{
+		{Name: "salt", Type: "bytes32"},
+		{Name: "data", Type: "bytes"},
+	},
+}
+
+// TODO: is this an appropriate hashing method for this state data?
```
awrichar, line 113: Open question here. Zeto has chosen to provide its own state hashes, so it must hash the transaction metadata as well. I put in a simple SHA-256 hash, but I'm not sure this is actually a correct choice.

solidity/contracts/domains/interfaces/IZetoFungible.sol
```diff
@@ -10,6 +10,7 @@ interface IZetoFungible {
     struct TransferParam {
         string to;
         uint256 amount;
+        bytes data;
```
awrichar, line 13: I chose to put the `data` inside this struct, so it is "per transfer" rather than "per transaction" - since all Zeto methods are built for batching.

However, because we ultimately just end up with a list of input, output, and info states, it will be difficult to "match" transfers and data when parsing events.

domains/zeto/pkg/types/config.go
```diff
@@ -91,7 +92,17 @@ var DomainInstanceConfigABI = &abi.ParameterArray{
 }
 
 // marks the version of the Zeto transaction data schema
-var ZetoTransactionData_V0 = ethtypes.MustNewHexBytes0xPrefix("0x00010000")
+var ZetoTransactionDataID_V0 = ethtypes.MustNewHexBytes0xPrefix("0x00010000")
```
awrichar, line 95: You could argue that this should be incremented, since I have changed the data format. However, I'm not sure how we feel about backwards compatibility at this staged.

jimthematrix (approved):
LGTM.

The sha256 based hash for the info state will be handled differently from the Poseidon based hash for the coin states. so not sure if your open question on this point needs a resolution for the PR to merge (I thought there shouldn't be a problem with the current approach of using different hashing algos, because the Poseidon hash is pretty specialized due to the usage of ZKPs)

domains/zeto/pkg/types/abi.go
```diff
@@ -61,13 +61,15 @@ type NonFungibleMintParams struct {
...
 }
+
 type FungibleTransferParams struct {
 	Transfers []*FungibleTransferParamEntry `json:"transfers"`
 }
 
 type FungibleTransferParamEntry struct {
 	To     string               `json:"to"`
 	Amount *pldtypes.HexUint256 `json:"amount"`
+	Data   pldtypes.HexBytes    `json:"data"`
```
peterbroadhurst, line 72: If I understand correct, this change will be a breaking API change as ABI does not have the concept of optional parameters.

domains/zeto/pkg/types/states.go
peterbroadhurst: Why would the same algorithm not be used?

I assume for this PR to be complete (and consistent with Noto), the Zeto transaction proofs must include validation of the hash?

awrichar: I don't think Poseidon is great for arbitrary sequences of bytes. It's more for fixed length inputs, and chosen for ease of use in ZKPs (which I don't think is applicable for this piece of data).

awrichar: I am implying that the data is passed "in the clear" on chain and could be validated by the receiver, but is not obscured or validated by a ZKP.

domains/zeto/pkg/types/abi.go
awrichar: Correct

domains/zeto/pkg/types/states.go
peterbroadhurst: Note discovered that we didn't update `ValidateStateHashes()` in back and forth in testing - adding that as part of validating your change.

peterbroadhurst: > I am implying that the data is passed "in the clear" on chain and could be validated by the receiver, but is not obscured or validated by a ZKP.

I think you mean the salted-hash of the data (no not leaking anything) as the data itself is passed private.

Make sense as I think about it that no ZKP is needed for this, as there is no processing of the data to validate (such as total conservation of value etc.)

awrichar: Thanks for flagging (and fixing) the validation of the state hashes.

And yes you're completely right on what's passed on chain: it's just the data hashes, not the full contents of the data - and it's passed to the smart contract only, but not incorporated into any proof.

awrichar:
Note: build currently failing due to coverage drop in Zeto.

solidity/contracts/domains/interfaces/IZetoFungible.sol
awrichar: I spent some more time thinking about this and considering the alternate option (placing a single `data` outside the list of transfers). Ultimately that option didn't feel better. State distribution was a little fuzzy - in the current model it's clear that the "data" states get distributed to the same recipients as the "coin" states, but with a single "data" state, it has to be distributed to all recipients that received a coin as part of the transfer.

That isn't necessarily wrong, but I couldn't convince myself it was better than what is here now.

PR #633, opened 2025-04-23
Add RPC module and basic UI for domain manager

awrichar:
This is quite barebones at the moment, but it's a start.

**RPCs**
- `domain_listDomains`
- `domain_getDomain`
- `domain_getDomainByAddress`
- `domain_querySmartContracts`
- `domain_getSmartContractByAddress`

**Screenshots**

Paladin

Paladin


ui/client/src/components/DomainButtons.tsx
```diff
@@ -0,0 +1,89 @@
...
+}) => {
+  const { t } = useTranslation();
+  const [buttons, setButtons] = useState<DomainButton[]>([]);
+  const [zetoMintDialogOpen, setZetoMintDialogOpen] = useState(false);
+  const [zetoTransferDialogOpen, setZetoTransferDialogOpen] = useState(false);
+
+  useEffect(() => {
+    const tmpButtons: DomainButton[] = [];
+
+    // TODO: should key off of the domain "type" instead of expecting a specific name
```
peterbroadhurst, line 45: This is an interesting observation. I guess we don't really have a concept like this, so it can't be solved in the UI alone.

peterbroadhurst (approved):
The JSON/RPC stuff looks great, and the initial UI panels are definitely a useful tool.

I feel like @gabriel-indik might have thoughts on wider improvements that could be made to the panel (both in terms of formatting/consistency and content), but those improvements can be made on top of this as the code seems clean.

PR #635, opened 2025-04-24
Event stream enhancements

annamcallister:
Adds the ability to specify a `fromBlock` when creating an event listener
* a specific block number
* `"latest"`

The default value is `0` which is the same as the existing behaviour: include events from all blocks that the block indexer has indexed.

Adds a status API for event listeners which returns the checkpoint block number.


core/go/pkg/blockindexer/block_indexer_test.go
```diff
@@ -1154,11 +1154,3 @@ func TestQueryNoLimit(t *testing.T) {
 	_, err = bi.QueryIndexedEvents(ctx, query.NewQueryBuilder().Query())
 	assert.Regexp(t, "PD011311", err)
 }
-
-func TestAddEventStreamBadName(t *testing.T) {
```
annamcallister, line 1158: moved to `event_streams_test.go`

core/go/pkg/blockindexer/event_stream_types.go
```diff
@@ -32,13 +33,15 @@ import (
 )
 
 type EventStreamConfig struct {
-	BatchSize    *int    `json:"batchSize,omitempty"`
-	BatchTimeout *string `json:"batchTimeout,omitempty"`
+	BatchSize    *int            `json:"batchSize,omitempty"`
+	BatchTimeout *string         `json:"batchTimeout,omitempty"`
+	FromBlock    json.RawMessage `json:"fromBlock,omitempty"`
```
peterbroadhurst, line 38: We use `pldtypes.RawJSON` in Paladin, which has some extensions on top of `json.RawMessage` (such as DB serialization etc.)

core/go/pkg/blockindexer/event_streams.go
```diff
@@ -52,6 +57,8 @@ type eventStream struct {
 	serializer     *abi.Serializer
 	detectorDone   chan struct{}
 	dispatcherDone chan struct{}
+	fromBlock      *int64       // nil == latest
```
peterbroadhurst, line 60: So one thing that's been a bit painful in the evolution of this code through the generations, is consistency / code-clarity in how `0` is handled.

I see below that `-1` is what's used to describe "the next block to deliver is block 0" (as we take `*es.flomBlock - 1`. Could you confirm you did this because that's what the other code in blockindexer does?

Bit of laziness on my part here, because I remember wrestling with this in blockindexer to try and avoid some of the weird signed/unsigned number conversion (which generates compiler warnings - for good reasons). But I can't remember which way we went there.

I'd like to make sure we're consistent.

peterbroadhurst (commented):
Couple of questions around code consistency I'd like to make sure we think through before we merge this one please @annamcallister 

Suspect I'm pointing out stuff where you've found an inconsistency elsewhere, but I'd like to make sure we're using the opportunity of fresh eyes on this code to make our best gut on the right choice (and if it's inconsistent with other places, thinking now might be the time to fix the other places too).

core/go/pkg/blockindexer/event_stream_types.go
annamcallister: I thought I had done a direct copy... that might have been a co pilot chipping in to "help"...

core/go/pkg/blockindexer/event_streams.go
annamcallister: > Could you confirm you did this because that's what the other code in blockindexer does?

Yes my intention was to keep the handling of `fromBlock` between them as consistent as possible.

I will double check that I have actually done that and not had the code evolve in a different direction though.

core/go/pkg/blockindexer/event_stream_types.go
annamcallister: No it's `json.RawMessage` in the block indexer too https://github.com/LF-Decentralized-Trust-labs/paladin/blob/main/config/pkg/pldconf/blockindexer.go#L26 - I'll try and update them both to use `pldtypes.RawJSON`

annamcallister: I thought I had it sorted... then I realised that having the block indexer use `pldtypes` brings back in the circular dependency between `config` and `sdk` I spent all the time unpicking the other week 😭

annamcallister: I can make this work if we do the following package reorg:
- move `log` out of `common` into it's own package
- move `pldtypes` out of `sdk` into either its own package or `common` (I'm not keen on putting it in common because it then becomes a mixture of things we'd expect to be consumed internally and externally)

Then we can have `sdk` -> `log` -> `config` -> `pldtypes` -> `common`

I feel this could be done in a separate PR

peterbroadhurst (approved):

PR #639, opened 2025-04-24
Dedup requested verifier resolutions from domain responses

jimthematrix:
This was believed to be responsible for the intermittent zeto minting errors seen in the builds.


peterbroadhurst (approved):
Thanks @jimthematrix - looks great

core/go/internal/privatetxnmgr/transaction_flow_actions.go
```diff
@@ -653,6 +653,9 @@ func (tf *transactionFlow) requestVerifierResolution(ctx context.Context) {
 	if tf.transaction.PreAssembly.Verifiers == nil {
 		tf.transaction.PreAssembly.Verifiers = make([]*prototk.ResolvedVerifier, 0, len(tf.transaction.PreAssembly.RequiredVerifiers))
 	}
+	// having duplicate requests for the same verifier can cause the same transaction to be sent multiple times
+	tf.transaction.PreAssembly.RequiredVerifiers = dedupResolveVerifierRequests(tf.transaction.PreAssembly.RequiredVerifiers)
```
peterbroadhurst, line 657: Just one thought before you merge @jimthematrix - rather than modifying in-place, maybe we should use a new variable for this?

PR #643, opened 2025-04-30
Include the indexed block in an indexed event dispatched on an event stream

annamcallister:
The indexed block includes useful information like the block timestamp.

Also added in changes to accept an eth address on transaction submission as well as an identifier.

Also added in changes to not make a paladin copy of code from firefly-common


sdk/go/pkg/rpcclient/config.go
```diff
@@ -74,12 +74,26 @@ func ParseHTTPConfig(ctx context.Context, config *pldconf.HTTPClientConfig) (*re
 	restyConf := ffresty.Config{
```
peterbroadhurst, line 74: @annamcallister  - not sure the use of `ffresty` is really useful, it's a bit of lingering DNA from code that was used as an accelerator.

Probably the right thing here (consistent with the direction of travel elsewhere in the codebase) is to take a copy of any parts of the FireFly common Apache 2 `ffresty` util that is genuinely useful to Paladin into a package inside of Paladin an completely decouple it.

That's what we've done with a number of other places that had some original DNA - including the `rpclient` module itself.

annamcallister: Agree with taking a copy of the what is useful.

I think that's a separate question though to "do we want the full set of configuration on that code?" (to which I believe the answer is yes- I don't think we need to restrict the configuration options to just a subset).

annamcallister: I've removed the commits which added in all the configuration options.

Instead I've made a Paladin copy of `ffresty` and `wsclient`, which directly use the `pldconf` config structures. These have been trimmed down where they support config options which aren't currently supported in Paladin.

`rpcclient` already had an option to wrap an existing resty client. I've made a change to also expose this one level higher in `paladinclient`. 

I'm going to look at adding in more config options now; however, with the option to wrap an existing resty client, I don't actually have a concrete use case that is driving adding more config options, so I will time box this as I don't think anyone is calling it out as a priority.

core/go/internal/keymanager/key_resolver.go
```diff
@@ -233,8 +233,15 @@ func (kr *keyResolver) resolveKey(ctx context.Context, identifier, algorithm, ve
 	kr.l.Lock()
 	defer kr.l.Unlock()
 
-	// Identifier must be a valid
-	if err := pldtypes.ValidateSafeCharsStartEndAlphaNum(ctx, identifier, pldtypes.DefaultNameMaxLen, "identifier"); err != nil {
+	// Identifier must be valid and not an eth address
```
peterbroadhurst, line 236: @annamcallister - can you share the thinking here?

Not sure I disagree with it, but it seems very architecturally significant on the signing side (unrelated to the topic).

e.g. it seems related to if I want to directly use the `0x` address of my keys to resolve them, and the chain of configuration between my application, and my backend storage system (file system with key files... or advanced plugged in key manager system with keys stored by `name` like AWS/Azure/GCP Cloud HSMs etc.).

sdk/go/go.mod
```diff
@@ -8,8 +8,9 @@ require (
 	github.com/aidarkhanov/nanoid v1.0.8
 	github.com/go-resty/resty/v2 v2.14.0
 	github.com/google/uuid v1.6.0
-	github.com/hyperledger/firefly-common v1.5.4
```
peterbroadhurst, line 11: 🥳  very cool

sdk/go/pkg/wsclient/wsclient.go
peterbroadhurst, line 1: As one last feature from the FireFly DNA, you might consider pulling in this recent PR (not necessarily in this PR): https://github.com/hyperledger/firefly-common/pull/180

It might allow some code consolidation, because it allows a **server** websocket to inherit the same heartbeating code as the client, by doing a `Wrap()` in the `wsclient` interface.

Given this WebSocket heartbeating code is so complex (and unsolved in the underlying gorilla lib), it might be nice to have just one copy of it.

peterbroadhurst: Specifically this could become a `wsclient.WSClient` (or maybe the package name could change in Paladin in a way it's harder to do in `firefly-common` to something like `wsconn.WSConn`), and heartbeating would get handled automatically: https://github.com/LF-Decentralized-Trust-labs/paladin/blob/ba21b3c46ed03f242fee85f2850390fcd1042b84/toolkit/go/pkg/rpcserver/websockets.go#L66

peterbroadhurst (commented):
Couple of thoughts here on the WebSocket client interface, and on the change to validation of key identifiers to decide architecturally if we prevent key identifiers from being named after the EthAddress of the key.

I don't know if you'd prefer to handle them in a single PR, or split out.

annamcallister:
Closing as this PR has been split into
#658
#659
#660

Issue #616, opened 2025-04-01
bug: Zeto_Anon.json is missing


loocurse:
### What happened?

I am trying to run the helloworld example. Getting issues with zeto anon not being found.

In sdk/typescript I ran
```
npm run abi

> @lfdecentralizedtrust-labs/paladin-sdk@0.0.0 abi
> node scripts/abi.mjs

file:///paladin/sdk/typescript/node_modules/copy-file/fs.js:32
        throw new CopyFileError(`lstat \`${path}\` failed: ${error.message}`, {cause: error});
              ^

CopyFileError: lstat `../../domains/zeto/zkp/artifacts/contracts/zeto_anon.sol/Zeto_Anon.json` failed: ENOENT: no such file or directory, lstat '../../domains/zeto/zkp/artifacts/contracts/zeto_anon.sol/Zeto_Anon.json'
    at file:///.paladin/sdk/typescript/node_modules/copy-file/fs.js:32:8
    at async copyFile (./paladin/sdk/typescript/node_modules/copy-file/index.js:37:16)
    at async file://paladin/sdk/typescript/scripts/abi.mjs:23:1 {
  errno: -2,
  code: 'ENOENT',
  syscall: 'lstat',
  path: '../../domains/zeto/zkp/artifacts/contracts/zeto_anon.sol/Zeto_Anon.json',
  [cause]: [Error: ENOENT: no such file or directory, lstat '../../domains/zeto/zkp/artifacts/contracts/zeto_anon.sol/Zeto_Anon.json'] {
    errno: -2,
    code: 'ENOENT',
    syscall: 'lstat',
    path: '../../domains/zeto/zkp/artifacts/contracts/zeto_anon.sol/Zeto_Anon.json'
  }
}

Node.js v20.19.0
```

### What did you expect to happen?

Zeto_Anon.json to be found. I can't find any file with that too

### How can we reproduce it (as minimally and precisely as possible)?

Go to hello world example readme, follow steps until the `npm run abi` in sdk/typescript

### Anything else we need to know?

I must be missing something

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
sw_vers
ProductName:            macOS
ProductVersion:         15.3.2
BuildVersion:           24D81
$ uname -a
[...]

PR #645, opened 2025-04-30
Update zeto example setup scripts and readme

jimthematrix:
Fixes issue #616 
Based on user report.

Updated the setup scripts in
- `sdk/typescript`
- `example/zeto`

so that only `npm` commands are required to get all the dependency artifacts, including downloading the zeto contract ABIs from the zeto releases.

Still kept the Gradle build as an option.


example/zeto/README.md
```diff
@@ -2,35 +2,85 @@
...
-See the [tutorial](https://lf-decentralized-trust-labs.github.io/paladin/head/tutorials/zeto/) for a detailed explanation.
+See the [tutorial](https://lf-decentralized-trust-labs.github.io/paladin/head/tutorials/zkp-cbdc/) for a detailed explanation.
 
 ## Pre-requisites
 
-Requires a local 3-node Paladin cluster running on `localhost:31548`, `localhost:31648`, and `localhost:31748`.
+### Local Paladin network
 
-## Run standalone
+This sample requires a local 3-node Paladin network running on `localhost:31548`, `localhost:31648`, and `localhost:31748`. You can get one setup by running the operator build:
```
awrichar, line 11: So are we now asserting that you cannot run the tutorials using a network configured as per the [Getting Started](https://lf-decentralized-trust-labs.github.io/paladin/head/getting-started/installation/) guide, but must build and deploy the operator from source?

That would be a significant pivot. We've seen many people want to play with the examples without taking on the burden of a whole source build.

jimthematrix: Oh I forgot about that, yes that should still be the primary way to get the local network setup. thanks for the feedback

example/zeto/README.md
```diff
@@ -2,35 +2,80 @@
...
+
+- IZetoFungible.json: private transaction ABI for the Zeto domain, to conduct private transactions like `mint`, `transfer`, `lock`, etc.
+- Zeto_Anon.json: public transaction ABI for the Zeto token implementation, to conduct public transactions "delegateLock`
+- SampleERC20.json: public transaction ABI for a sample ERC20 token, to conduct public transactions to mint ERC20 tokens
+
+These dependency resources can be obtained by running the following builds:
+
+```console
+$ ./gradlew :sdk:typescript:build
+$ ./gradlew :example:zeto:build
```
awrichar, line 39: You don't need to run both (the example build depends on the SDK build so it will trigger everything).

This is less of a pivot than the first pass, but is still asserting that you need to get the Gradle build working (even though you don't need to build the entire tree).

jimthematrix: this has now been updated to eliminate the need to run gradle at all. but instead only `npm run` from inside `sdk/typescript/scripts` and `example/zeto/scripts`, similar to the flow for the Noto example.

peterbroadhurst (approved):
I did `git clean -xdf` in my paladin to clean up everything that isn't checked in, then ran through the (non gradle) steps E2E - worked a charm.

Great update, thanks @jimthematrix

PR #648, opened 2025-05-01
Column alignment for domains

gabriel-indik:
Very minor and optional suggestion: update the alignment of the actions column. This is so that the table in the new Domains section appears more balanced.

image


awrichar:
Honestly... I'm not sure I love it. I agree there is too much blank space in the current design. But shifting the blank space from the right to the middle doesn't seem much better (to me). It still looks empty and the buttons are now further away from the address. I think we'll have to consider what else we could populate to fill some space, or think about another way to display these longterm.

gabriel-indik:
Agreed, this was not meant as a final solution, but something to make the contents of the table appear horizontally distributed while we think of a better way to show the information on screen. Maybe a table isn't the right component.

PR #653, opened 2025-05-15
Add RPC Client Cache to Prevent Operator OOMKill

dwertent:
## Overview

The operator was previously creating a **new HTTP client and transport** on every reconcile via `pldclient.New().HTTP()`. This led to:

* Thousands of lingering goroutines (`net/http.(*persistConn).readLoop` / `writeLoop`)
* Unbounded memory growth and GC pressure
* Eventual **OOM kills** due to connection buildup

## Solution

This PR introduces an in-memory **cache (`rpcCache`)** to store and reuse `PaladinClient` instances, keyed by the Paladin endpoint and timeout.
 
**Before:**
image

**After:**
image

I used [Outrig](https://github.com/outrigdev/outrig) for the graphs


operator/internal/controller/paladinregistration_controller.go
```diff
@@ -64,7 +64,8 @@ var registryABI = abi.ABI{
 // PaladinRegistrationReconciler reconciles a PaladinRegistration object
 type PaladinRegistrationReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme   *runtime.Scheme
+	RPCCache *rpcCache
```
onelapahead, line 68: nit: `RPCClientManager` ? Cache is a bit odd term for a map of HTTP clients, and implies expiry/LRU/etc which isn't implemented (yet).

operator/internal/controller/transaction_reconcile.go
```diff
@@ -215,9 +230,40 @@ func getPaladinRPC(ctx context.Context, c client.Client, nodeName, namespace str
...
+	cache.mux.Lock()
+	defer cache.mux.Unlock()
+
+	// Check again in the cache after acquiring the lock
+	// This is to ensure that another goroutine didn't create the client while we were waiting for the lock
+	if client, ok := cache.clients[key]; ok && client != nil {
+		return client, nil
+	}
+
+	client, err := pldclient.New().HTTP(ctx, &pldconf.HTTPClientConfig{
```
onelapahead, line 258: I guess - the operator creating a new client/transport every reconcile is really inefficient but that doesn't fully explain the memory pressure right ?

There must be something holding a reference in memory to this client after the reconcile loop that keeps it around ? Is that a bug in `pldclient` itself ? Bc that seems odd.

Still - a client manager in the operator to track clients per Paladin's is a great/defensive optimization either way.

operator/internal/controller/transaction_reconcile.go
```diff
@@ -215,9 +230,40 @@ func getPaladinRPC(ctx context.Context, c client.Client, nodeName, namespace str
...
 		URL:               url,
 		ConnectionTimeout: confutil.P(timeout),
 		RequestTimeout:    confutil.P(timeout),
 	})
+	if err != nil {
+		return nil, err
+	}
+
+	cache.clients[key] = client
```
onelapahead, line 267: Do we ever know when we could let go / release a client ? I.e. on `Paladin` delete do we need to check if the manager has a client for that node and can remove it as a result ?

operator/internal/controller/transaction_reconcile.go
onelapahead: I mentioned this bc in ff-common we had a really bad bug in the client library itself that caused a memory leak: https://github.com/hyperledger/firefly-common/pull/172

operator/internal/controller/transaction_reconcile.go
dwertent: This is a good point, I will need to think about it.
Also, I need to check what happens to the client in case the paladin pod restarts. If the client can reconnect easily, so the current implementation is fine because the projected number of clients is low enough that there is no real need in a "cleanup" mechanism. However, if the client has trouble reconnecting after a pod restart than we have a bigger problem to fix.

dwertent: I added a deletion mechanism, although I'm not sure it's necessary given the small number of clients.

onelapahead (approved):
📉

dwertent:
Same as [this comment](https://github.com/LF-Decentralized-Trust-labs/paladin/pull/684#issuecomment-2966963767) just for the [zeto example](https://github.com/LF-Decentralized-Trust-labs/paladin/blob/main/example/zeto/src/index.ts).

PR #654, opened 2025-05-16
Add advanced installation modes to Helm chart

dwertent:
## Overview

This PR extends our Helm chart with enhanced installation modes and comprehensive configuration options, making it easier to tailor Paladin deployments to a variety of environments:

1. **Four Installation Modes**
   * **devnet** (default): Sandbox/demo network with adjustable node count and prefixes.
   * **customnet**: Full control—declare each Paladin node, blockchain endpoints, transports, signers, etc.
   * **attach**: Point Paladin at an existing blockchain and smart contracts, reusing those addresses instead of deploying new ones.
   * **operator-only**: Install only the operator for manual CR application.

2. **Smart Contract Reference Flexibility**
   • Support for specifying either a `deployment` (new contracts) or an `address` (existing on‐chain contracts).
3. **Values File Enhancements**
4. **Dynamic Node Count Support (devnet)**
   • Allow `--set nodeCount=X` to spin up multiple Besu + Paladin node pairs without editing the chart.
   • Introduce `paladin.nodeNamePrefix` and `besu.nodeNamePrefix` for custom auto‐generated names.
5. **Custom Node Name Prefixes (devnet)**
   • Users can override auto‐naming by setting custom prefixes:

   ```bash
   helm install paladin charts/paladin-operator \
     --set nodeCount=5 \
     --set paladin.nodeNamePrefix=worker \
     --set besu.nodeNamePrefix=evm
   ```

Here is the full documentation supporting this PR: [advanced-installation](https://github.com/LF-Decentralized-Trust-labs/paladin/blob/installation-docs/doc-site/docs/getting-started/installation-advanced.md)
PR: #662


dwertent:
> it looks like all the smart contract deployments/transaction invokes all get created, even if the domains/registries are referencing contracts that already exist

That's a good point, I hadn’t fully considered that. On one hand, we use the CRs to track transaction hashes and statuses, so skipping their creation might break that logic. On the other hand, you’re right that there’s no reason to create them if the contracts already exist on-chain.

One potential solution is to skip CR creation when the user provides a contract address. But that introduces complexity,what if the user provides an address for only one domain and not the others? The Helm chart would need to understand the dependencies between domains and CRs, which feels overly complicated.

> I think there's a final piece to add a set of instructions to the docs for installing with customnet too?

Agreed. I’ve been going back and forth on how to best structure the docs. I’ll open a PR on your documentation branch so we can discuss the best way to include that.

dwertent:
After testing this, I think we should allow users to opt out of deploying all smart contracts.
To support different installation needs, we could formalize the following modes:
* **operator-only** – Deploy only the Paladin operator.
* **devnet (default)** – Deploy a demo/example network with bundled contracts.
* **customnet** – Deploy a fully customizable Paladin network, including CRs for smart contracts and domains.
* **attach** – Deploy Paladin configured to use an existing blockchain network and pre-deployed contracts.

The motivation for adding externalnet is that in many cases, users want to reuse an existing network without redeploying the contracts. This would also let us enforce use of the address field (rather than deployment) in the smartContractsReferences section.

PR #655, opened 2025-05-20
zeto: revert during assembly for failures in acquiring input states

awrichar:
Failures that involve malformed or insufficient inputs should result in a revert, rather than an error (which will retry forever).


jimthematrix (approved):
LGTM, thanks @awrichar

PR #658, opened 2025-05-21
Accept eth address as from on transaction submission

annamcallister:
If the value of `from` when submitting a transaction is an 0x address, use it as the verifier for the signing key. 

Following on from feedback on a [previous PR](https://github.com/LF-Decentralized-Trust-labs/paladin/pull/643/commits/c1a05c706de57c84042d017cd23dd20a414b8c3f#r2098369056) which had a first attempt at this change, I have **not** introduced a restriction on 0x addresses being used as identifiers. This is to accommodate the use case that keys are already pre-created in the backend signer,  are only loaded into Paladin when that key is resolved, and the 0x address is what is needed to link the two.

This has implications for the behaviour in different functions:

### `keymgr_resolveKey`/`ptx_resolveVerifier`/`keymgr_resolveEthAddress`

I have not changed the implementation here as I do not think it is necessary.

These methods are explicit in that they resolve from `identifier` to `verifier`. If a 0x address is provided as a parameter it will always be assumed to be an identifier. If the identifier does not exist it will result in the creation of a new identifier.

It is not possible to police what the signer does with that identifier when generating a verifier for the new identity. If we want to support key creation resulting in the signer linking the identifier to an existing key with the same 0x address as the verifier, then it will also be possible that the signer generates a completely different 0x address to be the verifier.

### `ptx_sendTransaction`/`ptx_sendTransactions`/`ptx_call`

This PR changes the behaviour of these functions.

Previous behaviour- `from` is always an identifier.

New behaviour- `from` may be an identifier or a verifier. If `from` is a `0x` address assume first it is a verifier. If the verifier doesn't exist, then treat it as an identifier (which may result in identifier creation on resolution).

### DIFF
--- a/core/go/internal/txmgr/transaction_submission.go
+++ b/core/go/internal/txmgr/transaction_submission.go
@@ -574,6 +574,15 @@ func (tm *txManager) resolveNewTransaction(ctx context.Context, dbTX persistence
 	bypassFromCheck := submitMode == pldapi.SubmitModePrepare || /* no checking on from for prepare */
 		(submitMode == pldapi.SubmitModeCall && tx.From == "") /* call is allowed no sender */
 	if !bypassFromCheck {
+		if _, err := pldtypes.ParseEthAddress(tx.From); err == nil {
+			// Doing the reverse lookup here means that we can persist the identity on the transaction. It does mean that the
+			// identity will later be resolved back to the verifier but this should just be a cache read
+			mapping, err := tm.keyManager.ReverseKeyLookup(ctx, dbTX, algorithms.ECDSA_SECP256K1, verifiers.ETH_ADDRESS, tx.From)
+			if err != nil {
+				return nil, err
+			}
+			tx.From = mapping.Identifier
+		}
 
 		identifier, node, err := pldtypes.PrivateIdentityLocator(tx.From).Validate(ctx, tm.localNodeName, false)
 		if err != nil || node != tm.localNodeName {
--- a/core/go/internal/txmgr/transaction_submission.go
+++ b/core/go/internal/txmgr/transaction_submission.go
@@ -575,13 +575,13 @@ func (tm *txManager) resolveNewTransaction(ctx context.Context, dbTX persistence
 		(submitMode == pldapi.SubmitModeCall && tx.From == "") /* call is allowed no sender */
 	if !bypassFromCheck {
 		if _, err := pldtypes.ParseEthAddress(tx.From); err == nil {
-			// Doing the reverse lookup here means that we can persist the identity on the transaction. It does mean that the
-			// identity will later be resolved back to the verifier but this should just be a cache read
+			// Doing the reverse lookup here means that we can persist the identifier on the transaction. It does mean that the
+			// identifier will later be resolved back to the verifier but this should just be a cache read
 			mapping, err := tm.keyManager.ReverseKeyLookup(ctx, dbTX, algorithms.ECDSA_SECP256K1, verifiers.ETH_ADDRESS, tx.From)
-			if err != nil {
-				return nil, err
+			if err == nil {
+				tx.From = mapping.Identifier
 			}
-			tx.From = mapping.Identifier
+			// else 'from' does not exist as a verifier, treat it as an identifier
 		}
 
 		identifier, node, err := pldtypes.PrivateIdentityLocator(tx.From).Validate(ctx, tm.localNodeName, false)
--- a/core/go/internal/txmgr/transaction_submission_test.go
+++ b/core/go/internal/txmgr/transaction_submission_test.go
@@ -201,6 +201,35 @@ func TestSubmitBadFromAddr(t *testing.T) {
 	assert.Regexp(t, "bad address", err)
 }
 
+func TestSubmitUnknownAddr(t *testing.T) {
+	addr := pldtypes.RandAddress().String()
+	ctx, txm, done := newTestTransactionManager(t, false,
+		mockEmptyReceiptListeners,
+		func(conf *pldconf.TxManagerConfig, mc *mockComponents) {
+			mc.keyManager.On("ReverseKeyLookup", mock.Anything, mock.Anything, algorithms.ECDSA_SECP256K1, verifiers.ETH_ADDRESS, addr).Return(nil, fmt.Errorf("unknown address"))
+			mc.db.ExpectBegin()
+			mc.db.ExpectExec("INSERT.*abis").WillReturnResult(driver.ResultNoRows)
+			mc.db.ExpectExec("INSERT.*abi_entries").WillReturnResult(driver.ResultNoRows)
+		})
+	defer done()
+
+	exampleABI := abi.ABI{{Type: abi.Function, Name: "doIt"}}
+	callData, err := exampleABI[0].EncodeCallDataJSON([]byte(`[]`))
+	require.NoError(t, err)
+
+	_, err = txm.sendTransactionNewDBTX(ctx, &pldapi.TransactionInput{
+		TransactionBase: pldapi.TransactionBase{
+			Type:     pldapi.TransactionTypePublic.Enum(),
+			Function: exampleABI[0].FunctionSelectorBytes().String(),
+			From:     addr,
+			To:       pldtypes.RandAddress(),
+			Data:     pldtypes.JSONString(pldtypes.HexBytes(callData)),
+		},
+		ABI: exampleABI,
+	})
+	assert.Regexp(t, "unknown address", err)
+}
+
 func TestResolveFunctionHexInputOK(t *testing.T) {
 	ctx, txm, done := newTestTransactionManager(t, false,
 		mockEmptyReceiptListeners,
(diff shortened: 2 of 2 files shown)


peterbroadhurst:
Thank you for the great write up.

One item where we need to make a decision, is on identifiers for cryptography that are not Ethereum addresses.

Paladin does already examples some of these, with the Zeto domain, which uses a different verifier scheme with a 32 byte (rather than 20 byte) encoding.

Now there are a number of ways we could approach this, and some have performance impacts.

1. Only perform your logic for Ethereum address verifiers
2. Assume all verifiers start with `0x`
3. Introduce a more specific spelling for lookup by verifier that can have zero overlap with `identifiers`, like `verifier:ABCD`
   - Noting we're restrictive on notation in identifier, so they cannot contain characters like `:`

Now (1) is where the code in your PR is right now.
I personally don't like (2) very much, so am not going to focus on it.

I think (3) is worth a thought experiment:

Negatives:

- You have to do `verifier:0x035cB1e11b2B3AF9644c8ACd93CFa6e02b87bAe6` , rather than just `0x035cB1e11b2B3AF9644c8ACd93CFa6e02b87bAe6`.
- Allows all variations of eth address like checksum free `0x035cb1e11b2b3af9644c8acd93cfa6e02b87bae6` and pure hex `0x035cb1e11b2b3af9644c8acd93cfa6e02b87bae6`

Positives:

- Extensible to all verifier types now and in the future
- Provides identical behavior between submission `from": "verifier:..."` and the `keymgr_reverseKeyLookup` behavior
- The behavior in this scenario is deterministic _error_ rather than creation of the "wrong key"
    - Backend keystore is defined that happens to have identifier `0x035cB1e11b2B3AF9644c8ACd93CFa6e02b87bAe6`
    - Paladin has in the past done a forward-resolution to this key
    - API call is made with `from": "verifier:0x035cB1e11b2B3AF9644c8ACd93CFa6e02b87bAe6"`
    - The key manager reverse-lookup _would not match_ as it can only match after a forward-lookup is done
    - Due to the `verifier:` prefix, we know to return an error here _rather than go and create a key that is named 0x035cB1e11b2B3AF9644c8ACd93CFa6e02b87bAe6, but maps to a key with a different cryptographically derived identifier_.

... I didn't know I thought this until I worked through this mentally writing the comment, but I'm currently having a feeling that the positives of (3) outweigh the benefits when compared to (1).

What do you think?

Chengxuan:
> Previous behaviour: from is always an identifier.

I think this is nice and clear for a "string" field. 

Overloading a "string" field to cover other types of identifiers, naming rules of which are defined by another system, will need careful spelling so that the code can parse out the identifier type out of that spelling. 

Given Paladin is designed to be a programmable engine that could support lots of systems, which all could have their own naming rules for their "identifier". 

There are some other solutions worth thinking through:
- (1) Keep `from` as a strict Paladin identifier, let user translate verifiers into Palaidin identier before calling the API
- (2) Introduce a field to indicate what type of value is passed in by `from` field.
- (3) Keep `from` as a strict Paladin identifier, introduce other fields for other types of verifier. e.g. "ethFrom", "zetoFrom"....

(1) is exactly the problem we are trying to solve here.
(3) could result in a long list of "xxxFrom" 


`(2)` is literally @peterbroadhurst 's `(3)` above, apart from trying to embed the type into a string value. I guess it would slightly increase the combinations and we need to clearly document which "from type" is supported by each transaction type.

e.g. 
- tx type: "private" => "paladin-id", "zeto-id", "eth-address".....
- tx type: "public"  =>"paladin-id", "eth-address"

annamcallister:
Option 3 makes sense to me. It is a relatively small addition to the existing code, but means that rather than trying to infer the intent of the value in the `from` field, we are being explicit.

@Chengxuan I'm not sure that your option 2 is the same. I think it is much larger in scope because it a type field which can be set to an enum which include types both identifier (which you've named `paladin-id`?) and the different types of verifier. 

I believe distinguishing between `identifier` and `verifier` is sufficient provided that for each type of verifier, each identifier has exactly one of them?

If an identifier has multiple verifiers of different types then I would expect context in which it is being used to determine which verifier?

annamcallister:
> I believe distinguishing between identifier and verifier is sufficient provided that for each type of verifier, each identifier has exactly zero or one of them?

https://github.com/LF-Decentralized-Trust-labs/paladin/blob/5466192e7781c15b393e427ec26a7cc228c520cd/core/go/db/migrations/postgres/000011_create_key_mapping_tables.up.sql#L27
and 
https://github.com/LF-Decentralized-Trust-labs/paladin/blob/5466192e7781c15b393e427ec26a7cc228c520cd/core/go/db/migrations/postgres/000011_create_key_mapping_tables.up.sql#L31

With these constraints distinguishing between `identifier` and `verifier` should be sufficient. There are bigger problems to solve here if it is not.

Chengxuan:
If there is a global unique constraint on the value of the verifier e.g. 

> CREATE UNIQUE INDEX key_verifier_globally_unique ON key_verifiers ("verifier");

I think `(3)` is a good option.


Currently, only the value of `identifier` is globally unique.

annamcallister:
I think you're right there- if verifier does is not unique then we can't perform the reverse look up back to the identifier.

Given that we only support a single signer right now, this feels like a good time to add that constraint.

peterbroadhurst:
ok... this is a great point.

So the `verifier` is currently asserted to be unique within the context of the `algorithm` and `verifierType`:

https://github.com/LF-Decentralized-Trust-labs/paladin/blob/ea2bd0e1829041f4d6279e3a1f74272c4c7a712f/core/go/internal/keymanager/key_resolver.go#L398-L400

You're proposing to flatten this to global uniqueness, meaning we can never have two `verifierType`s that resolve to the same output declared by two different cryptography algorithms, and used by the same key.

I do think that is significant, and I don't have an argument against it being a valid constraint.

However, it's not a trivial change though and might come at a significant performance cost to fix (like requiring an additional unique index):
https://github.com/LF-Decentralized-Trust-labs/paladin/blob/ea2bd0e1829041f4d6279e3a1f74272c4c7a712f/core/go/internal/keymanager/key_resolver.go#L429-L432

You see the performance critical code here refers to:
https://github.com/LF-Decentralized-Trust-labs/paladin/blob/ea2bd0e1829041f4d6279e3a1f74272c4c7a712f/core/go/internal/keymanager/key_resolver.go#L379-L383

The code relies on being able to assert a clash means a re-insert.
That can be established with an _additional_ index (rather than changing the current on) and updating the conflict detection to be specific to the full set of fields.

However, I wonder - given we believe this is covering a very, very specific edge case that we believe doesn't exist in practice....

Is it better to _handle on query_ rather than failing on insert?

e.g. if you do `Limit(2)` on the query side without an `algorithm`+`verifierType`, and the length is >1 you can I believe arbitrarily pick the first one, and just `WARN` in the logs.

annamcallister:
I am leaning into agreement that this is `a very, very specific edge case that we believe doesn't exist in practice`.

Particularly as: 
- All interfaces (in code and on the API) for retrieving or resolving a verifier require the algorithm and verifier type to be provided (i.e. there is nowhere currently to add query handling if more than one got returned)
- This feels something that is unlikely to change. In the case that you need to uniquely retrieve a verifier, it is to use it for some kind of signing, this is going to need to algorithm and verifier type.

The remaining missing piece for me that I'm not clear on is that wanted to enforce global uniqueness on verifier, I'm not clear  @peterbroadhurst on why we would need a separate new index, rather than changing the existing one? Anything that is unique with the existing `("verifier", "algorithm", "type")` index would be unique with just a `("verifier")`?

annamcallister:
I'm going to make the change to add the `verifier:` prefix without changing the indexes for now.

doc-site/docs/architecture/key_management.md
```diff
@@ -171,7 +182,7 @@ Every `key mapping` and `folder` gets two attributes automatically:
 - `name`: the part of the `key identifier` representing this key / folder
 - `index`: a numeric identifier, assured to be unique at this folder level
 
-### 3. Public identifiers and algorithms
+### 3. Public verifiers and algorithms
```
peterbroadhurst, line 185:
```suggestion
### 3. Public key identifiers (or "verifiers") and algorithms

Cryptographic keys use public/private cryptography, meaning every key can be identified publicly in a way that does not leak the key information itself.

Any party can _verify_ that a transaction was signed with a particular private key, by _recovering_ the public key used to sign that transaction.

However, to do this we need a standard way to represent a public key so that it can be verified in a standard way against a transaction. Each cryptographic ecosystem actually does this differently, even when using the same private key, and the same _algorithm_ (such as SECP256K1).

The most obvious example of this is the Ethereum address. This is a 20 byte compressed representation of a SECP256K1 public key, derived using a well documented algorithm. It looks like `0xfdcba455d748cb3e085472cc6f49b4ae86ee4d1f` in hex, and sometimes is represented in a case-sensitive check-summed way like `0xFdcBa455D748cB3e085472CC6F49B4AE86eE4d1F` per the EIP-55 standard.

Paladin supports multiple of these "verifiers" to be calculated, and stored, for public keys.
```

core/go/internal/txmgr/transaction_submission.go
```diff
@@ -574,6 +574,19 @@ func (tm *txManager) resolveNewTransaction(ctx context.Context, dbTX persistence
 	bypassFromCheck := submitMode == pldapi.SubmitModePrepare || /* no checking on from for prepare */
 		(submitMode == pldapi.SubmitModeCall && tx.From == "") /* call is allowed no sender */
 	if !bypassFromCheck {
+		if strings.HasPrefix(tx.From, "verifier:") {
+			addr := strings.TrimPrefix(tx.From, "verifier:")
+			if _, err := pldtypes.ParseEthAddress(addr); err != nil {
```
peterbroadhurst, line 579: So just to be clear this means we'll only support Ethereum Address verifiers at the end of this PR.

PR #660, opened 2025-05-21
Include the indexed block in an indexed event dispatched on an event stream

annamcallister:
The indexed block includes useful information like the block timestamp.

Indexed block was already part of the indexed event struct, and is included when querying events. However, it is not desirable when receiving events on an event stream to have to make additional calls to enrich those events.


peterbroadhurst (approved):
No concerns with the change - although i couldn't see an external change here to any of the clients, or the removal of any query code to backfill this info.

PR #664, opened 2025-05-27
Kubernetes friendliness updates

davecrighton:
When running in an actual k8s cluster rather than kind I encountered a few issues.

1.) The image uses the root user
2.) The startup sequence of paladin extracts a native library to java.io.tmpdir, this defaults to /tmp which in a real k8s cluster is mounted a read only (since it is the root file system)
3.) During startup the either the key manager of registry plugin creates a UDS socket which requires a place to write the socket file to that is not mounted as read only. This used /app/lib previously but this ends up mounted read only in a real k8s cluster

I couldn't solve this by mounting a volume from k8s into /app/config since that makes the paladin jar inaccessible so I introduced a new volume mount /app/jna to hold the extracted jna libs and the UDS socket files. Additionally I modified the docker file to use USER 1001.

Also updated db migrations to not require postgres 15


core/go/db/migrations/postgres/000010_create_registry_tables.up.sql
```diff
@@ -17,7 +17,7 @@ CREATE TABLE reg_entries (
 );
 
 -- The name is scoped uniquely within the parent
-CREATE UNIQUE INDEX reg_entries_name ON reg_entries("registry", "name", "parent_id") NULLS NOT DISTINCT;
+CREATE UNIQUE INDEX reg_entries_name ON reg_entries("registry", "name", "parent_id");
```
dwertent, line 20: @davecrighton  Are you sure about this behavioral change?

peterbroadhurst: I suggest a unit test that confirms you are unable to add two root records would be a good place to confirm this index needs the `parent_id IS NULL` protection, and would let you test a PSQL <15 solution

davecrighton: Tests and alternative SQL for the migration added in https://github.com/LF-Decentralized-Trust-labs/paladin/pull/664/commits/4da1a205953d0e266cf5c0614ea5aa339e0bccdf

core/go/db/migrations/postgres/000010_create_registry_tables.up.sql
```diff
@@ -17,7 +17,7 @@ CREATE TABLE reg_entries (
 );
 
 -- The name is scoped uniquely within the parent
-CREATE UNIQUE INDEX reg_entries_name ON reg_entries("registry", "name", "parent_id") NULLS NOT DISTINCT;
+CREATE UNIQUE INDEX reg_entries_name ON reg_entries ("registry", "name", ("parent_id" IS NULL)) WHERE "parent_id" IS NULL;
```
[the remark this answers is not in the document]
peterbroadhurst: As an aside, if you do create two indexes by modifying the existing migration - you need to ensure the down migration has `IF EXISTS` on the delete - because anyone who ran the PSQL 15 version of the 10 migration will only have one index.

peterbroadhurst (approved):
Thanks for going back round the loop on the DB fix, and adding the test 👍

PR #665, opened 2025-05-28
Adjust Noto examples to use "lock" in all cases instead of "approve"

awrichar:
The Noto "lock" functionality is the only reliable way to execute pre-approved transactions. While the "approveTransfer" and "prepare" flow can work in a few simple cases, it has numerous shortcomings and may be removed in the future. Switch all tests and examples to reflect this.

This PR also includes a few other cleanup items in domains, including adding domain receipts to testbed transaction to replace some problematic usage of internal domain data in tests.


jimthematrix (approved):
LGTM with only a minor glitch in the readme for a link

PR #671, opened 2025-06-02
Fixes after trying to follow the hello world tutorial

matthew1001:
I found that the download link for the latest release's download didn't work. It looks like maybe there's no direct link to the downloads for the `latest` GH release? I've tweaked the instruction to go to the `latest` release and then download `abis.tar.gz`. Let me know if you think there's a direct link that will work instead.

I also found that I needed to run `npm run abi` in the option 1 steps, so I've added that into the instructions.

### DIFF
--- a/operator/Makefile
+++ b/operator/Makefile
@@ -1,5 +1,7 @@
 CLUSTER_NAME ?= paladin
 NAMESPACE ?= default
+NODE_COUNT ?= 3
+MODE ?= devnet
 
 OPERATOR_IMAGE_NAME ?= paladin-operator
 OPERATOR_IMAGE_TAG ?= test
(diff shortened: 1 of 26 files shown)


example/helloworld/README.md
```diff
@@ -20,6 +21,7 @@ Check out the [tutorial](https://lf-decentralized-trust-labs.github.io/paladin/h
 cp -rf src/abis ../../sdk/typescript/src/domains/
 cd ../../sdk/typescript
 npm install
+npm run abi
```
dwertent, line 24: The reason we download the ABIs is so we won't need to build it later

dwertent: I see the `Zeto_Anon.json` is missing from the artifacts. I'm looking into it

dwertent: Ok, I see the `npm run abi` was introduced 2 weeks ago. It's fine by me

example/helloworld/README.md
```diff
@@ -11,7 +11,8 @@ Check out the [tutorial](https://lf-decentralized-trust-labs.github.io/paladin/h
 ## Option 1: Use the Latest Stable Version
 
 1. **Download and Extract Contracts**
-   - [Download Solidity contracts](https://github.com/LF-Decentralized-Trust-labs/paladin/releases/download/latest/abis.tar.gz)
```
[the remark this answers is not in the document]
matthew1001: Thanks @dwertent that one works. I've updated the PR

dwertent (approved):

PR #677, opened 2025-06-03
Remove incomplete implementation of auto-fueling

annamcallister:
Kaleido contributed significant orchestration engine capability into the public TX manager of this from our closed source. 

This included some code related to an advanced feature in the Kaleido platform, but the parts of the implementation that came across were incomplete and non-functional.

There is no known requirement for this feature to be in the base Paladin codebase currently, so the presence of the incomplete code is confusing for maintainers of the code.

So this PR cleans up the code to remove the non-functional cruft.


annamcallister:
Messages removed, as well as a general sweep and clean up of comments about auto fuelling.

I've opted to fully delete the messages rather than just comment them out.

While in general I don't think this is best practice, since it opens the door to reuse of error codes, I don't think that's a risk that applies here if no one has ever been able to hit the errors. I've gone for a clean slate instead.

peterbroadhurst (approved):

PR #679, opened 2025-06-03
Zeto balances

ka1ii:
addresses https://github.com/LF-Decentralized-Trust-labs/paladin/issues/594

### DIFF
--- a/domains/zeto/build.gradle
+++ b/domains/zeto/build.gradle
@@ -25,7 +25,7 @@ ext {
     maxCoverageBarGap = 1
     coveragePackages = [
         "github.com/kaleido-io/paladin/domains/zeto/internal/...",
-        "github.com/kaleido-io/paladin/domains/zeto/pkg/types",
+        "github.com/kaleido-io/paladin/domains/zeto/pkg/types/states.go",
         "github.com/kaleido-io/paladin/domains/zeto/pkg/zetosigner",
     ]
 
--- a/domains/zeto/internal/zeto/signer/common/babyjubjub_test.go
+++ b/domains/zeto/internal/zeto/signer/common/babyjubjub_test.go
@@ -110,3 +110,50 @@ func TestNewBabyJubJubPrivateKey(t *testing.T) {
 		})
 	}
 }
+
+func TestEncodeBabyJubJubPublicKey(t *testing.T) {
+	tests := []struct {
+		name     string
+		pubKey   *babyjub.PublicKey
+		expected string
+	}{
+		{
+			name:     "mock public key",
+			pubKey:   mockPubKey(),
+			expected: "0x51fa904bb6142e89f85aebb2a933a879e2efd5b682021deec4f717a8dbcbbd8e",
+		},
+		{
+			name: "zero public key",
+			pubKey: &babyjub.PublicKey{
+				X: big.NewInt(0),
+				Y: big.NewInt(1),
+			},
+			expected: "0x0100000000000000000000000000000000000000000000000000000000000000",
+		},
+		{
+			name: "generated key pair",
+			pubKey: func() *babyjub.PublicKey {
+				privKey := babyjub.NewRandPrivKey()
+				return privKey.Public()
+			}(),
+			expected: "",
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			encoded := EncodeBabyJubJubPublicKey(tc.pubKey)
+
+			assert.Len(t, encoded, 66, "encoded key should be 66 characters")
+
+			if tc.expected != "" {
+				assert.Equal(t, tc.expected, encoded)
+			}
+
+			decoded, err := DecodeBabyJubJubPublicKey(encoded)
+			require.NoError(t, err)
+			assert.Equal(t, tc.pubKey.X.String(), decoded.X.String())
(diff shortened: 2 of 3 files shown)


domains/zeto/internal/zeto/fungible/states.go
```diff
@@ -215,3 +216,51 @@ func findAvailableStates(ctx context.Context, callbacks plugintk.DomainCallbacks
...
+	ctx context.Context,
+	callbacks plugintk.DomainCallbacks,
+	coinSchema *pb.StateSchema,
+	useNullifiers bool,
+	stateQueryContext, accountKey string,
+) (*big.Int, error) {
+	total := big.NewInt(0)
+	var lastStateTimestamp int64
+
+	for {
```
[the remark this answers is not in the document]
jimthematrix: I agree with the suggestion to add a limit to avoid this sitting in the loop for a long time. And the per-query limit can be more than 10 (it looks like this number came from the same query used in the `getAvailableCoins()` method, which uses this low number due to the input UTXO array size of 10 for Zeto transfers).

If the query returns as a result of hitting the limit, I wonder if the right response is "here's the balance number I've got but it might be larger because I stopped short of exhausting all the active coins you have". So that the UX can reflect the "non-authoritative result". This might be useful if my goal is making sure I can pay certain amount, but don't necessarily need to know the exact amount.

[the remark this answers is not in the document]
ka1ii: Although if there is a substantial amount of coins then its possible for the request to timeout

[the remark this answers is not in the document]
awrichar: Thinking further, I'm wondering if the loop is even useful. It might be more efficient to perform a single database query rather than multiple. It's useful with normal coin selection because we may break it early.

solidity/contracts/domains/interfaces/IZetoFungible.sol
```diff
@@ -24,4 +24,5 @@ interface IZetoFungible {
     function deposit(uint256 amount) external;
     function withdraw(uint256 amount) external;
     function setERC20(address erc20) external;
+    function balanceOf(string memory account) external view returns (uint256 balance, string memory balanceNote);
```
awrichar, line 27: I'm not sure about returning errors as a string (and requiring users to check for an empty string to ensure success?)

One alternate suggestion - we could include the total number of states found and a flag to indicate if we stopped counting... something like `(uint256 totalStates, uint256 totalBalance, bool overflow)`.

ka1ii: thats a good idea, although is the boolean overflow needed if we document the max totalStates we can process is 1000

awrichar: I think it helps if you don't read the docs :) at least you wonder "what does this mean".

domains/zeto/internal/zeto/fungible/states.go
```diff
@@ -215,3 +216,42 @@ func findAvailableStates(ctx context.Context, callbacks plugintk.DomainCallbacks
...
+
+	states, err := findAvailableStates(
+		ctx, callbacks, coinSchema,
+		useNullifiers, stateQueryContext, queryBuilder.Query().String(),
+	)
+	if err != nil {
+		return 0, nil, false, i18n.NewError(ctx, msgs.MsgErrorQueryAvailCoins, err)
+	}
+
+	// sum up this page
```
awrichar, line 243: it's no longer a page... it's everything (commented probably not needed)

domains/zeto/internal/zeto/fungible/handler_balanceOf.go
```diff
@@ -0,0 +1,99 @@
...
+	param := tx.Params.(*types.FungibleBalanceOfParam)
+	resolvedAccount := domain.FindVerifier(param.Account, h.getAlgoZetoSnarkBJJ(), zetosignerapi.IDEN3_PUBKEY_BABYJUBJUB_COMPRESSED_0X, req.ResolvedVerifiers)
+	if resolvedAccount == nil {
+		return nil, i18n.NewError(ctx, msgs.MsgErrorResolveVerifier, param.Account)
+	}
+	useNullifiers := common.IsNullifiersToken(tx.DomainConfig.TokenName)
+	totalStates, totalBalance, overflow, err := getAccountBalance(ctx, h.callbacks, h.stateSchemas.CoinSchema, useNullifiers, req.StateQueryContext, resolvedAccount.Verifier)
+	if err != nil {
+		// don't think we are wrapping errors currently, throwing this out here see if folks are fond of it.
+		return nil, i18n.WrapError(ctx, err, msgs.MsgErrorGetAccountBalance, param.Account)
```
awrichar, line 92: Yes, wrapping errors is probably a good practice.

domains/zeto/internal/zeto/fungible/handler_balanceOf.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	// Format balance as JSON string
+	balanceJson := fmt.Sprintf(`{"totalBalance":"%s","totalStates":"%s","overflow":"%s"}`, fmt.Sprint(totalBalance), fmt.Sprint(totalStates), fmt.Sprint(overflow))
```
awrichar, line 95: Can we define a struct for this and use `json.Marshal`?

PR #684, opened 2025-06-04
Noto balance

ka1ii:
addresses https://github.com/LF-Decentralized-Trust-labs/paladin/issues/594


dwertent:
This is great!
Can we add this to the examples?
e.g. [get balance after minting](https://github.com/LF-Decentralized-Trust-labs/paladin/blob/main/example/notarized-tokens/src/index.ts#L42) and [after a trnsaction](https://github.com/LF-Decentralized-Trust-labs/paladin/blob/main/example/notarized-tokens/src/index.ts#L55) and [here](https://github.com/LF-Decentralized-Trust-labs/paladin/blob/main/example/notarized-tokens/src/index.ts#L68).

domains/noto/pkg/types/abi.go
```diff
@@ -124,3 +124,13 @@ type NotoTransferMetadata struct {
...
 }
+
+type BalanceOfParam struct {
+	Account string `json:"account"`
+}
+
+type BalanceOfResult struct {
+	TotalBalance string `json:"totalBalance"`
+	TotalStates  string `json:"totalStates"`
+	Overflow     string `json:"overflow"`
```
awrichar, line 135: Shouldn't these types be things like `HexUint256` and boolean, rather than all strings?

domains/noto/internal/noto/noto.go
```diff
@@ -532,41 +533,61 @@ func (n *Noto) validateDeploy(tx *prototk.DeployTransactionSpecification) (*type
...
+		return nil, zero, i18n.NewError(ctx, msgs.MsgUnknownFunction, functionABI.Name)
 	}
 
-	contractAddress, err := ethtypes.NewAddress(tx.ContractInfo.ContractAddress)
-	if err != nil {
-		return nil, nil, err
+	// check if the handler implements the ValidateParams method cause generic T
+	validator, ok := any(handler).(interface {
+		ValidateParams(ctx context.Context, domainConfig *types.NotoParsedConfig, paramsJson string) (any, error)
+	})
```
awrichar, line 566: Maybe pull this interface out as a separate definition, instead of defining it inline?

ka1ii: the type check here is just so that compiler doesn't complain about handler.validateparams

PR #686, opened 2025-06-09
Upgrade to mockery v3

annamcallister:
Upgrade to mockery v3- the changes to do this largely fall into three categories

### 1. Config file rather than CLI options
It is no longer possible to pass options in on the CLI in v3; a config file must be used.

V3 favours colocating mocks with the implementation package, but I have kept with the v2 style of a mocks directory. This is in large part to keep `gradle clean` and `.gitignore` files as simple as possible. The one exception to this is the `TransactionFlow` mock which is colocated with its package. Because of the common config file, we can no longer generate two sets of mocks for the same package (not without making another config file which isn't desirable), so this is now the only mock for this interface.

### 2. More consistency in naming conventions
I have tried to keep the set of mocks generated as close to what we generating in v2 as possible. However, because the config file allows for common properties to be set at the top level, there are changes to the names of some packages/files, rather than encoding past inconsistency into a brand new file when there is no good reason to do so.

### 3. Fallout from change in how mockery handles arguments in variadic functions
In v2, if we had a function with signature `func (f *F) foo(params ...interface{})`, and it was called with three arguments: `1, 2, 3`, the mock could match against these as separate parameters: `f.On("foo", 1, 2, 3)`. In functions that make the argument array available, e.g. `Run`, they would be in the argument array as `[1, 2, 3]`.

In v3, these parameters are not separated out, so we have to match as `f.On("foo", []interface{}{1, 2, 3})`. In the argument array they are available as `[[1, 2, 3]]`.

This has resulted in a lot of code changes to how mocks are set up, particularly in the `blockindexer` package.


matthew1001 (approved):
LGTM, small comment but happy as-is

PR #693, opened 2025-06-12
Distribute failure receipts back to sender

peterbroadhurst:
If you have a transaction that reverts before making it to the blockchain, during the `Assembly` phase - then the error only makes it to the coordinator. It never makes it into a receipt in the node that performed the send.

This PR is specific to this failure scenario, identifies the original sender of the transaction, and distributes the receipt back to them. As long as the sender can be determined, the delivery of the receipt is reliable and retried.

Note: this was left as a TODO when we put in the reliable messaging types, as the architectural design for this had been discussed back then. However, we didn't close the implementation:

https://github.com/LF-Decentralized-Trust-labs/paladin/blob/37f89630ba341a71a4195b0d4f83f2454b36ea61/core/go/internal/transportmgr/peer.go#L389-L391

### DIFF
--- a/core/go/internal/groupmgr/rpc_eventstreams_test.go
+++ b/core/go/internal/groupmgr/rpc_eventstreams_test.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/kaleido-io/paladin/config/pkg/confutil"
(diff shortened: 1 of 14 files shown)


core/go/internal/privatetxnmgr/syncpoints/finalize.go
```diff
@@ -63,20 +67,44 @@ func (s *syncPoints) writeFailureOperations(ctx context.Context, dbTX persistenc
 	//
 	// However, a syncpoint gets triggered for every finalize so that we can flush the Domain Context to the DB
 	// so that all states are stored, before we clear out the transaction from the in-memory Domain Context.
-	failureReceipts := make([]*components.ReceiptInput, 0)
+	//
+	// Receipts need to go back to their originator, so we either store the receive ourselves locally - or
```
matthew1001, line 71:
```suggestion
	// Receipts need to go back to their originator, so we either store the receipt ourselves locally - or
```

core/go/internal/privatetxnmgr/syncpoints/finalize.go
```diff
@@ -63,20 +67,44 @@ func (s *syncPoints) writeFailureOperations(ctx context.Context, dbTX persistenc
...
-		return s.txMgr.FinalizeTransactions(ctx, dbTX, failureReceipts)
+	var err error
+	if len(localFailureReceipts) > 0 {
+		err = s.txMgr.FinalizeTransactions(ctx, dbTX, localFailureReceipts)
+	}
+	if err == nil && len(remoteSends) > 0 {
+		// We log and ignore errors here, because if it is a DB transaction error we will
+		// fail the DB transaction. If it is a peer resolution error - then we are not
+		// going to be able to get this receipt back. So we can only log it.
+		if peerErr := s.transportMgr.SendReliable(ctx, dbTX, remoteSends...); peerErr != nil {
```
hosie, line 104: I've not managed to convince myself that this error is definitely never going to be a transient error.  However, I think it should be true and if not, then that is probably a bug in the transport manager.

From code inspection, it seems like it might be possible for errors from `connectPeer` to percolate up here.  I might be wrong and it may be the case that we would never end in that code path but I am not finding it easy to convince myself that is the case.  So I am wondering if we really ought to explicitly declare the behaviour of `SendReliable` as "will only return error in case of DB transaction error or an invalid node name.   And maybe add tests ( and possible code changes) to transportManager to make assert that to be true.

I don't think that needs to block this PR because at worst, it does seem edge case where a recoverable error is thrown and this PR improves the behavior in all other cases.

 @matthew1001 maybe something to think about once you get a chance to page into the intent and mechanics of the transport manager the reliable messaging handshake.

hosie (approved):
LGTM

awrichar:
Might resolve https://github.com/LF-Decentralized-Trust-labs/paladin/issues/620?

PR #697, opened 2025-06-14
Ignore conflicts when inserting privacy groups

awrichar:


peterbroadhurst (approved):
My page back into this code indicates the data is validated via state hash at the point we get to this code, so handling a duplicate insert like this is valid and does not expose any integrity risk.

PR #673, opened 2025-06-02
Distributed sequencer protocol

hosie:
work in progress.  Complete TODO list to follow...
 - [ ] ...

PR #704, opened 2025-06-17
Distributed sequencer initial integration

matthew1001:
## Overview

This PR is the result of large engineering effort to introduce a distributed sequencer to Paladin.

Paladin transactions for domains that produce and consume shared state (such as the `Noto` and `Pente` reference domains) need coordinating in order to prevent state being spent more than once, and to allow one transaction to spend state after another transaction has created it. Coordinating or "sequencing" transactions efficiently across multiple Paladin nodes requires nodes to agree on who will be the coordinator at any given time. This PR introduces a new distributed sequencer designed to efficiently assemble private transactions, while also allowing nodes to take over as the sequencer should another node suffer an outage.

The design of the new sequencer (which replaces the original private transaction manager) is based around a set of state machines. These state machines track the role a paladin node is playing for a particular domain contract, and also the state of transactions within the domain as they are submitted, assembled, endorsed, and then finally submitted to the base ledger.

The design of the state machine allows for transactions to move from later stages in the state machine back to earlier stages. For example if an assembled transaction reverts on the base ledger, it will be moved back to assembling state to be re-assembled. This can continue theoretically indefinitely until a transaction reaches a terminal state.

This PR brings together the implementation of the distributed state machine and the integration of the state machine with the rest of Paladin. For example:

 - State machine initialization when a domain contract is being used
 - Communication between the state machines on different nodes to agree on which node is currently coordinating the domain contract, requesting and gathering endorsements, and updating nodes on transaction outcome.
 - Public transaction manager integration to submit base ledger transactions for endorsed private transactions
 - Public transaction manager integration to update the state machine based on the outcome of base ledger transactions

Because the new transaction sequencer is only an in-memory engine for transaction coordination, this PR does not introduce any external API changes to Paladin, or to the existing reference domains `Noto`, `Pente`, and `Zeto`. Existing applications should continue to work with the new sequencer and testing has focused on ensuring this is the case.

Fixes https://github.com/LF-Decentralized-Trust-labs/paladin/issues/703

## Detailed breakdown of the PR work and progress

 - [x] All samples run through Paladin using the new sequencer
 - [x] Existing component & sequencer component unit tests green
 - [x] Metrics are enhanced to show sequencer-specific activity
 - [ ] ~Moderate load testing has shown that the sequencer is stable over hours of TX processing~ See https://github.com/LF-Decentralized-Trust-labs/paladin/issues/880
 - [x] Docs have been updated with a clear explanation of how the sequencer works and how to diagnose issues
 - [x] Encode a typical DvP demo scenario into component tests

### High-level discussions yet to close out

 - [x] Handling of in-memory state-machine state after a crash
   - This is primarily a case of reloading any in-progress (un-confirmed) transactions into the state machine and re-coordinating from scratch
 - [x] State machine heartbeat behaviour. Currently there are N grace period heartbeats before TX finalisation, need discussion on the correct place to leave this.
 - [x] Algorithm for selecting the current sequencer
   - We have an approach already present in the private TX manager that uses a modulus function on the current block number to have a deterministic next coordinator to attempt to use
   - I think there is one outstanding discussion in this space, specifically how a node determines that it is giving up on that initial coordinator. This is needed for `<100%` endorsement domains in the future, but maybe doesn't need solving before the first distributed sequencer can be merged.
 - [ ] Choose alternative sequencer if the expected one doesn't start coordinating within expected time
 - [x] Transaction dispatch is controlled by limiting the number of transaction assemblies we're willing to be optimistic about. Reverts are tested and cause re-assemble automatically, but I'm not currently sure if a public revert should cause all assembled-but-not-dispatched transactions to be discarded and re-assembled.

### Changes I've made to the initial design that would be good to highlight for specific concerns/discussion

I'll check them off when I feel there has been sufficient visibility of them with other maintainers.

 - [x] Changed the term `sender` to `originator`
 - [ ] Assemble responses are now passing back the pre-assembly as well as post-assembly, so the coordinator can build the endorse request
 - [ ] Heartbeat handling
    - I think there is some tidy up to do here in the state machine's definition of a heartbeat vs a heartbeat interval. I'm not sure it needs any changes for V1 but I'm leaving this open to discuss with John before checking it off
 - [x] The original sequencer design used the term "committee", which I think got overloaded somewhat (or at least wasn't always clear if it was a committee of senders or a committee of coordinators, or both. I've renamed this internally to be a `originatorNodePool` for now, since I think that was its purpose in most cases. The fair-selection of transactions (see below) was one area it was perhaps being treated differently as a `originatorIdentityPool` (i.e. 5 identities from the same node were equally served by the TX selection). Having temporarily removed the fair-tx-selection logic for now I think `originatorNodePool` is nice and clear and we have use a separate pool for different purposes if we have one.
 - [ ] The sender transaction state machine now allows `Event_AssembleRequestReceived` while in `State_Assembling`
   - Without this there is a case where a temporary error during assembly leaves the sender's transaction in `State_Assembling` but where assembly hasn't actually been successful. For example if the assembling node can't resolve the identity of one or more required verifiers because their nodes are offline we don't have any reason to think assembly won't be successful with the chosen states, we just need to be allowed to try again based on the coordinator's timeout configuration. Rather than treat a state action error as "let's go back to the previous state" which introduces an entirely new set of changes to the state machine I think it's reasonable for us to go into assembling but be allowed to re-assemble while in that state.
 - [ ] The state machine also now accepts `Event_Confirmed` at any point in the TX lifecycle. For example, a coordinator successfully gathers endorsements, then submits for dispatch, then crashes. On reload the state machine will revert to `Pooled`, but the indexer may then confirm the transaction and cause us to jump straight from `Pooled` to `Confirmed`.
 - [ ] The state machine also now accepts `Event_AssembleRequestReceived` when in `State_Submitted` to manage a race between the coordinator deciding re-assembly needs to take place and the originator moving to the correct state for re-assembly (namely, `State_Delegated)
 - [ ] The coordinator state machine also accepts `Event_EndorsementRequested` when in idle, which covers an edge case where the coordinator is not in `State_Observing` and hence is in a window where it cannot elect itself to take over
 - [ ] The originator state machine has a new `validator_TransactionDoesNotExist` check so the state machine is effectively idempotent from the point of view of `Event_TransactionCreated` events. This is particularly necessary to support resuming transactions from the DB safely.

### Specific edge-cases to test/validate with the new sequencer

 - [ ] locked state but prepare with that state fails
 - [x] Distribute receipts of upstream TXs for chained private, back to sender (see https://github.com/matthew1001/paladin/commit/7582d20baf88aac304134615b75fd48a9e794b25)
 - [x] Distributed notary with pente hooks receiving the receipt back at the notary

### Config points still to add and minor `TODO`s

 - [x] Max number of in-flight coordinated transactions
 - [x] Max distance ahead to dispatch transactions
 - [ ] ~PreDispatch requests currently piggy back on `TransactionConfirmed` protobuf message types - needs their own protobuf messages~ See issue **https://github.com/LF-Decentralized-Trust-labs/paladin/issues/879**

### Candidates to defer to a v-next of the distributed sequencer

 - I've spent some time considering the fair selection of transactions for sequencing. @hosie had an initial design for this in the branch I have been building on. I found it hard to massage that design into something that worked well for a varying pool of sender nodes e.g. notarized tokens where any new party could be sent tokens and/or send them elsewhere. I may reinstate his design for privacy-group and self-coordinated domains and leave notarized domains as first-come-first-serve, but I may just leave all domains as first-come-first-served depending on the priority of closing out the rest of the feature.
 - Sender->coordinator buffering to prevent coordinator overload while also avoiding unnecessary/wasteful coordinator handover

### DIFF
--- a/core/go/noderuntests/README.md
+++ b/core/go/noderuntests/README.md
@@ -3,11 +3,13 @@
 ## Run the whole test suite using gradle
 
 Run tests with gas free chain:
+
 ```
 gradle core:go:componentTestSQLite
 ```
 
 Run tests with chain that uses gas:
+
 ```
 gradle core:go:componentTestWithGasSQLite
 ```
@@ -22,11 +24,16 @@ gradle core:go:coordinationTestPostgres
 
 # Run in VS code
 
-To run individual tests with the `Go` VS Code extension 
+To run individual tests with the `Go` VS Code extension
 
 1. Start the test infrastructure
-  ```
-  gradle startTestInfra
-  ```
+
+```
+gradle startTestInfra
+```
+
 1. Set `go.testEnvFile` to the absolute path to `core/go/noderuntests/.env` file
 
+## Unit tests
+
+There are also coordination unit tests which are alongside the code. The tests in `/noderuntests` are not unit tests but in fact launch all of the node components and run full nodes, and hence have been kept here alongside component tests for clarity.


config/pkg/pldconf/privatetxmgr.go
```diff
@@ -58,4 +59,5 @@ type PrivateTxManagerSequencerConfig struct {
 	StaleTimeout                        *string `json:"staleTimeout,omitempty"`
 	RoundRobinCoordinatorBlockRangeSize *int    `json:"roundRobinCoordinatorBlockRangeSize,omitempty"`
 	AssembleRequestTimeout              *string `json:"assembleRequestTimeout,omitempty"`
+	BlockRangeSize                      *int64  `json:"blockRangeSize,omitempty"`
 }
```
peterbroadhurst: Not sure I understand this structure and where it fits (I see `SequencerConfig` in a dedicated file)

config/pkg/pldconf/sequencermgr.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+type SequencerConfig struct {
+	AssembleTimeout               *string           `json:"assembleTimeout"`
+	RequestTimeout                *string           `json:"requestTimeout"`
+	BlockHeightTolerance          *uint64           `json:"blockHeightTolerance"`
+	BlockRange                    *uint64           `json:"blockRange"`
+	ClosingGracePeriod            *int              `json:"closingGracePeriod"`
+	MaxInflightTransactions       *int              `json:"maxInflightTransactions"`
+	MaxDispatchAhead              *int              `json:"maxDispatchAhead"`
```
peterbroadhurst: Definitely not an immediate thing, but I wonder over time if some of these need to negotiated smart contract scoped config - rather than independent node config. It's the same kind of balance as something like max gas per block in the underlying ledger consensus.

core/go/internal/domainmgr/private_smart_contract.go
```diff
@@ -250,11 +252,14 @@ func (dc *domainContract) AssembleTransaction(dCtx components.DomainContext, rea
 		// - Attestation plan identities
 		// - State distributions
 		// - Associated nullifier requests
+		log.L(dCtx.Ctx()).Info("Qualifying assembly identities")
 		dc.fullyQualifyAssemblyIdentities(res)
 
 		// Note the states at this point are just potential states - depending on the analysis
 		// of the result, and the locking on the input states, the engine might decide to
 		// abandon this attempt and just re-assemble later.
+		log.L(dCtx.Ctx()).Debugf("Number of potential output states: %+v", len(res.AssembledTransaction.OutputStates))
```
peterbroadhurst: These log entries feel like intermediate log entries from code dev, which need to be consolidated into a log entry from this function that can be left on.

core/go/internal/domainmgr/private_smart_contract.go
```diff
@@ -477,12 +488,14 @@ func (dc *domainContract) EndorseTransaction(dCtx components.DomainContext, read
 	if err != nil {
 		return nil, err
 	}
-	return &components.EndorsementResult{
+	log.L(dCtx.Ctx()).Infof("Endorse transaction returning endorsement result with payload length %d: %+v", len(res.Payload), res.Payload)
```
peterbroadhurst: Not sure the length of the payload is the most significant summary item to return here

core/go/internal/flushwriter/flushwriter.go
```diff
@@ -246,7 +255,9 @@ func (w *writer[T, R]) worker(i int) {
 		}
 
 		if b != nil && (timedOutOrFlush || (len(b.ops) >= w.batchMaxSize)) {
-			b.timeoutCancel()
+			if b.timeoutContext != nil {
```
peterbroadhurst: What's the situation covered by this? I can't see the bug above

core/go/internal/flushwriter/flushwriter.go
```diff
@@ -197,7 +197,16 @@ func (w *writer[T, R]) queue(ctx context.Context, value T, flush bool) *op[T, R]
 }
 
 func (w *writer[T, R]) worker(i int) {
-	defer close(w.workersDone[i])
+	defer func() {
+		if w.workersDone[i] != nil {
```
peterbroadhurst: Why would it be closed already? Does this go routine now own the channel? What's the channel design here?

core/go/internal/identityresolver/identityresolver.go
```diff
@@ -173,13 +173,31 @@ func (ir *identityResolver) ResolveVerifierAsync(ctx context.Context, lookup str
...
+				return
+			case err, ok := <-errChan:
+				if !ok {
+					return
+				}
+				log.L(ctx).Errorf("Failed to send resolve verifier request to %s: %s", remoteNodeId, err)
+				ir.handleResolveVerifierError(ctx, resolveVerifierRequestBytes, requestID.String())
+				return
+			}
+		}()
```
peterbroadhurst: This just feels a little odd.
If I want to do a send-once-don't-retry function, shouldn't that be the interface?
What causes the function to exit? Why does the context cancel?

core/go/internal/plugins/domains.go
```diff
@@ -221,13 +222,17 @@ func (br *domainBridge) InitTransaction(ctx context.Context, req *prototk.InitTr
...
+	log.L(ctx).Infof("[Plugin] assembling transaction %s", req.Transaction.TransactionId)
 	err = br.toPlugin.RequestReply(ctx,
 		func(dm plugintk.PluginMessage[prototk.DomainMessage]) {
 			dm.Message().RequestToDomain = &prototk.DomainMessage_AssembleTransaction{AssembleTransaction: req}
 		},
 		func(dm plugintk.PluginMessage[prototk.DomainMessage]) bool {
 			if r, ok := dm.Message().ResponseFromDomain.(*prototk.DomainMessage_AssembleTransactionRes); ok {
 				res = r.AssembleTransactionRes
+				for i, attRequest := range res.AttestationPlan {
+					log.L(ctx).Infof("[Plugin] assembled transaction %s, assemble response %d name '%s'", req.Transaction.TransactionId, i, attRequest.Name)
```
peterbroadhurst: Why info level for just this one request to a plugin, and why this info?
We have very detailed debug logging to plugins. If there's a good reason for escalated info level logging that's something be be considered carefully I think.

core/go/internal/publictxmgr/in_flight_transaction_stage_controller.go
```diff
@@ -391,6 +417,10 @@ func (it *inFlightTransactionStageController) processSigningStageOutput(ctx cont
 				Created:         pldtypes.TimestampNow(),
 				TransactionHash: *rsc.StageOutput.SignOutput.TxHash,
 				GasPricing:      gasPriceJSON,
+				PrivateTXID:     txnID,
+			}
+			if rsc.InMemoryTx.GetTo() != nil {
+				rsc.StageOutputsToBePersisted.TxUpdates.NewSubmission.ContractAddress = rsc.InMemoryTx.GetTo().String()
```
peterbroadhurst: Why would this change between submissions?

core/go/internal/publictxmgr/in_flight_transaction_state_generation.go
```diff
@@ -176,7 +176,12 @@ func (v *inFlightTransactionStateGeneration) StartNewStageContext(ctx context.Co
 			signedMessage = v.SignedMessage
 			calculatedTxHash = v.TransactionHash
 		}
-		v.stageTriggerError = v.TriggerSubmitTx(ctx, signedMessage, calculatedTxHash)
+
+		toAddress := ""
+		if rsc.InMemoryTx.GetTo() != nil {
+			toAddress = rsc.InMemoryTx.GetTo().String()
+		}
+		v.stageTriggerError = v.TriggerSubmitTx(ctx, signedMessage, calculatedTxHash, toAddress)
```
peterbroadhurst: (question above on why this is needed)

core/go/internal/publictxmgr/transaction_manager.go
```diff
@@ -352,6 +391,103 @@ func (ptm *pubTxManager) WriteNewTransactions(ctx context.Context, dbTX persiste
...
+				PrivateTXID:     binding.Transaction,
+			}
+			persistedSubmissions = append(persistedSubmissions, dbSubmission)
+
+			err = dbTX.DB().
+				Table("public_txn_bindings").
+				Select(`"pub_txn_id"`).
+				Where(`"transaction" IN (?)`, binding.Transaction).
+				Find(&dbSubmission.PublicTxnID).
+				Error
```
peterbroadhurst: Seems inefficient to do a single DB query for each item, vs. an aggregated query (there's already an `IN` clause here)

core/go/internal/publictxmgr/transaction_manager.go
```diff
@@ -352,6 +391,103 @@ func (ptm *pubTxManager) WriteNewTransactions(ctx context.Context, dbTX persiste
...
+
+	// All the nonce processing to this point should have ensured we do not have a conflict on nonces.
+	// It is the caller's responsibility to ensure we do not have a conflict on transaction+resubmit_idx.
+	if len(persistedSubmissions) > 0 {
+		log.L(ctx).Debugf("WriteReceivedTransactions persisted submissions: %d", len(persistedSubmissions))
+		err = dbTX.DB().
+			WithContext(ctx).
+			Table("public_submissions").
+			Create(persistedSubmissions).
+			Error
```
peterbroadhurst: Handling of dups?

core/go/internal/publictxmgr/transaction_orchestrator.go
```diff
@@ -437,6 +481,42 @@ func (oc *orchestrator) pollAndProcess(ctx context.Context) (polled int, total i
...
 		}
 
+		for _, tx := range additional {
+			var txId string
+			err := oc.p.DB().
+				Table("public_txn_bindings").
+				Select(`"transaction"`).
+				Where(`"pub_txn_id" IN (?)`, tx.PublicTxnID).
+				Find(&txId).
+				Error
```
peterbroadhurst: We've just done this query above.

core/go/internal/sequencer/common/commonmocks.go
peterbroadhurst: Do questions on the deviation from the standard pattern:
1. It's checked in (bigger point)
2. It's in a package outside of mocks (I assume removed from coverage individually)

core/go/internal/sequencer/common/engine_integration.go
```diff
@@ -0,0 +1,334 @@
...
+// Here we define the subset that we rely on in this package
+
+type Hooks interface {
+	GetBlockHeight() int64
+	GetNodeName() string
+}
+
+type EngineIntegration interface {
+	// WriteLockStatesForTransaction is a method that writes a lock to the states for a transaction
+	WriteLockStatesForTransaction(ctx context.Context, txn *components.PrivateTransaction) error
```
peterbroadhurst: Does `Write` mean write in this case? I'm trying to work out if we have a world where we _persist_ locks - because that would be a significant decision.

core/go/internal/sequencer/common/engine_integration.go
```diff
@@ -0,0 +1,334 @@
...
+
+	ResetTransactions(ctx context.Context, transactionID uuid.UUID)
+
+	GetStateLocks(ctx context.Context) ([]byte, error)
```
peterbroadhurst: Feels like this could mean a lot of things - particularly as it returns `[]byte`

core/go/internal/sequencer/common/engine_integration.go
```diff
@@ -0,0 +1,334 @@
...
+type Hooks interface {
+	GetBlockHeight() int64
+	GetNodeName() string
+}
+
+type EngineIntegration interface {
+	// WriteLockStatesForTransaction is a method that writes a lock to the states for a transaction
+	WriteLockStatesForTransaction(ctx context.Context, txn *components.PrivateTransaction) error
+
+	ResetTransactions(ctx context.Context, transactionID uuid.UUID)
```
peterbroadhurst: What's the purpose and use of this function?

core/go/internal/sequencer/common/engine_integration.go
```diff
@@ -0,0 +1,334 @@
...
+	"fmt"
+
+	"github.com/LF-Decentralized-Trust-labs/paladin/common/go/pkg/i18n"
+	"github.com/LF-Decentralized-Trust-labs/paladin/common/go/pkg/log"
+	"github.com/LF-Decentralized-Trust-labs/paladin/core/internal/components"
+	"github.com/LF-Decentralized-Trust-labs/paladin/core/internal/msgs"
+	"github.com/LF-Decentralized-Trust-labs/paladin/sdk/go/pkg/pldtypes"
+	"github.com/LF-Decentralized-Trust-labs/paladin/toolkit/pkg/prototk"
+	"github.com/google/uuid"
+	mock "github.com/stretchr/testify/mock"
```
peterbroadhurst: Definitely do not think this package should be pulled into runtime code modules.

core/go/internal/sequencer/common/engine_integration.go
```diff
@@ -0,0 +1,334 @@
...
+
+	// Some validation that we are confident we can execute the given attestation plan
+	for _, attRequest := range transaction.PostAssembly.AttestationPlan {
+		switch attRequest.AttestationType {
+		case prototk.AttestationType_ENDORSE:
+		case prototk.AttestationType_SIGN:
+		case prototk.AttestationType_GENERATE_PROOF:
+			errorMessage := "AttestationType_GENERATE_PROOF is not implemented yet"
+			log.L(ctx).Error(errorMessage)
+			return nil, i18n.NewError(ctx, msgs.MsgSequencerInternalError, errorMessage)
```
peterbroadhurst: Is this a functional regression in the branch at this point in time?

e.g. do we not support domains like Zeto?

core/go/internal/sequencer/common/request.go
```diff
@@ -0,0 +1,67 @@
...
+	clock            Clock
+	timeout          Duration
+}
+
+func NewIdempotentRequest(ctx context.Context, clock Clock, timeout Duration, sendRequest func(ctx context.Context, idempotencyKey uuid.UUID) error) *IdempotentRequest {
+	idempotencyKey := uuid.New()
+	r := &IdempotentRequest{
+		idempotencyKey:   idempotencyKey,
+		clock:            clock,
+		send:             sendRequest,
```
peterbroadhurst: Don't see where this is used

core/go/internal/sequencer/coordinator/coordinator.go
```diff
@@ -0,0 +1,600 @@
...
+		syncPoints:                         syncPoints,
+		readyForDispatch:                   readyForDispatch,
+		coordinatorStarted:                 coordinatorStarted,
+		coordinatorIdle:                    coordinatorIdle,
+		nodeName:                           nodeName,
+		metrics:                            metrics,
+		coordinatorEvents:                  make(chan common.Event, 50), // TODO >1 only required for sqlite coarse-grained locks. Should this be DB-dependent?
+		externalQueries:                    make(chan QueryRequest, 10),
+		stopEventLoop:                      make(chan struct{}),
+		coordinatorSelectionBlockRange:     blockRangeSize,
```
peterbroadhurst: Feels like there's a bunch of inputs on this function that are really a `config` and it would be easier to just copy one whole struct in so tuning could expand more easily in the future.

core/go/internal/sequencer/coordinator/coordinator.go
```diff
@@ -0,0 +1,600 @@
...
+		log.L(ctx).Debugf("coordinator event loop waiting for next event")
+		select {
+		case query := <-c.externalQueries:
+			log.L(ctx).Debugf("coordinator handling external query %s", query.Type)
+			if query.Type == QueryTypeDispatchedTransactions {
+				dispatchingTransactionCount := len(c.getTransactionsInStates(ctx, []transaction.State{transaction.State_Dispatched, transaction.State_Submitted}))
+				log.L(ctx).Tracef("coordinator responding to %s: %d dispatched transactions", query.Type, dispatchingTransactionCount)
+				query.Response <- dispatchingTransactionCount
+			} else {
+				log.L(ctx).Errorf("%s", i18n.NewError(ctx, msgs.MsgSequencerInternalError, "external query type unknown").Error())
```
peterbroadhurst: Don't you need to ensure you respond to the query with the error here?

core/go/internal/sequencer/coordinator/coordinator.go
```diff
@@ -0,0 +1,600 @@
...
+	go c.eventLoop(ctx)
+
+	// Start dispatch queue loop
+	go c.dispatchLoop(ctx)
+
+	return c, nil
+
+}
+
+func (c *coordinator) eventLoop(ctx context.Context) {
```
peterbroadhurst: Need to have a `done` channel for these routines to wait for them to stop?

core/go/internal/sequencer/coordinator/coordinator.go
```diff
@@ -0,0 +1,600 @@
...
+		}
+	} else if c.domainAPI.ContractConfig().GetCoordinatorSelection() == prototk.ContractConfig_COORDINATOR_ENDORSER {
+		// E.g. Pente
+		// Make a fair choice about the next coordinator, but for now just choose the node this request arrived at
+		if len(c.originatorNodePool) == 0 {
+			log.L(ctx).Warnf("no pool to select a coordinator from yet")
+			return "", nil
+		} else {
+			// Round block number down to the nearest block range (e.g. block 1012, 1013, 1014 etc. all become 1000 for hashing)
+			effectiveBlockNumber := c.currentBlockHeight - (c.currentBlockHeight % c.coordinatorSelectionBlockRange)
```
peterbroadhurst: Interesting in learning how inactive coordinators get evicted from the `originatorNodePool` in N of M HA scenarios

core/go/internal/sequencer/coordinator/coordinator.go
```diff
@@ -0,0 +1,600 @@
...
+		c.originatorNodePool = append(c.originatorNodePool, originatorNode)
+		slices.Sort(c.originatorNodePool)
+	}
+}
+
+// TODO consider renaming to setDelegatedTransactionsForOriginator to make it clear that we expect originators to include all inflight transactions in every delegation request and therefore this is
+// a replace, not an add.  Need to finalize the decision about whether we expect the originator to include all inflight delegated transactions in every delegation request. Currently the code assumes we do so need to make the spec clear on that point and
+// record a decision record to explain why.  Every  time we come back to this point, we will be tempted to reverse that decision so we need to make sure we have a record of the known consequences.
+// originator must be a fully qualified identity locator otherwise an error will be returned
+func (c *coordinator) addToDelegatedTransactions(ctx context.Context, originator string, transactions []*components.PrivateTransaction) error {
```
peterbroadhurst: Is this the whole list of full `PrivateTransaction` object that needs to be passed around? ... or surely just the TXIDs and originator information are passed.

core/go/internal/sequencer/coordinator/coordinator.go
```diff
@@ -0,0 +1,600 @@
...
+	for _, txn := range c.transactionsByID {
+		if !nonMatchingStates[txn.GetState()] {
+			matchingTxns = append(matchingTxns, txn)
+		}
+	}
+	return matchingTxns
+}
+
+// MRW TODO - is there a reason we need to find by nonce and not by TX ID?
+func (c *coordinator) findTransactionBySignerNonce(ctx context.Context, signer *pldtypes.EthAddress, nonce uint64) *transaction.Transaction {
```
peterbroadhurst: What's the scenario for this function?

I can imagine scenarios like "Maybe this public transaction was submitted as part of this private transaction", but it's not clear to me in isolation here.

common/go/pkg/log/log.go
```diff
@@ -118,6 +120,15 @@ func WithLogField(ctx context.Context, key, value string) context.Context {
 	return WithLogger(ctx, loggerFromContext(ctx).WithField(key, value))
 }
 
+// WithComponent adds the specified component to the logger in the context
+func WithComponent(ctx context.Context, component Component) context.Context {
```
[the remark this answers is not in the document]
matthew1001: Yeah there's some existing context to this. A couple of PRs have gone to `main` around logging:
 - https://github.com/LFDT-Paladin/paladin/pull/834 (mine)
 - https://github.com/LFDT-Paladin/paladin/pull/839 (Anna's use of mine)

The sequencer PR probably needs to align with `main`. For context, here's an example log entry from `main` today:

```
INFO Upserting state 0x808e0a...6c159f (no create lock) component=domain-ctx-noto pid=1 role=sequencer-0x980caa...09dc9
```

So I think in most cases code will continue to call `log.L(ctx)` because `ctx` will typically be updated once with the extra log component. See [839](ttps://github.com/LFDT-Paladin/paladin/pull/839) for how it's being applied today.

In some cases I'd like a small subset of sequencer logs to have a specific component to make grepping sets of logs easier. Here's an example of the PR as it stands:

```
DEBUG sdr      | 0x3e0747 | *originator.TransactionCreatedEvent | Idle -> Sending component=state pid=1 role=sequencer
```

i.e. an entry that relates to one of the sequencer state machines. But I think the majority of log lines across the codebase will remain `log.L(ctx)`

config/pkg/pldconf/sequencermgr.go
```diff
@@ -0,0 +1,74 @@
...
+	TargetActiveCoordinators      *int              `json:"targetActiveCoordinators"`
```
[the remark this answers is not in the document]
matthew1001: Yeah this is slightly different. I currently have 2 config points for resource management within a node:

 - `targetActiveCoordinators`
   - the target/maximum number of contracts this node should be coordinator for
 - `targetActiveSequencers`
   - the target/maximum number of contracts this node should continue running a sequencer for, which might just mean having the originator in memory ready for originating new transactions.

So for example, a non-notary node will have lots of sequencers for all the Notos it's participating in, but never any coordinators, just originators. In that situation `targetActiveCoordinators` is meaningless.

A notary node will have lots of sequencers for all the Notos it's participating in, but possibly never any originators, just coordinators. In that situation `targetActiveCoordinators` can be used to say "I'm coordinating too many contracts, can someone take over one or more of them". Today this would only really apply to a privacy group where another node can take over, but at some point it will apply to multiple-notary/m-of-n notarized domains.

`targetActiveSequencers` is just a way of saying "page out of memory one of the contracts, I've got too many". This is regardless of whether the node is coordinating those contracts. If a node get's a new transaction for a paged-out sequencer it will ultimately just load it back in to memory and page out another one.

core/go/internal/domainmgr/private_smart_contract.go
matthew1001: Agreed - similar to my previous comment, I think I was using this for a `>0` or not check in the logs. I'll make a note to review our logging towards the end of the review process (possibly post-merge)

core/go/internal/domainmgr/private_smart_contract.go
matthew1001: I've raised https://github.com/LFDT-Paladin/paladin/issues/890 on the V1 project board to make sure we don't miss the job of reviewing logs

operator/config/samples-3-nodes/cert_issuer_selfsigned.yaml
```diff
@@ -0,0 +1,7 @@
+---
```
annamcallister: It looks like this `operator/config/samples-3-nodes` directory is a copy of `operator/config/samples` that you've made for your own testing. It doesn't look like it was intentional to deliver it back?

In particular the generated smart contract deployments and their byte code have been checked in which we need to avoid.

domains/integration-test/build.gradle
```diff
@@ -62,6 +62,8 @@ dependencies {
 }
 
 task copySolidity(type: Copy) {
+    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
```
annamcallister: Same question as earlier about choice of duplicates strategy

examples/common/contracts/abis/AribtraryWidgets.json
```diff
@@ -0,0 +1,119 @@
+{
```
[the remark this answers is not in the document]
matthew1001: I think `.gitignore` is present, but I'm wondering if it was temporarily absent at some point and that caused the ABIs to get committed. I've removed them now

core/go/internal/publictxmgr/transaction_manager.go
```diff
@@ -309,10 +314,20 @@ func (ptm *pubTxManager) WriteNewTransactions(ctx context.Context, dbTX persiste
...
+				To:              txi.To,
+				Data:            txi.Data,
+				From:            txi.From,
+				PublicTxOptions: txi.PublicTxOptions,
+			},
+			PublicTxOptions: txi.PublicTxOptions,
+		}
 	}
 	// All the nonce processing to this point should have ensured we do not have a conflict on nonces.
 	// It is the caller's responsibility to ensure we do not have a conflict on transaction+resubmit_idx.
```
annamcallister: not your comments, but these comments look very old and out of date (nonce assignment is not after this point) so it's a chance to just remove these lines
```suggestion
```

core/go/internal/publictxmgr/transaction_orchestrator.go
```diff
@@ -426,6 +427,49 @@ func (oc *orchestrator) pollAndProcess(ctx context.Context) (polled int, total i
...
+					log.L(ctx).Warnf("Orchestrator poll and process: context cancelled while retrieving binding for %d: %s", tx.PublicTxnID, err)
+					continue
+				}
+				tx.Binding.Transaction = txIdUUID
+			}
+			if oc.sequencerManager == nil {
+				log.L(ctx).Warnf("Orchestrator poll and process: sequencer manager is nil for %d", tx.PublicTxnID)
+			} else {
+				if tx.To == nil {
+					log.L(ctx).Warnf("Orchestrator poll and process: to address is nil for %d", tx.PublicTxnID)
```
annamcallister: to will legitimately be nil for a contract deploy- a warn level log feels a bit extreme

core/go/internal/sequencer/coordinator/transaction/ready_for_dispatch.go
```diff
@@ -0,0 +1,156 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/LF-Decentralized-Trust-labs/paladin/common/go/pkg/i18n"
+	"github.com/LF-Decentralized-Trust-labs/paladin/common/go/pkg/log"
+	"github.com/LF-Decentralized-Trust-labs/paladin/core/internal/msgs"
+)
+
+func (t *Transaction) isNotReady() bool {
```
annamcallister: another `isNot` function- this might just be me but I feel like I'd expect to read it as `!is`

core/go/internal/txmgr/transaction_query.go
```diff
@@ -295,14 +290,21 @@ func (tm *txManager) GetResolvedTransactionByID(ctx context.Context, id uuid.UUI
 }
 
 func (tm *txManager) GetTransactionByID(ctx context.Context, id uuid.UUID) (*pldapi.Transaction, error) {
```
annamcallister: This function can be refactored to use `GetTransactionByIDWithDBTX` passing in `tm.p.NOTX()`

core/go/noderuntests/componenttest/component_test.go
annamcallister: I'm not sure on the directory structure/names here- might be worth a chat at a suitable time

core/go/noderuntests/coordinationtest/coordination_test.go
```diff
@@ -0,0 +1,1166 @@
+/*
+ * Copyright © 2024 Kaleido, Inc.
```
annamcallister:
```suggestion
 * Copyright © 2025 Kaleido, Inc.
```

core/go/noderuntests/componenttest/utils_for_test.go
```diff
@@ -0,0 +1,199 @@
+/*
+ * Copyright © 2024 Kaleido, Inc.
```
annamcallister:
```suggestion
 * Copyright © 2025 Kaleido, Inc.
```

core/go/internal/sequencer/coordinator/transaction/assembling.go
```diff
@@ -0,0 +1,297 @@
...
+		// however we go here, the answer to the question is "false" because there is no pending request to timeout but log this as it is a strange situation
+		// and might be an indicator of another issue
+		log.L(ctx).Infof("assembleTimeoutExceeded called on transaction %s with no pending assemble request", t.ID)
+		return false
+	}
+	return t.clock.HasExpired(t.pendingAssembleRequest.FirstRequestTime(), t.assembleTimeout)
+
+}
+
+func (t *Transaction) isNotAssembled() bool {
```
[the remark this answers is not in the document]
matthew1001: Yeah for a few of these I'll probably defer to the original sequencer PR https://github.com/LFDT-Paladin/paladin/pull/673. Generally speaking I've made minimal changes to the state machine(s) (some exceptions outlined in this PR's description) but I'm inclined to leave the current state machine design relatively untouched - at least for relatively minor comments/issues such as this one. We might want to iterate on it in the future, but the less I mod it for now the easier it will be for John to comment on the changes I _*have*_ made.

core/go/internal/sequencer/coordinator/transaction/ready_for_dispatch.go
matthew1001: Same reply as https://github.com/LFDT-Paladin/paladin/pull/704#discussion_r2510274627 I think

core/go/internal/publictxmgr/submission_writer.go
```diff
@@ -53,6 +61,29 @@ func (sw *submissionWriter) runBatch(ctx context.Context, tx persistence.DBTX, v
 	if err != nil {
 		return nil, err
 	}
+
+	// The sequencer needs to distribute the public submission to all relevant nodes. We submit a reliable message
+	// under the same DBTX as the local persist.
```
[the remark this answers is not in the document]
matthew1001: For the discussion about distributing TX info, I've added a diagram to the docs under the latest commit https://github.com/LFDT-Paladin/paladin/pull/704/commits/f5761489cce29ce2a401f671d9d576b5aed6bfa8 as a way to force the discussion around what "correct" looks like. Shall we try to find time for an in-person sync on this topic?

examples/common/contracts/abis/AribtraryWidgets.json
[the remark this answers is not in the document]
matthew1001: This should all be sorted now so I'll mark this conversation resolved

core/go/componenttest/component_test.go
peterbroadhurst: This test was the primary test coverage for the previous private transaction manager, justifying (partially) some debt on the coverage in that codebase. Noting that resolving that debt was primarily pushed to be a significant part of the exercise of building the new sequencer engine. So if you can create an issue to describe(related to earlier comment) the test plan, and what stage we're at in the journey so we can ensure before release we get to robust functional, regression, and code coverage.

core/go/internal/components/publictxmgr.go
```diff
@@ -67,6 +67,10 @@ type PublicTxManager interface {
 	ValidateTransaction(ctx context.Context, dbTX persistence.DBTX, transaction *PublicTxSubmission) error
 	// Write a set of validated transactions to the public TX mgr database, notifying the relevant orchestrator(s) to wake, assign nonces, and start the submission process
 	WriteNewTransactions(ctx context.Context, dbTX persistence.DBTX, transactions []*PublicTxSubmission) ([]*pldapi.PublicTx, error)
+	// Write a set of received transactions to the public TX mgr database. These are informational transactions, not for us to act on
+	WriteReceivedTransactions(ctx context.Context, dbTX persistence.DBTX, transactions []*pldapi.PublicTxToDistribute) ([]*pldapi.PublicTx, error)
+	// Write a set of received public transaction submissions to the public TX mgr database. These are informational submissions, not for us to act on
+	WriteReceivedPublicTransactionSubmissions(ctx context.Context, dbTX persistence.DBTX, submissions []*pldapi.PublicTxToDistribute) ([]*pldapi.PublicTx, error)
```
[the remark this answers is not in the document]
peterbroadhurst: Noting we still need to have and document this architectural discussion.

core/go/internal/domainmgr/private_smart_contract.go
peterbroadhurst: Will be a non-trivial, but important item, to really validate what Info/Debug/Trace should be with real scenarios. Let's not consider it an optional path on the V1 journey.

If you found turning something into trace mode too hard, and it lead to adding debug entries instead - we really *must* solve that. Because it will be a pattern all developers repeat over time.

So while it's fresh, raising a must-fix item for whichever test/env you're testing with that is hard to put into trace mode I think is just as (if not more) important than the reconciling of log entries.

core/go/internal/identityresolver/identityresolver.go
```diff
@@ -175,13 +175,33 @@ func (ir *identityResolver) ResolveVerifierAsync(ctx context.Context, lookup str
...
-		})
+		}, errChan)
+
+		go func() {
+			// Handle transport-layer errors (e.g. the transport plugin couldn't reach the remote node)
+			// If sending fails, fail the request in the same manor we would if the remote node had returned an error response
+			select {
+			case <-ctx.Done():
+				return
+			case err, ok := <-errChan:
```
peterbroadhurst: I've not seen the pattern of "close means good". I'm overall just confused why this isn't a synchronous function `SendSync` - and we just call it in a go routine. The interface just feels half-formed to me.

core/go/internal/sequencer/common/commonmocks.go
peterbroadhurst: Reminder this file should not have been checked in

core/go/internal/sequencer/coordinator/coordinator.go
```diff
@@ -0,0 +1,620 @@
...
+			})
+
+			// We almost never need to wait for the state machine's event loop to process the update to State_Dispatched
+			// but if we hit the max dispatch ahead limit after dispatching this transaction we do, because we can't be sure
+			// in-flight will be accurate on the next loop round
+			if len(c.inFlightTxns)+dispatchedAhead >= c.maxDispatchAhead {
+				for c.inFlightTxns[tx.ID] == nil {
+					c.inFlightMutex.Wait()
+				}
+				dispatchedAhead = 0
```
peterbroadhurst: I don't really understand what `dispatchedAhead` means here - we're arbitrarily setting it to zero here, but just because we waited for a state machine to update.

If all we're doing is propagating items from one go routine to a state machine, that's what a Go channel size is for - so we wouldn't need this layer.

So there must be something else going on here, and I've not understood it yet.

core/go/internal/sequencer/coordinator/coordinator.go
```diff
@@ -0,0 +1,620 @@
...
+	coordinator := ""
+	if c.domainAPI.ContractConfig().GetCoordinatorSelection() == prototk.ContractConfig_COORDINATOR_STATIC {
+		// E.g. Noto
+		if c.domainAPI.ContractConfig().GetStaticCoordinator() == "" {
+			return "", fmt.Errorf("static coordinator mode is configured but static coordinator node is not set")
+		}
+		log.L(ctx).Debugf("coordinator %s selected as next active coordinator in static coordinator mode", c.domainAPI.ContractConfig().GetStaticCoordinator())
+		// If the static coordinator returns a fully qualified identity extract just the node name
+		if strings.Contains(c.domainAPI.ContractConfig().GetStaticCoordinator(), "@") {
+			parts := strings.Split(c.domainAPI.ContractConfig().GetStaticCoordinator(), "@")
```
peterbroadhurst: This identity parsing code exists in exactly one place, and must be used from there.

core/go/internal/sequencer/coordinator/select_transaction.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+func (c *coordinator) AddTransactionToBackOfPool(ctx context.Context, txn *transaction.Transaction) {
+	c.pooledTransactionsMutex.Lock()
+	defer c.pooledTransactionsMutex.Unlock()
+	c.pooledTransactions = append(c.pooledTransactions, txn)
+}
+
+/* Functions of the TransactionPool interface required by the transactionSelector */
+func (c *coordinator) GetNextPooledTransaction(ctx context.Context) *transaction.Transaction {
```
peterbroadhurst:
```suggestion
func (c *coordinator) PopNextPooledTransaction(ctx context.Context) *transaction.Transaction {
```

peterbroadhurst: Or could be remove... but think it needs to be clear this isn't a getter, but it mutates the queue by removing the first element.

core/go/internal/sequencer/coordinator/select_transaction.go
```diff
@@ -0,0 +1,206 @@
...
+	return selector
+}
+
+type originatorLocator struct {
+	node     string
+	identity string
+	txID     uuid.UUID
+}
+
+// Determine if there is still an active assembler, so we know if it's safe to select a new transaction to start processing
```
peterbroadhurst: Have really struggled to read this function.

Title suggest it's asking a question `transactionStillBeingAssembled`?.

Function has a `TODO` to change it - which sound very true... because it's function seems to be able trying to set a variable `currentAssemblingOriginator` to `nil` based on processing a `TransactionStateTransitionEvent`.

So overall it feels like this function is a placeholder for a core part of the state machine, which is processing state transition event.

core/go/internal/sequencer/coordinator/snapshot.go
```diff
@@ -0,0 +1,130 @@
...
+		case transaction.State_Endorsement_Gathering:
+			fallthrough
+		case transaction.State_PreAssembly_Blocked:
+			fallthrough
+		case transaction.State_Assembling:
+			fallthrough
+		case transaction.State_Pooled:
+			pooledTransactions = append(pooledTransactions, &common.Transaction{
+				ID: txn.ID,
+			})
```
peterbroadhurst: I'm pleased (per earlier comments) it's just the transaction ID that's managed here, but `common.Transaction` seems like a wrong description of the object.

Both the `common` package feels very odd - which maybe comes to earlier comments that this really isn't part of `core` and is thinking it's big enough to have it's own package structure.

But also the `Transaction` is odd, if it's just a `TransactionReference`, or a `SnapshotPooledTransaction` (I guess with the ability for extensibility in the future - so it's not just an array of IDs.)

core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -0,0 +1,491 @@
...
+				Event_HeartbeatReceived: {
+					Transitions: []Transition{{
+						To: State_Observing,
+					}},
+				},
+				Event_EndorsementRequested: { // We can assert that someone else is actively coordinating if we're receiving these
+					Transitions: []Transition{{
+						To: State_Observing,
+					}},
+				},
```
peterbroadhurst: Feels like an over simplification of an area we've spent a lot of time talking about. When we're in a state (temporarily, or due to network partitioning) that we have a "split brain" it's not clear to me that the situation is resolved ideally by each side simply going silent when they notice it happening.

Now we're not in a terrible situation for split brain, because the blockchain of course is the arbiter that avoids problems. But we don't want some very temporary thing to result in a cross-in-the-post scenario where both sides go idle.

I wonder if maybe the best way to help readers of this critical state machine, is to annotate the architecture diagram / document on the protocol with named transitions, and put references across.

That way complex topics like this can be documented in a single place, then code-readers get the trivial comment (like this one) combined with a pointer to the right bit of the protocol spec.

core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -0,0 +1,491 @@
...
+						},
+					},
+				},
+			},
+		},
+		State_Standby: {
+			Events: map[EventType]EventHandler{
+				common.Event_HeartbeatInterval: {},
+				Event_TransactionsDelegated:    {},
+				Event_NewBlock: {
```
peterbroadhurst: Surely we'd only do this on an event that is `NewBlockRange`?

core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -0,0 +1,491 @@
...
+			OnTransitionTo: action_StopHeartbeating,
+			Events: map[EventType]EventHandler{
+				common.Event_HeartbeatInterval: {
+					Transitions: []Transition{{
+						To: State_Idle,
+						If: guard_ClosingGracePeriodExpired,
+					}},
+				},
+			},
+		},
```
peterbroadhurst: If these questions in comments aren't answered in the architecture currently (I know we've discussed them in whiteboard interactions), then probably critical to think about them in the protocol specification as well as the code.

core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -0,0 +1,491 @@
...
+	}
+	return nil
+}
+
+func (c *coordinator) evaluateTransitions(ctx context.Context, event common.Event, eventHandler EventHandler) error {
+	sm := c.stateMachine
+
+	for _, rule := range eventHandler.Transitions {
+		if rule.If == nil || rule.If(ctx, c) { //if there is no guard defined, or the guard returns true
+			log.L(log.WithLogField(ctx, common.SEQUENCER_LOG_CATEGORY_FIELD, common.CATEGORY_STATE)).Debugf("coord    | %s   | %T | %s -> %s", c.contractAddress.String()[0:8], event, sm.currentState.String(), rule.To.String())
```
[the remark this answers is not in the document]
peterbroadhurst: I note there's a whole conversation to be had about this, after a great recent discussion I had with @onelapahead 

Paladin might want overall to move to doing a lot more of this, and maybe move to `github.com/uber-go/zap` where a big efficiency gain can be obtained by avoiding `fmt` printing with reflection, and instead using fields for the log entries. But that's a much larger job/discussion.

I mention it only because getting the `log` library convention right, might aid the transition later.

core/go/internal/sequencer/coordinator/transaction/assembling.go
```diff
@@ -0,0 +1,313 @@
...
+func (t *Transaction) revertTransaction(ctx context.Context, revertReason string) error {
+	var tryFinalize func()
+	tryFinalize = func() {
+		t.syncPoints.QueueTransactionFinalize(ctx, t.Domain, pldtypes.EthAddress{}, t.originator, t.ID, revertReason,
+			func(ctx context.Context) {
+				log.L(ctx).Debugf("finalized deployment transaction: %s", t.ID)
+			},
+			func(ctx context.Context, err error) {
+				log.L(ctx).Errorf("error finalizing deployment: %s", err)
+				tryFinalize()
```
peterbroadhurst: Tight loop?

core/go/internal/sequencer/coordinator/transaction/pooled.go
```diff
@@ -0,0 +1,139 @@
...
+				//This should be relatively rare so worth logging as an info
+				log.L(ctx).Infof("dependency %s not found in memory for transaction %s", dependencyID, t.ID)
+				continue
+			}
+
+			//TODO this should be idempotent.
+			if dependencyTxn.PreAssembly.Dependencies == nil {
+				dependencyTxn.PreAssembly.Dependencies = &pldapi.TransactionDependencies{}
+			}
+			dependencyTxn.PreAssembly.Dependencies.PrereqOf = append(dependencyTxn.PreAssembly.Dependencies.PrereqOf, t.ID)
```
peterbroadhurst: I'm not following how these get updated when the grapher gets reset for complex scenarios. This is a bi-directional dependency on the transactions in a linked list, but there's also the grapher object, and the functions on the grapher object take arrays of transactions for ordering. That's a complex inter-related set of order-management objects on these transactions.

It either needs some very careful explanation and health-warnings (as well as tests) to make sure engineers don't make changes that brake the interrelations....

Or it could just be moved into the grapher and thus simplified.

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -0,0 +1,556 @@
...
+	Event_RequestTimeoutInterval                                                         // event emitted by the state machine on a regular period while we have pending requests
+	Event_StateTransition                                                                // event emitted by the state machine when a state transition occurs.  TODO should this be a separate enum?
+	Event_HeartbeatInterval                                                              // event emitted by the state machine on a regular period while we have pending requests
+	Event_AssembleTimeout                                                                // the assemble timeout period has passed since we sent the first assemble request
+)
+
+type StateMachine struct {
+	currentState    State
+	lastStateChange time.Time
+}
```
peterbroadhurst: Does seem like this should be a generic thing shared across state machines in the codebase.

That would allow over time:
- Enforcement of common naming/constructs
- Consistent logging for this critical system to diagnose
- RPC debug interfaces to query state machine state to be built once

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -0,0 +1,556 @@
...
+func init() {
+	// Initialize state definitions in init function to avoid circular dependencies
+	stateDefinitionsMap = map[State]StateDefinition{
+		State_Initial: {
+			Events: map[EventType]EventHandler{
+				Event_Received: { //TODO rename this event type because it is the first one we see in this struct and it seems like we are saying this is a definition related to receiving an event (at one level that is correct but it is not what is meant by Event_Received)
+					Transitions: []Transition{
+						{
+							To: State_Pooled,
+							If: guard_And(guard_Not(guard_HasUnassembledDependencies), guard_Not(guard_HasUnknownDependencies)),
```
peterbroadhurst: Unclear to me what event changes `HasUnassembledDependencies`, and which states you are allowed to be in where it is valid for dependencies to be changed.

core/go/internal/sequencer/coordinator/transaction/transactionmocks.go
peterbroadhurst: Another mock to clean up that shouldn't be checked in

core/go/internal/sequencer/originator/originator.go
```diff
@@ -0,0 +1,205 @@
...
+	err = newTxn.HandleEvent(ctx, createdEvent)
+	if err != nil {
+		log.L(ctx).Errorf("error handling CreatedEvent for transaction %s: %v", txn.ID.String(), err)
+		return err
+	}
+	return nil
+}
+
+func (o *originator) transactionsOrderedByCreatedTime(ctx context.Context) ([]*transaction.Transaction, error) {
+	//TODO are we actually saving anything by transactionsOrdered being an array of IDs rather than an array of *transaction.Transaction
```
peterbroadhurst: Does seem like `transactionsOrdered` should just be a pointer array

core/go/internal/sequencer/originator/originatormocks.go
peterbroadhurst: Another mock incorrectly checked in

core/go/internal/sequencer/originator/state_machine.go
```diff
@@ -0,0 +1,347 @@
...
+	"github.com/google/uuid"
+)
+
+type State int
+type EventType = common.EventType
+
+const (
+	State_Idle      State = iota //Not acting as a originator and not aware of any active coordinators
+	State_Observing              //Not acting as a originator but aware of a node (which may be the same node) acting as a coordinator
+	State_Sending                //Has some transactions that have been sent to a coordinator but not yet confirmed TODO should this be named State_Monitoring or State_Delegated or even State_Sent.  Sending sounds like it is in the process of sending the request message.
```
peterbroadhurst: `State_Inflight` maybe - agree with feedback, but not sure the proposals are much clearer.

- `State_Monitoring` - sounds identical to `State_Observing` to me
- `State_Delegated` - the observer never delegates responsibility, quite the opposite in the model (it's always responsible). It delegates transactions individually, true - but that feels very different to me.
- `State_Sent` - it might still be sending, because it's an aggregate state across everything it's doing

core/go/internal/sequencer/originator/state_machine.go
```diff
@@ -0,0 +1,347 @@
...
+		err = o.confirmTransaction(ctx, event.From, event.Nonce, event.Hash, event.RevertReason)
+	default:
+		log.L(ctx).Debugf("no action defined for event %T", event)
+
+	}
+	if err != nil {
+		log.L(ctx).Errorf("error applying event %v: %v", event.Type(), err)
+	}
+	return err
+}
```
peterbroadhurst: Same again (combined with strong feedback on common-ing up the framework) that we should push this into separate functions in the state machine definition.

core/go/internal/sequencer/originator/test_utils.go
peterbroadhurst: Think I've given the feedback elsewhere, that test utils should be either:
- In a `originatortest` package - if it's needed by _other_ packages
- In a file ending in `_test.go` - if it's just for this package

core/go/internal/sequencer/originator/transaction/assembling.go
```diff
@@ -0,0 +1,102 @@
...
+		return i18n.NewError(ctx, msgs.MsgSequencerInternalError, "No assemble request found")
+	}
+
+	requestID := txn.latestAssembleRequest.requestID
+
+	// The following could be offloaded to a separate goroutine because the response is applied to the state machine via an event emission
+	// However, we do pass the preAssembly by pointer so there may be a need to add locking or pass by value if we off load to a separate thread
+	// lets keep it synchronous for now given that the whole contract is single threaded on the assemble stage anyway, this is unlikely to have a huge negative impact
+	// but from a flow of data perspective and the state machine logic, it _could_ be converted to async
+	postAssembly, err := txn.engineIntegration.AssembleAndSign(ctx, txn.ID, txn.PreAssembly, txn.latestAssembleRequest.stateLocksJSON, txn.latestAssembleRequest.coordinatorsBlockHeight)
```
peterbroadhurst: `stateLocksJSON` feels clunky here to be passed on the boundary. I assume this is the domain-context data exchange format. If we're using it as such, I think it should be modeled so nodes know what they're sending/receiving (and ensured to be as trimmed down as possible for the bundled sending across nodes).

It's as important to version as anything else on the gRPC interface.

core/go/internal/sequencer/originator/transaction/assembling.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		log.L(ctx).Errorf("failed to assemble and sign transaction: %s", err)
+		//This should never happen but if it does, the most likely cause of failure is an error in the local domain code or state machine logic so best thing to abend the originator state machine
+		return err
```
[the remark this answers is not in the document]
peterbroadhurst: I note this "should never happen" is something that has in reality happened a lot over the last 18m and needed to be handled a lot. It's fair to observe that many of those cases are code bugs in domain implementations and extensions, but I don't think all of them have been.

core/go/internal/sequencer/originator/transaction/state_machine.go
```diff
@@ -0,0 +1,586 @@
...
+							To: State_Parked,
+							On: action_SendAssembleParkResponse,
+						},
+					},
+				},
+				Event_CoordinatorChanged: {
+					//would be very strange to have missed a bunch of heartbeats and switched coordinators if we recently received an assemble request but it is possible so we need to handle it
+					Transitions: []Transition{
+						{
+							To: State_Delegated,
```
peterbroadhurst: Feels very odd to go straight to `State_Delegated` here - surely we go to `State_Pending` (although that's under discussion) so we can send the delegation off again.

Or maybe this transition has an `On` function in reality that performs `reDelegate`

core/go/internal/sequencer/originator/transaction/state_machine.go
```diff
@@ -0,0 +1,586 @@
...
+			},
+		},
+		State_Prepared: {
+			Events: map[EventType]EventHandler{
+				Event_Dispatched: {
+					//Note: no validator here although this event may or may not match the most recent dispatch confirmation response.
+					// It is possible that we timed out  on Prepared state, delegated to another coordinator, got as far as prepared again and now just learning that
+					// the original coordinator has dispatched the transaction.
+					// We can't do anything to stop that, but it is interesting to apply the information from event to our state machine because we don't know which of
+					// the many base ledger transactions will eventually be confirmed and we are actually not too fussy about which one does
```
peterbroadhurst: I made the comment on one of the other state machines, that this is really critical stuff these comments. It's protocol specification stuff, not code.

I think these comments are great examples of where we should maintain the comments in the spec along with the state diagram, and refer to each line in the graph by a cross-reference. Maybe even force that.

We could get clever and have the comments in the code, to generate the docs - but that might be a step further than we need/can-afford here.

core/go/internal/sequencer/sequencer.go
```diff
@@ -0,0 +1,1000 @@
...
+func (sMgr *sequencerManager) pollForIncompleteTransactions(ctx context.Context, rePollInterval time.Duration) {
+	// Repeat getting pending transactions until none are returned. Run in a goroutine to avoid blocking the main thread
+	go func() {
+	waitForIndexerReady:
+		for {
+			// On startup we can't assemble any transactions without having a confirmed block height so
+			// wait until the indexer is ready
+			_, err := sMgr.components.BlockIndexer().GetConfirmedBlockHeight(ctx)
+			if err == nil {
+				break
```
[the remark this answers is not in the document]
peterbroadhurst: (we should at least log it, even if the right thing to do is to ignore it and continue - but I suspect we should either return if the error is only cancelled context because of internal retry, or _we_ should retry)

core/go/internal/sequencer/sequencer.go
```diff
@@ -0,0 +1,1000 @@
...
+		Address: *tx.To,
+		Intent:  intent,
+	}, &txi.ResolvedTransaction, true)
+}
+
+// HandleNewTx synchronously receives a new transaction submission
+// TODO this should really be a 2 (or 3?) phase handshake with
+//   - Pre submit phase to validate the inputs
+//   - Submit phase to persist the record of the submission as part of a database transaction that is coordinated by the caller
+//   - Post submit phase to clean up any locks / resources that were held during the submission after the database transaction has been committed ( given that we cannot be sure on completeion of phase 2 that the transaction will be committed)
```
peterbroadhurst: This is an important one. Don't think this is optional in any way to solve

core/go/internal/sequencer/sequencer.go
```diff
@@ -0,0 +1,1000 @@
...
+			TransactionHash: txHash,
+			GasPricing:      []byte(gasPricing),
+			Bindings: []*pldapi.PublicTxBinding{
+				{
+					Transaction: txID,
+				},
+			},
+		}
+
+		senderNode := strings.Split(sender, "@")[1]
```
peterbroadhurst: Another case we need to avoid duplicating this already standardized code

core/go/internal/sequencer/sequencermocks.go
peterbroadhurst: Another checked in mock

core/go/internal/sequencer/transport/transport_writer.go
```diff
@@ -0,0 +1,674 @@
...
+	go tw.loopbackSend(ctx)
+	return nil
+}
+
+func (tw *transportWriter) SendDelegationRequest(
+	ctx context.Context,
+	coordinatorLocator string,
+	transactions []*components.PrivateTransaction,
+	blockHeight uint64,
+) error {
```
peterbroadhurst: Overall feels a bit clunky to have struct defs for these (noting comment on `protobuf` being a potential good way to generate those and have strong commented and typed efficient transfer standards for our protocol), but then not to use them as the input/output format for the code, but rather have flat long lists of input parameters to functions.

core/go/internal/transportmgr/manager.go
```diff
@@ -256,6 +258,12 @@ func (tm *transportManager) LocalNodeName() string {
 // See docs in components package
 func (tm *transportManager) Send(ctx context.Context, send *components.FireAndForgetMessageSend) error {
 	ctx = log.WithComponent(ctx, "transportmanager")
+	return tm.SendWithNack(ctx, send, nil)
+}
+
+// See docs in components package
+func (tm *transportManager) SendWithNack(ctx context.Context, send *components.FireAndForgetMessageSend, errChan chan error) error {
```
peterbroadhurst: Comments on this interface above (around options and callback, rather than a slightly confused sync+channel interface with a close-channel as `success`)

core/go/internal/transportmgr/reliable_msg_handler.go
```diff
@@ -33,13 +33,15 @@ import (
...
-	RMHMessageTypeReceipt             = string(pldapi.RMTReceipt)
-	RMHMessageTypePreparedTransaction = string(pldapi.RMTPreparedTransaction)
-	RMHMessageTypePrivacyGroup        = string(pldapi.RMTPrivacyGroup)
-	RMHMessageTypePrivacyGroupMessage = string(pldapi.RMTPrivacyGroupMessage)
+	RMHMessageTypeAck                         = "ack"
+	RMHMessageTypeNack                        = "nack"
+	RMHMessageTypeStateDistribution           = string(pldapi.RMTState)
+	RMHMessageTypeReceipt                     = string(pldapi.RMTReceipt)
+	RMHMessageTypePublicTransaction           = string(pldapi.RMTPublicTransaction)
+	RMHMessageTypePublicTransactionSubmission = string(pldapi.RMTPublicTransactionSubmission)
```
peterbroadhurst: Noting few places in the comments to talk about these - with an action to lay out the model (public arch docs).

core/go/internal/txmgr/transaction_submission.go
```diff
@@ -358,18 +358,73 @@ func (tm *txManager) PrepareChainedPrivateTransaction(ctx context.Context, dbTX
...
+			Create(transactionDeps).
+			Error
+	}
+	if err != nil {
+		return -1, err
+	}
+	rowsAffected := txInsertResult.RowsAffected
+
+	return rowsAffected, nil
+}
```
peterbroadhurst: This is very significant - a transaction is a decision by users of this node, to do something with their signing key. We've now shared that same DB table with a completely different object - a transaction that is owned by another node, which we somehow want a record of.

If that really is necessary for some reason (feels extremely confusing to me right now), we have to be extremely careful over it.

core/go/pkg/proto/engine.proto
peterbroadhurst: Still got `kata` in here :)

Per comments above, needs reconciling. Things like `state_data_json` that are critical typed protocol defined objects for cross-node versioning.

There's even this comment below to highlight the issue:

> TODO we are using google.protobuf.Any here for TransactionSpecification which is defined in toolkit protos

peterbroadhurst (approved):
Thanks @matthew1001, @hosie and @annamcallister for all the work that's got us to this significant point where we're merging this into the `v1-develop` branch.

There are many important comments in this PR that need to not get lost.

The size of this PR is a reflection of about a year of evolution of this architecture and code so far and we need to move to a model where we have discrete manageable PRs to review.

The following are things that stick in my mind from the couple of person-days over the last weeks I've spent on the code review. I think it is important you make your own judgement @matthew1001 based on the individual comments (maybe 150 of them) what the actual items are to work through though.

### Code/build structure and test coverage

Feels like were in an intermediate state. Lots of things that don't conform with the Paladin norms (checked in mocks, non-conforming UT coverage levels). Not obvious the exact answer at the end, but it's clear we're mid-journey on this.

### State machines

There's an awesome framework and standard evolving for state machines, but it's not fully pulled over the line. It needs to be one framework, and it needs to be consistent.  This includes reference to the spec and how it's documented, and making that standard via a single implementation. However, this is more than skin deep - the threading model and state safety isn't there completely in any implementation. They aren't yet full state machines in implementation, but each one of them is close. There are shortcuts (mainly with TODOs around them), and events that aren't defined that are handled not via the event model. Time-based events stands out for me as something not modeled and critical to the protocol spec in each state machine, but that's just one example.

### Cross-node type definitions

We're stuck halfway between JSON and protobufs, in an ugly place. Either we abandon protobufs, or we follow through and sort out the `prototk` package to have our node<->node protocol standard, as well as our node<->domain standard in the same package.

Given we went on a proto -> JSON -> back to proto journey recently for the signer interface, there's precedent that proto is a good fit for us. So my suggestion is pull it over the line to a defined proto standard. I think that fits nicely with the other points on matching docs to code (in the state machine etc.)

### Error handling

One of the biggest early challenges with the current private tx manager, was tight loops and error management. We addressed it (in parallel to progressing this PR) in the old model, with a big-reset in the code that can cascade up between domains that are interlinked. That works, but it's the blunt instrument that pushes errors all the way back out to apps, and this new protocol is designed to fundamentally supersede and improve on that. So we need the error retry paths - such as non-revert errors, and partial unwinding of the `grapher` to be worked through. Doing the design on paper in parallel I think matters for this one - as well as the scenario-driven testing.

### Cross-node persistence model

How one node influences the persistence of another node by sending things to them (transaction objects not originated on that node, indications of submissions to the blockchain it might find informative). This has some security implications I've noted above, so can't be dismissed. It's also taking an area that's a little confusing already in Paladin, and to my view making it more confusing.

- Before: You consistently didn't have some things you might assume you have, but we can explain why. It's deterministic.
- After: You might in some cases, randomly find you have some of those things you previously expected. But only sometimes, and in a non-deterministic way.

I guess I'm worried we've actually moved backwards not forwards here, but I also think it's extremely solvable and not huge work. We just need to write it down and reason it through. That's the plan you've already suggested, with diagrams and docs. I'm going to do us the service of remaining skeptical here, as someone who feels like they've contributed to the "before" above and knows strongly the whys of it (hence if we find a new "right" I feel I need to be bought in).

### Originator node persistence/scheduler model

This feels like there's a significant additional rev required. We need to page in/out sequencers. We need those sequencers to manage/delegate transactions. Those transactions need to be handled regardless of scale in the DB, but not blow up memory on the node. We need to handle deployment and post-deployment sequencers and transactions.

We've written a lot of code like this - in the public TX manager, and all the generations before/after that for management of these sequence engines. In general we've separated concerns between a scheduler, and a sequence engine. Both have DB code. Both have limits on in-memory scale (horizontal in number of engines, vertical in number of txns per engine).

core/go/internal/domainmgr/private_smart_contract.go
matthew1001: Yes understood, will add a comment to https://github.com/LFDT-Paladin/paladin/issues/890 to ensure this is factored in.

PR #711, opened 2025-06-21
Manifests: Allow domains to report private state completion

### LINKED ISSUE #614: Domain transactions need a better way to mark "completion"

The design of both the Noto and Zeto domains is such that most parties involved in a transaction will get some - but not all - of the private state data. For instance, if I transfer you a coin, then I will send the details on the new state owned by you, but not details on the input state(s). If I do a bulk transfer to many recipients, each one will only know about the coins they received, and not about the coins received by others.

This presents a difficult problem when attempting to decide if a transaction is "complete" and can be processed. Especially when trying to build an event-driven system, Paladin does not know when to deliver the transaction events. Currently it provides only [two modes](https://github.com/LF-Decentralized-Trust-labs/paladin/blob/75dc7b79da059d4a6d567d0a8a7842d7f08b1022/toolkit/go/pkg/pldapi/tx_receipt_listener.go#L38): the default is to block until all state data is received (not likely to ever happen for Noto/Zeto), and the other option is to deliver private events immediately (which only happens to work if the necessary private state data was delivered in a timely manner, but risks sending events that the consumer won't be able to actually process due to missing state data).

The rough solution in my head for this problem is:
1. Add an "info" state to every transaction which describes the transaction inputs, and ensure it is distributed to all participants of the transaction, so they all know what to expect.
2. Allow domains to express an opinion on what states they expect to receive, and therefore have a configurable notion of when the transaction states are "complete". They would _always_ expect to receive the "info" from (1), and would then parse that info to determine their role in the transaction and extrapolate which other states they should receive.
3. Add a new listener configuration that says "trigger when info+required states are available" that does not imply any order (does not block processing new transactions while awaiting private state info).

### Why is this needed?

For reliable processing of confidential transactions, applications must be able to request that they're notified _only_ when a transaction has assembled all of the necessary on-chain and private data in order to be processed.

awrichar:
Resolves #614

This enhancement requires 3 parts:

- [x] Add a new receipt delivery mode (tentatively called `complete_only`) which delivers receipts when they are "complete" (have been joined with all of their private state data). Out-of-order delivery is OK if old receipts become complete by receiving more private data.
- [x] Add a new domain callback so that domains can control their own notion of what "complete" means (because they do not always expect to receive every private state for every transaction).
- [ ] Implement appropriate "complete" checks for Noto and Zeto (note Pente will continue to require 100% of private states for a transaction to be "complete").

### DIFF
--- a/core/go/internal/domainmgr/domain_test.go
+++ b/core/go/internal/domainmgr/domain_test.go
@@ -294,6 +294,7 @@ func TestDomainInitStates(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, td.d, byAddr)
 	assert.True(t, td.d.Initialized())
+	assert.Empty(t, td.d.FixedSigningIdentity())
 
 }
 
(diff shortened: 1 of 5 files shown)


toolkit/go/build.gradle
```diff
@@ -21,7 +21,7 @@ ext {
         include "pkg/**/*.go"
         include "domainstarter/**/*.go"
     }
-    targetCoverage = 100
+    targetCoverage = 99.9
```
awrichar, line 24: I'm confused about how this was ever reporting 100. There are definitely uncovered lines (such as in the "reference" package). I don't think I've reduced coverage (I've actually tried to increase it by adding tests in other packages that had none).

peterbroadhurst:
I'm going to split off a branch to track my changes from this point, and PR back to this branch

PR #713, opened 2025-06-22
Onboard Zeto token implementation with Anonymous KYC

jimthematrix:
The `Zeto_AnonNullifierKyc` token implementation uses a sparse merkle tree to prove KYC membership in the ZKP circuit. This is the initial PR for the basic onboarding of the new token. Operator support and docs will follow in a future PR


domains/zeto/internal/zeto/handler_events.go
```diff
@@ -149,6 +152,37 @@ func (z *Zeto) handleLockedEvent(ctx context.Context, smtTree *merkleTreeSpec, s
 	return nil
 }
 
+func (z *Zeto) handleIdentityRegisteredEvent(ctx context.Context, smtKycTree *merkleTreeSpec, ev *prototk.OnChainEvent, tokenName string, res *prototk.HandleEventBatchResponse) error {
+	var registered IdentityRegisteredEvent
+	if err := json.Unmarshal([]byte(ev.DataJson), &registered); err == nil {
+		txData, err := decodeTransactionData(ctx, registered.Data)
+		if err != nil || txData == nil {
+			newTxID := uuid.New()
+			txId := pldtypes.Bytes32UUIDFirst16(newTxID)
```
peterbroadhurst, line 161: Feels like we should sort this out, if not in this PR then shortly after.

I am worried about it being really confusing having these records with a txid, that isn't actually a transaction anywhere in the system.

However, I can see why you've done it - the `state_confirm_records` DB uses `UUID` as the type so you can't put the public TX in there.

I personally think a zero TXID would be better than generating one - which probably means relaxing a restriction in the Paladin code that receives these. But otherwise I do think it would work.

Do you mind taking a look to see if that's a quick change?

peterbroadhurst (changes requested):
The Zeto code all looks great for me, but I wonder if @Chengxuan or @awrichar should also take a quick look there to be aware of these changes going in.

I made one small comment there, that's take it or leave it.

However, I would like to see if we can make a small tweak to avoid the "fake" TXIDs. Relaxing a protection on domains that exists today (all-zero TXIDs) feels much clearer to me than the potential confusion of these generated ones.

domains/zeto/internal/zeto/handler_events.go
```diff
@@ -149,6 +151,36 @@ func (z *Zeto) handleLockedEvent(ctx context.Context, smtTree *merkleTreeSpec, s
...
+			}
+		}
+		z.recordTransactionInfo(ev, txData, res)
+		if common.IsKycToken(tokenName) {
+			// calculate the Poseidon hash of the public key, to use as the leaf node index and value in the SMT
+			publicKeyHash, err := poseidon.Hash([]*big.Int{registered.PublicKey[0].Int(), registered.PublicKey[1].Int()})
+			if err != nil {
+				return i18n.NewError(ctx, msgs.MsgErrorHandleEvents, fmt.Sprintf("IdentityRegistered. %s", err))
+			}
+			err = z.updateMerkleTree(ctx, smtKycTree.Tree, smtKycTree.Storage, txData.TransactionID, []pldtypes.HexUint256{*pldtypes.MustParseHexUint256("0x" + publicKeyHash.Text(16))})
```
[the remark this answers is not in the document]
awrichar: Maybe there is some overlap in the smt storage, such that each contract isn't properly isolated to its own root? Nothing has jumped out to me immediately, but "key not found" suggests there is an issue here with state consistency across contracts.

domains/zeto/internal/zeto/handler_events.go
```diff
@@ -149,6 +151,36 @@ func (z *Zeto) handleLockedEvent(ctx context.Context, smtTree *merkleTreeSpec, s
...
+		txData, err := decodeTransactionData(ctx, registered.Data)
+		if err != nil || txData == nil {
+			txId := pldtypes.MustParseBytes32("0000000000000000000000000000000000000000000000000000000000000000")
+			log.L(ctx).Infof("IdentityRegistered event [publicKey=%+v] has no tx data. Inserting zero tx ID", &registered.PublicKey)
+			txData = &types.ZetoTransactionData_V0{
+				TransactionID: txId,
+				InfoStates:    []pldtypes.Bytes32{},
+			}
+		}
+		z.recordTransactionInfo(ev, txData, res)
```
awrichar, line 166: I'm not sure there's any transaction info to record, especially if txID might be 0? Do we expect "decodeTransactionData" to actually decode something or not (or we're unsure depending on the transaction)?

domains/zeto/internal/zeto/handler_events.go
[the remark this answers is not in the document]
jimthematrix: this turned out to be a state ID uniqueness issue. while the smt tree root node's hash takes into account the tree name (which has the contract address in it, making it globally unique), the tree node's hash fails to include the tree name. When the same public key gets registered against multiple contracts, the state IDs clashed

domains/zeto/internal/zeto/handler_events.go
jimthematrix: indeed. the call is removed.

domains/integration-test/pvp_test.go
```diff
@@ -406,6 +406,9 @@ func (s *pvpTestSuite) TestNotoForZeto() {
 	})
 	lockedZeto, _ := lockedZetoCoins[0].Data.Hash(ctx)
 
+	// TODO: remove
```
peterbroadhurst, line 409: Would love to see this removed before merge - just a check-loop would be better

peterbroadhurst (approved):
👍 great to see this enabled now in the Paladin layer

domains/integration-test/pvp_test.go
jimthematrix: yep the check-loop was already being done with the `findAvaillableCoins()` call immediately above it. it was added during debugging and isn't needed anymore. removed.

PR #717, opened 2025-06-24
Build new stablecoin example leveraging Zeto+KYC

awrichar:
This is a work in progress. It builds on https://github.com/LF-Decentralized-Trust-labs/paladin/pull/713 to showcase how Zeto could be used for privacy-preserving stablecoins with compliant KYC.

The example passes 🎉 but it may need some more work in honing it down and cleaning up the docs.

Live copy of the docs can be previewed here:
https://awrichar.github.io/paladin/head/tutorials/
https://awrichar.github.io/paladin/head/tutorials/enterprise-stablecoin/


domains/zeto/internal/zeto/handler_events.go
```diff
@@ -172,7 +173,12 @@ func (z *Zeto) handleIdentityRegisteredEvent(ctx context.Context, smtKycTree *co
 			}
 			err = z.updateMerkleTree(ctx, smtKycTree.Tree, smtKycTree.Storage, txData.TransactionID, []pldtypes.HexUint256{*pldtypes.MustParseHexUint256("0x" + publicKeyHash.Text(16))})
 			if err != nil {
-				return i18n.NewError(ctx, msgs.MsgErrorUpdateSMT, "IdentityRegistered", err)
+				// Check if this is a "key already exists" error - if so, treat it as success for idempotency
```
awrichar, line 176: This string parsing was mainly a workaround for bugs with state uniqueness. @jimthematrix do you have a thought on how we should actually handle idempotency if events happen to be replayed? I think we either have to catch an error and ignore (although I'd rather have something more explicit than string parsing), or we have to query before insert.

example/private-stablecoin/README.md
```diff
@@ -0,0 +1,261 @@
+# Example: Enterprise Stablecoin with KYC and Nullifiers
```
awrichar, line 1: Should be "private stablecoin"

awrichar: This README was AI-assisted and could probably be pruned down overall. I feel like it went a little overboard 😅

example/private-stablecoin/src/index.ts
```diff
@@ -0,0 +1,426 @@
...
+  url: "http://127.0.0.1:31548",
+});
+const paladin2 = new PaladinClient({
+  url: "http://127.0.0.1:31648",
+});
+const paladin3 = new PaladinClient({
+  url: "http://127.0.0.1:31748",
+});
+
+async function getBabyjubPublicKey(verifier: PaladinVerifier): Promise<string[]> {
```
awrichar, line 25: Is there a simpler way to do this, or a way to make Paladin/Zeto do it for us? Feels like this utility will be necessary for anyone using this variant of Zeto.

example/private-stablecoin/src/index.ts
```diff
@@ -0,0 +1,426 @@
...
+  } catch (error) {
+    // If balance query fails, return 0 for demo purposes
+    console.warn(`Failed to get ERC20 balance: ${error}`);
+    return 0;
+  }
+}
+
+async function main(): Promise<boolean> {
+  // Generate unique identity names for this run to avoid Merkle tree conflicts
+  const runId = Math.random().toString(36).substring(2, 8);
```
awrichar, line 144: This can be reverted now that Zeto handles multiple contracts better

example/private-stablecoin/src/index.ts
jimthematrix: Agree, the best approach I think is adding a new key spec for the uncompressed form, in addition to the compressed form, to be returned from the `Resolve()` RPC call. I'll work on adding that.

PR #723, opened 2025-06-28
Ensure Noto/Pente instances are upgradeable

awrichar:
Clones are not upgradeable out of the box - use ERC1967 proxy instead.

### DIFF
--- a/solidity/contracts/domains/noto/NotoFactory.sol
+++ b/solidity/contracts/domains/noto/NotoFactory.sol
@@ -2,7 +2,6 @@
 pragma solidity ^0.8.20;
 
 import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
-import {Clones} from "@openzeppelin/contracts/proxy/Clones.sol";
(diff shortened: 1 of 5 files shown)


solidity/contracts/domains/noto/NotoFactory.sol
```diff
@@ -2,7 +2,6 @@
 pragma solidity ^0.8.20;
 
 import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
-import {Clones} from "@openzeppelin/contracts/proxy/Clones.sol";
 import {ERC1967Proxy} from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
```
matthew1001, line 5: Was this just not being used before?

awrichar: I think we had just used Clones in some other places, and propagated the pattern to here... without connecting the dot that it then made the instances non-upgradeable (even though they implement UUPSUpgradeable, it doesn't work unless it's coupled with ERC1967).

PR #724, opened 2025-06-30
pente: treat some plugin errors as revert conditions

### LINKED ISSUE #720: bug: Submitting a Pente transaction with an incorrect input causes infinite retry
### What happened?

I submitted a private blockchain transaction to Pente, and accidentally passed a blank string for a `bytes32` value. Rather than failing, the transaction was stuck in an infinite submission loop, failing (and sucking 300% CPU) with this error:

`[2025-06-27T20:59:51.405Z]  INFO transactionFlow:Action TransactionID='e1579eea-0c09-4dfc-bd0c-ab92d1126b9f' Status='new' LatestEvent='TransactionAssembleFailedEvent' LatestError='PD011802: Error assembling transaction: PD011206: DOMAIN pente returned error: java.util.concurrent.ExecutionException: java.lang.Exception: PD011634: ABI encoding request failed: FF22043: Insufficient data elements on input expected=32 found=0 for ABI encoding of component [0][3][0][0][2]' : Transaction not assembled. Waiting for assembler to return pid=1 role=pctm-loop-0x94f4937d6a8edeba7fcb850a0068373b19292458`

Transaction should have simply failed and been removed from the queue.

[paladin2-truncated.log](https://github.com/user-attachments/files/20957649/paladin2-truncated.log)

_No response_

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
# paste output here
```

</details>

awrichar:
Fixes #720


peterbroadhurst:
I note we did have an API for this I thought 🤔

domains/pente/src/main/java/io/kaleido/paladin/pente/domain/PenteDomain.java
```diff
@@ -282,6 +285,17 @@ protected CompletableFuture<AssembleTransactionResponse> assembleTransaction(Ass
                      setAssemblyResult(AssembleTransactionResponse.Result.REVERT).
                      setRevertReason(e.getMessage()).
                      build());
+         } catch (ExecutionException e) {
+             if (e.getCause() instanceof ErrorResponseException) {
+                // Any error response from a plugin during assembly is considered a revert.
+                // These can stem from things like an invalid ABI or inputs.
+                LOGGER.error(new FormattedMessage("Error response from plugin during assemble for TX {}", request.getTransaction().getTransactionId()), e);
+                return CompletableFuture.completedFuture(AssembleTransactionResponse.newBuilder().
```
peterbroadhurst, line 293: oh - this is just using the spelling that already exists, but in the Java plugin.

peterbroadhurst: Not quite what I'm saying.

I think the only code that can determine if a failure is permanent, is the plugin.

So Paladin should have a spelling to tell the difference explicitly - **not** with PD codes, because those are out of its scope.

- Plugin returns any error that isn't explicitly a "please treat this as a revert": Keep retrying
- Plugin returns an explicit revert: This is a revert

I now understand you've put code into the the Java SDK - to try and guess the difference, and translate that into the Paladin SDK. I disagree with that approach.

I think instead the Java SDK should force the Java plugin to do the right thing.

That could be by defining a specific type of Class of Exception that the code that throw to explicitly revert.

peterbroadhurst: Thanks Andrew, this is super helpful. Yes 👍

awrichar: I've updated this interface with an "error_type" that currently is UNKNOWN or INVALID_INPUT. The encode/decode operations have been adjusted to return INVALID_INPUT in (I think) appropriate cases, and Pente is adjusted to interpret _only this_ error code as a revert.

awrichar: Note that there is an earlier `catch` specifically for EVM execution errors, which was already using the "revert" behavior. This is just adding another `catch` for this other category of failure.

peterbroadhurst (approved):
Thanks for working this through with me, looks great @awrichar 👍

PR #726, opened 2025-07-01
Additional block indexer logs for transactions/events

annamcallister:
There isn't currently a way to tell from the logs what transactions and events we have indexed in a block.


matthew1001 (approved):
LGTM Anna. Did you want to add any extra logs to `MatchUpdateConfirmedTransactions()` or is that not necessary to diagnose confirmation/receipt matching issues in the future?

PR #728, opened 2025-07-01
Additional mitigation for race conditions in rpc_eventstreams_test

awrichar:
May fix https://github.com/LF-Decentralized-Trust-labs/paladin/issues/727


core/go/internal/groupmgr/rpc_eventstreams_test.go
```diff
@@ -405,20 +437,14 @@ func TestRPCEventListenerE2ENack(t *testing.T) {
...
-	// Wait for subscription reply, then trigger reader's NACK logic
-	var sid string
-	select {
-	case sid = <-subIDChan:
-		require.NotEmpty(t, sid)
-	case <-time.After(5 * time.Second):
-		t.Fatal("Subscription timeout")
-	}
+	// Small delay to allow message processing to begin reliably
+	time.Sleep(100 * time.Millisecond)
```
dwertent, line 441: Do we have a better way to wait for the message processing to begin?

awrichar: We could drop this for now and see if the test is improved just with the other mitigations.

dwertent (approved):

PR #731, opened 2025-07-08
sdk: do not block on receipt polling by default

awrichar:
In general, the TypeScript SDK should never block on receiving a transaction receipt. The caller should receive the submitted transaction ID, and can look up the status asynchronously.

To support ease of use for development, returned transaction IDs have been wrapped in "future" objects that support blocking calls "waitForReceipt" and "waitForDeploy" as appropriate.

This is a breaking SDK change, but I believe the current approach is fundamentally wrong (and the migration is relatively easy using the new helpers).

### DIFF
--- a/.github/workflows/build-chart.yaml
+++ b/.github/workflows/build-chart.yaml
@@ -85,7 +85,7 @@ jobs:
 
       - name: E2E Test
         run: |
-          ./gradlew --no-daemon --parallel e2e \
+          ./gradlew --no-daemon e2e \
             -PclusterName=${{ env.CLUSTER_NAME }} \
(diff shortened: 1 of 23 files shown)


matthew1001:
The change seems reasonable given it’s moving the SDK in the right direction. The only thing I wonder about is the spelling, specifically whether Wrapper is the right term. It feels a little generic. Not really a strong opinion, but maybe SyncTransactionWrapper or something else that explains the meaning a little

awrichar:
I updated the terminology to use "TransactionFuture", "NotoFuture", etc - do you think that's better?

matthew1001 (approved):
The change to use `*Future` looks clearer to me, thanks

PR #750, opened 2025-07-22
Stabilizing Examples and SDKs

dwertent:
### Overview

This PR focuses on the **first phase** of improving the stability and testability of the Paladin examples. It lays the groundwork for a consistent developer experience and paves the way for future rollout testing, which will be handled in a **separate PR**.

The primary goal here is to ensure that all examples are up-to-date, work against published SDKs and contracts, and can be tested automatically in CI.

---

### Scope of This PR

✅ This PR **does include**:

* Refactoring the examples directory structure and logic
* Updating examples to use the published SDKs and ABIs
* Adding workflows to automatically test each example in CI
* Adding helper scripts to validate installations and run examples end-to-end

🚫 This PR **does not include**:

* Rollout/upgrade testing (moved to a follow-up PR)
* Metadata files for version compatibility
* Devnet/customnet upgrade workflows

---

#### GitHub Workflows

* [x] Add CI tests for each example (runs on every PR and SDK/contract change)
* [x] Use published SDKs (no need to build locally)
* [x] Validate Helm-based installation (based on getting started guide)
* [x] Add workflow to run all examples end-to-end

#### Scripts

* [x] Script to validate cluster setup (pods, CRs, etc.)
* [x] Script to run all examples for functional validation

#### Examples

* [x] Rename `example` → `examples` (aligning with current usage)
* [x] Add `contracts` mode that downloads published ABIs instead of building Solidity
* [x] Update documentation and scripts to use a consistent installation path
* [x] Add support scripts for:
  * [x] bond
  * [x] event-listener
  * [x] helloworld
  * [x] notarized-tokens
  * [x] privacy-storage
  * [x] private-stablecoin
  * [x] public-storage
  * [x] swap
  * [x] zeto

### DIFF
--- a/examples/notarized-tokens/src/verify-deployed.ts
+++ b/examples/notarized-tokens/src/verify-deployed.ts
@@ -202,60 +202,7 @@ async function main(): Promise<boolean> {
     logger.error(`Error: ${error}`);
     return false;
   }
-
-  // STEP 5: Restore the original state by transferring back
-  logger.log("STEP 5: Restoring original state...");
-  try {
-    const restoreTransferAmount = 50;
-    logger.log(`STEP 5: Transferring ${restoreTransferAmount} units back from Node2 to Node1...`);
-    
-    const restoreTransferReceipt = await cashToken
-      .using(paladinClientNode2)
-      .transfer(verifierNode2, {
-        to: verifierNode1,
-        amount: restoreTransferAmount,
-        data: "0x",
-      })
-      .waitForReceipt(10000);
-
-    if (!restoreTransferReceipt?.transactionHash) {
-      logger.error("STEP 5: Restore transfer failed!");
-      return false;
-    }
-
-    logger.log("STEP 5: Restore transfer completed successfully!");
-
-    // Final verification
-    const finalBalanceNode1 = await cashToken.balanceOf(verifierNode1, {
-      account: verifierNode1.lookup,
-    });
-
-    const finalBalanceNode2 = await cashToken.balanceOf(verifierNode1, {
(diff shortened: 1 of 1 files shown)


.github/workflows/release-candidate.yaml
```diff
@@ -74,9 +81,18 @@ jobs:
 
   release-solidity-contracts:
     uses: ./.github/workflows/release-solidity-contracts.yaml
+  
+  release-typescript-sdk:
+    needs: [set-variables]
+    uses: ./.github/workflows/release-typescript-sdk.yaml
+    with:
+      sdk_version: ${{ needs.set-variables.outputs.tag }}
+      sdk_tag: "rc"
```
matthew1001, line 90: Should we set the tag to include the version this is a release candidate for? As a tag it's not like `latest` or `stable`, it's a release candidate of a specific version. I think including the version would make it easier to avoid accidentally pulling some old `rc`

.github/workflows/test-examples.yaml
```diff
@@ -0,0 +1,116 @@
...
+        default: false
+      run_commands:
+        description: 'Comma-separated list of commands to run (e.g., start,verify)'
+        required: false
+        type: string
+        default: 'start'
+
+jobs:
+  run-examples:
+    # here are the steps:
```
matthew1001, line 43:
```suggestion
    # here are the steps this job runs:
```

matthew1001 (approved):
Small suggestion on the way we tag release candidates.

I'd also suggest updating the PR title to say "Examples" not "Tutorials" as per the other comments on the PR

Issue #756, opened 2025-07-29
Add separate node count configuration for Paladin and Besu nodes


dwertent:
# Overview

## Problem Statement

Currently, the Paladin operator Helm chart uses a single `nodeCount` parameter that controls both Paladin and Besu node deployment simultaneously. This creates resource constraints in environments with limited resources, particularly in GitHub Actions workflows where we need to test Paladin networks but cannot afford to run multiple Besu nodes.

## Current Behavior

- `nodeCount: 3` deploys 3 Paladin nodes + 3 Besu nodes
- Each Paladin node connects to its corresponding Besu node (1:1 mapping)
- No flexibility to run different numbers of Paladin vs Besu nodes

## Desired Solution

Add a new configuration option `paladinNodeCount` that allows:
- When `paladinNodeCount` is set: Deploy Y Paladin nodes + 1 Besu node
- When `paladinNodeCount` is not set: Fall back to current behavior using `nodeCount`
- All Paladin nodes should point to the single Besu node

## Proposed Configuration

```yaml
# Current behavior (unchanged)
# Number of Paladin and Besu nodes to deploy. Relevant only when mode is set to "devnet".
nodeCount: 3  # Deploys 3 Paladin + 3 Besu nodes

# Number of Paladin nodes when using single Besu node mode
# When set, overrides nodeCount for Paladin nodes and sets Besu nodes to 1
paladinNodeCount: 3  # Deploys 3 Paladin + 1 Besu node
```

PR #758, opened 2025-07-30
756: add separate node count configuration for paladin and besu nodes

dwertent:
### Overview

This PR introduces support for a new `paladinNodeCount` configuration option in the Helm chart, allowing separate control over the number of Paladin and Besu nodes.

This is particularly useful for optimizing deployments in constrained environments (e.g., CI pipelines), where running multiple Besu nodes may be unnecessary or too resource-intensive.

This PR also adds a CI test to validate the `nodeCount` and `paladinNodeCount` configurations during installation. The test is integrated into the Gradle-based test suite to ensure both symmetric and asymmetric node setups deploy successfully.

Resolves: #756

### Updated Node Configuration

The chart now supports flexible node deployment through the following parameters:

#### Default Mode (Symmetric)

```yaml
nodeCount: 3  # Deploys 3 Paladin nodes + 3 Besu nodes (1:1 mapping)
```

#### Resource-Optimized Mode (Asymmetric)

```yaml
paladinNodeCount: 5  # Deploys 5 Paladin nodes + 1 Besu node
```

> When `paladinNodeCount` is set, the besuNodeCount automatically defaults to 1.

#### Node Connectivity

* **Single Besu Node**: All Paladin nodes connect to `besu-node1`
* **Multiple Besu Nodes**: Each Paladin node connects to a corresponding Besu node (1:1)


operator/charts/paladin-operator/values.yaml
```diff
@@ -8,6 +8,10 @@ mode: devnet
 # Number of Paladin and Besu nodes to deploy. Relevant only when mode is set to "devnet".
 nodeCount: 3
 
+# Number of Paladin nodes to deploy. When not set, defaults to nodeCount.
```
matthew1001, line 11: I wonder if this would be clearer if it was `besuNodeCount`, which defaults to `null` i.e. "same number of Besus as Paladins`, but when set to e.g. `1` then you'd get `nodeCount` Paladins and 1 Besu

operator/build.gradle
```diff
@@ -327,88 +377,160 @@ task prepareArtifacts(type: Exec, dependsOn: [buildContractSamples]) {
...
 
-        println 'Verifying all registries are published...'
-        waitForAllRegistriesPublished(namespace, 120)
+    doLast {
+        println 'Uninstalling operator...'
+        uninstallOperator(namespace)
     }
 }
 
+task ciOperatorRename(dependsOn: [installOperatorRename]) {
```
matthew1001, line 524: It feels like this is really `ciOperatorCustomName` rather than `ciOperatorRename`? Likewise for `installOperateRename`

matthew1001 (approved):
Approved with some slight structural/naming comments

operator/charts/paladin-operator/values.yaml
dwertent: [Here is how](https://github.com/LF-Decentralized-Trust-labs/paladin/pull/758/commits/23a8fbe8619f83e1453c9fd10e81a4adeda951f6#diff-3c051958b99c539b089e90f60189a7d11a1d8c307d3e77f62b7a865ce82f7b4bR5) I ended up working around the number of besu nodes

PR #760, opened 2025-07-31
Add a test for TX type not being present

matthew1001:
Adding another unit test to exercise the function added in https://github.com/LF-Decentralized-Trust-labs/paladin/pull/754, and make `txType` an omitable field.


dwertent (approved):
LGTM

PR #770, opened 2025-08-06
[examples] Making the Logs more Visual

onelapahead:
Fun use of 🤖 !


dwertent (changes requested):
I think it's a great idea and can enhance the user experience, but if we include emojis and fun prints, we should apply them consistently across all examples.

PR #772, opened 2025-08-07
Chained receipts

### LINKED ISSUE #769: Private chained transactions results should be linked to originating transaction

In the majority of cases, privacy domain transactions result directly in a "public" (visible/verifiable by all) blockchain transaction that is the finalizing construct for the transaction.

However, it is also possible for a privacy domain to delegate finalization to _another_ private transaction in a separate privacy domain.

https://github.com/LF-Decentralized-Trust-labs/paladin/blob/08e202d695185ffb07160ff9bcefb6ffb42082e2/toolkit/proto/protos/to_domain.proto#L418-L428

The primary example of this today, is the `Noto` reference domain, which provides the option for a privacy-preserving EVM smart contract in a `Pente` privacy domain to be the "Notary" for the transaction. 

It's a really powerful feature that allows decentralized notarization, and encoding of full notary logic in deterministic EVM/Solidity 🦾 

However, when things go wrong - it's very hard to diagnose today because _there is no linkage recorded in the original `Noto` transaction to the `Pente` transaction - which can fail._

In the failure case where the EVM private notary rejects the transaction (policy enforcement "on-chain" in the private EVM etc.) the `Pente` transaction is never submitted to the public blockchain. The `Pente` transaction is marked failed, but the original `Noto` transaction never resolves.

### Why is this needed?

The submitter of a transaction, must be able to establish:
1. If their original transaction was finalized or not
2. If it failed, what the error was

peterbroadhurst:
Fixes #769

Distributes a receipt back to the originator of a transaction, in the case that a transaction results in a downstream secondary chained transaction.

The approach (second rev - see below) is as follows:
- Private TX manager on the **coordinator** node now writes a `ChainedPrivateTransaction` DB record _locally_
    - This is the node where the chained transaction initiates _not_ the original transaction
    - For example a notary node for `Noto`
- The receipt for that chained transaction gets to `FinalizeTransactions` in Transaction Manager
    - This could involve a round trip to another node and back, for example the coordinator of the Pente privacy group
    - It could be a success, or a failure
- The  `FinalizeTransactions` code does a local lookup for the `ChainedPrivateTransaction` record
- If found, a duplicate of the receipt is generate with these stashed differences:
    - The original `sender` - so we know where to route it to (might be local, might be remote)
    - The original `transactionId` - the key piece
    - The original `domain` - because the chained receipt is likely to be a different domain
    - The original `contract_address` (of the domain smart contrac)
- This is passed the Private TX manager recursively in the same DB transaction, which either:
    - Determines this is also local, so it should call back into `FinalizeTransactions` again
    - Otherwise, it sends it over to the transport manager
- If the parent transaction has an in-memory sequencer in-flight
   - The queue of transactions ahead of the chain will be invalid at this point
   - This addition addition was proposed in https://github.com/kaleido-io/paladin/pull/21 after initial testing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,4 @@ bin/
 ### Proto ###
 *.pb.go
 
-contracts/
-
 **.tar.gz
--- a/doc-site/docs/architecture/noto.md
+++ b/doc-site/docs/architecture/noto.md
@@ -91,8 +91,34 @@ new UTXO states will be created, in order to facilitate the requested transfer o
 }
 ```
 
+* **to** - lookup string for the identity that will receive transferred value
+* **amount** - amount of value to transfer
+* **data** - user/application data to include with the transaction (will be accessible from an "info" state in the state receipt)
+
+### transferFrom
+
+Transfer value from a specified account to another recipient. Available UTXO states will be selected for spending, and
+new UTXO states will be created, in order to facilitate the requested transfer of value.
+
+!!! important
+    This method is only available in hooks mode. It is disabled in basic mode.
+
+```json
+{
+    "name": "transferFrom",
+    "type": "function",
+    "inputs": [
+        {"name": "from", "type": "string"},
+        {"name": "to", "type": "string"},
+        {"name": "amount", "type": "uint256"},
+        {"name": "data", "type": "bytes"}
+    ]
+}
+```
+
 Inputs:
 
+* **from** - lookup string for the identity whose tokens will be transferred
 * **to** - lookup string for the identity that will receive transferred value
 * **amount** - amount of value to transfer
 * **data** - user/application data to include with the transaction (will be accessible from an "info" state in the state receipt)
@@ -147,12 +173,31 @@ Inputs:
 * **amount** - amount of value to burn
 * **data** - user/application data to include with the transaction (will be accessible from an "info" state in the state receipt)
 
+### burnFrom
+
+Burn value from a specified account. Available UTXO states will be selected for burning, and new UTXO
+states will be created for the remaining amount (if any).
+
+!!! important
+    This method is only available in hooks mode. It is disabled in basic mode.
+
+```json
+{
+    "name": "burnFrom",
+    "type": "function",
+    "inputs": [
+        {"name": "from", "type": "string"},
+        {"name": "amount", "type": "uint256"},
+        {"name": "data", "type": "bytes"}
+    ]
+}
+```
+
 Inputs:
 
-* **inputs** - input states that will be spent
-* **outputs** - output states that will be created
-* **data** - encoded Paladin and/or user data
-* **delegate** - address of the delegate party that will be able to execute this transaction once approved
+* **from** - lookup string for the identity whose tokens will be burned
+* **amount** - amount of value to burn
+* **data** - user/application data to include with the transaction (will be accessible from an "info" state in the state receipt)
 
 ### lock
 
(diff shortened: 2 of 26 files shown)


core/go/internal/txmgr/persisted_receipt.go
```diff
@@ -151,14 +151,35 @@ func (tm *txManager) FinalizeTransactions(ctx context.Context, dbTX persistence.
 	}
 
 	if len(receiptsToInsert) > 0 {
+		// Do a query for chained transactions (note only one level here - double chaining not supported)
```
awrichar, line 154: Seems like an important caveat (although not a practical problem at this time).

So what we're doing here is copying the receipt from the chained transaction back to its parent. But if you ever had two-deep, the grandparent transaction would never resolve. Is that right?

I guess the alternative to this would be that the parent receipt goes into some special "chained" state as soon as the child transaction is spawned..? And the client/user would have to manually perform another fetch?

I don't disagree with the approach, but trying to weigh pros/cons and whether there are any alternatives.

peterbroadhurst:
I tested the initial approach to this in the PR and found:
```
2025/08/07 18:49:57 [31;1m/app/core/go/internal/txmgr/transaction_submission.go:421 [35;1mERROR: insert or update on table "chained_private_txns" violates foreign key constraint "chained_private_txns_transaction_fkey" (SQLSTATE 23503)
[0m[33m[7.876ms] [34;1m[rows:0][0m INSERT INTO "chained_private_txns" ("transaction","chained_transaction") VALUES ('fc82fa5e-6eb3-4ef8-857f-debb36e08181','f25019c4-38cb-4aa3-ad01-c4c8d49dcb60') ON CONFLICT DO NOTHING
```

This is because the chained transaction creation is happening on the coordinator of the transaction (the notary for example), not the submitter of the transaction. So we can't attach the chained transaction directly to the original transaction, because they are being submitted by different parties.

However, we can record the chaining - so we can still detect it when we get the receipt for the chained transaction, and detect the original transaction. But at this point, we need to involve the private TX manager (rather than just writing an extra receipt directly) - because the receipt needs to be pushed back to the node that originated the upstream transaction at that point.

core/go/internal/txmgr/persisted_receipt.go
peterbroadhurst: This has been removed with the 2nd rev approach.

core/go/db/migrations/postgres/000020_create_private_tx_binding_tables.up.sql
```diff
@@ -0,0 +1,11 @@
+BEGIN;
+
+CREATE TABLE chained_private_txns (
+  "chained_transaction"       UUID            NOT NULL REFERENCES transactions ("id") ON DELETE CASCADE,
+  "transaction"               UUID            NOT NULL, -- we might not be the originator of the original transaction, so might not have it in our DB
+  "sender"                    TEXT            NOT NULL, -- sender of the original transaction, so we know where to route the receipt
+  "domain"                    TEXT            NOT NULL,
+  PRIMARY KEY ("chained_transaction", "transaction")
```
peterbroadhurst, line 8: Note the lookup direction is only by `chained_receipts` in the revised proposal

examples/bond/package-lock.json
peterbroadhurst, line 1: hmmn, I didn't mean to generate any changes in the examples

core/go/internal/txmgr/persisted_receipt.go
```diff
@@ -171,6 +176,39 @@ func (tm *txManager) FinalizeTransactions(ctx context.Context, dbTX persistence.
 		}
 	}
 
+	// Do a query for chained transactions (note only one level here - double chaining not supported)
```
awrichar, line 179: I think the caveat here can be removed now

PR #776, opened 2025-08-07
pente: treat IllegalArgumentException as a revert condition

awrichar:
Also be very specific about the remaining exceptions that can cause retry (should only be those expected to be transient).

Avoid throwing Exception (always throw a more specific exception).


peterbroadhurst (approved):
I agree with this approach 👍 

e.g. that it's safe to assume all cases of `IllegalArgumentException` are non-retryable cases.

PR #779, opened 2025-08-08
Add transferFrom and burnFrom to Noto

awrichar:
Both are disabled in "basic" mode (only burn and transfer are allowed), but these may be allowed in "hooks" mode.


peterbroadhurst (approved):
Looks like great addition, and pleased to see the common code.

PR #781, opened 2025-08-09
ui: add "check balance" for Noto/Zeto

awrichar:


gabriel-indik:
Branch got out of sync, and even if conflicts are resolved, it will soon go out of sync again because of ongoing work on the domains panel. For that reason, the changes in here have been incorporated into PR https://github.com/LFDT-Paladin/paladin/pull/1192

PR #784, opened 2025-08-11
v must be 0/1

annamcallister:
https://github.com/LF-Decentralized-Trust-labs/paladin/issues/766

This PR modifies the Paladin signer to return protocol agnostic signatures with V as 0/1 (previously 27/28).

This update means that a number of other places across the code base have needed changes to map back to legacy 27/28 values when relevant to the protocol they are using the signature with:

* Pente domain: requires EIP-712 signatures to have legacy V values. This mapping is done after all the endorsements have been collected and the values passed back to the domain in the node that will build the public transaction. Much of the endorsement gathering code is in the Paladin domain manager and not specific to the domain, so this was the only place to make the change where it wouldn't affect other future domains which may make a different choice on how to handle V.
* Domain manager: when building EIP-155 signatures calls `FinalizeLegacyEIP155WithSignature` from `firefly-signer` which expects legacy V values as input. This code path is not currently used by any domain though.
* ethclient: when handling EIP-155 or legacy transaction types needs to map back to legacy V values


peterbroadhurst (approved):
This is great @annamcallister - like that we have:
1. consistency
2. good comments on why
3. tests that mean if the deps change (`firefly-signer` etc.) in this regard we'll detect it

PR #785, opened 2025-08-11
Implement Automatic SDK Revert to Prevent Git Changes in Examples

dwertent:
### Overview
Prevents unwanted git changes to `package.json` files after building examples by automatically reverting local SDK dependencies to published versions.

### DIFF
--- a/examples/bond/build.gradle
+++ b/examples/bond/build.gradle
@@ -35,7 +35,7 @@ dependencies {
     buildCommon project(path: ':examples:common', configuration: 'buildCommon')
 }
 
-task install_sdk(type: Exec) {
+task npm_install_sdk(type: Exec) {
     executable 'npm'
     args 'install' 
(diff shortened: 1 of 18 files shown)


sdk/typescript/build.gradle
```diff
@@ -64,6 +68,16 @@ task build(type: Exec, dependsOn: [install, copyABI]) {
     outputs.dir("build")
 }
 
+
+task link(type: Exec, dependsOn: [build]) {
+    executable 'npm'
+    args 'link'
```
awrichar, line 74: I still don't love using global commands (this will install a global symlink on my system for the package that is built from source here). Especially if I don't know what's happening, it can lead to weird issues. `npm link` can be a useful thing for a dev, but I'm worried about doing it transparently on their system.

Can't we find a way to contain the different build patterns within the build folder, without any global changes or in-source changes?

PR #798, opened 2025-08-14
noto: fix sender address in transferFrom hook

awrichar:


dwertent (approved):
LGTM. 
Did we miss this in the unittests?

PR #796, opened 2025-08-14
Handle nil contractAddr on transaction indexing

peterbroadhurst:
Issue after application of #795

```
[2025-08-14T00:02:25.961Z] ERROR Panic within database transaction: runtime error: invalid memory address or nil pointer dereference
goroutine 308 [running]:
runtime/debug.Stack()
    /usr/local/go/src/runtime/debug/stack.go:26 +0x5e
github.com/LF-Decentralized-Trust-labs/paladin/core/pkg/persistence.(*provider).Transaction.func1()
    /app/core/go/pkg/persistence/gorm.go:175 +0x85
panic({0x7fdc2ec45dc0?, 0x7fdc2f712620?})
    /usr/local/go/src/runtime/panic.go:792 +0x132
github.com/LF-Decentralized-Trust-labs/paladin/core/internal/privatetxnmgr.(*privateTxManager).NotifyFailedPublicTx(0xc0006cc3c0, {0x7fdc2ee70048, 0xc000a954d0}, {0x7fdc2ee784a0, 0xc000994500}, {0xc00070a378, 0x1, 0x7fdc6c1cbf30?})
    /app/core/go/internal/privatetxnmgr/private_txn_mgr.go:976 +0xb9
github.com/LF-Decentralized-Trust-labs/paladin/core/internal/txmgr.(*txManager).blockIndexerPreCommit(0xc0001c6000, {0x7fdc2ee70048, 0xc000a954d0}, {0x7fdc2ee784a0, 0xc000994500}, {0xc0006cc3c0?, 0x7fdc2ebb9a40?, 0x7fdc2ee5bdc0?}, {0xc000a89a80, 0x9, ...})
    /app/core/go/internal/txmgr/block_indexing.go:76 +0x70a
github.com/LF-Decentralized-Trust-labs/paladin/core/pkg/blockindexer.(*blockIndexer).writeBatch.func1.1({0x7fdc2ee70048, 0xc000a954d0}, {0x7fdc2ee784a0, 0xc000994500})
    /app/core/go/pkg/blockindexer/block_indexer.go:617 +0x18d
github.com/LF-Decentralized-Trust-labs/paladin/core/pkg/persistence.(*provider).Transaction.func2(0xc00078f830?)
    /app/core/go/pkg/persistence/gorm.go:197 +0xa2
gorm.io/gorm.(*DB).Transaction(0xc00078f830, 0xc0007a3800, {0x0?, 0x0?, 0x0?})
[...]

PR #802, opened 2025-08-14
Fix double lock

peterbroadhurst:
Two more tweaks needed on top of #796, based on iterative testing with #799 recreate scenario (which I've not added into the build, as it's in the sample).

PR comments in-line in code below with the changes

### DIFF
--- a/core/go/internal/publictxmgr/transaction_manager.go
+++ b/core/go/internal/publictxmgr/transaction_manager.go
@@ -774,6 +774,7 @@ func (ptm *pubTxManager) MatchUpdateConfirmedTransactions(ctx context.Context, d
 		for _, match := range lookups {
 			if txi.Hash.Equals(&match.Submission.TransactionHash) {
 				// matched results in the order of the inputs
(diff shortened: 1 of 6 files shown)


core/go/internal/publictxmgr/transaction_manager.go
```diff
@@ -757,7 +757,7 @@ func (ptm *pubTxManager) MatchUpdateConfirmedTransactions(ctx context.Context, d
 	var lookups []*bindingsMatchingSubmission
 	err := dbTX.DB().
 		Table("public_txn_bindings").
-		Select(`"transaction"`, `"tx_type"`, `"Submission"."pub_txn_id"`, `"Submission"."tx_hash"`).
+		Select(`"transaction"`, "sender", "contract_address", `"tx_type"`, `"Submission"."pub_txn_id"`, `"Submission"."tx_hash"`).
```
peterbroadhurst, line 760: This was critical - meant `Sender` wasn't set (for cross-node propagation), and `ContractAddress` for forcing a reset of the in-memory state of the privacy domain when transactions fail after dispatch.

hosie (approved):
LGTM

PR #805, opened 2025-08-19
Always retrieve account balance when waiting for balance

annamcallister:
https://github.com/LF-Decentralized-Trust-labs/paladin/issues/790

If the result of processing transactions is that we are waiting for the account balance to be updated we should always be getting a new refreshed value from the chain for the next iteration.


matthew1001:
I've been testing with this change locally and it's been working as expected

matthew1001 (approved):
LGTM, and tested successfully locally

PR #807, opened 2025-08-20
L1/L2 chain initial setup docs

matthew1001:
We're actively working on improvements to make public chains easier to deploy against, but having some of the config considerations in the docs gives us a place to start pointing users who want to test basic deployments.

### DIFF
--- a/doc-site/docs/getting-started/l1-and-l2-chains.md
+++ b/doc-site/docs/getting-started/l1-and-l2-chains.md
@@ -1,0 +1,75 @@
+# Running Paladin with a public L1 or L2 chain as the base ledger
+
+## Introduction
+
+Paladin supports any EVM compatible chain as the base ledger for public transactions. This includes using public chains that use paid gas, such as Ethereum Mainnet L1 chain or the Linea Mainnet L2 chain. There
+are certain considerations needed to run Paladin using a paid gas base ledger.
+
+## Indexing
+
+A public chain will have a long chain history. Paladin indexes the base ledger in order to confirm new transactions have been included in blocks. By default Paladin indexes the base ledger from block `0`. For a public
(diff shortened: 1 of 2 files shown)


doc-site/docs/getting-started/l1-and-l2-chains.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Indexing
+
+A public chain will have a long chain history. Paladin indexes the base ledger in order to confirm new transactions have been included in blocks. By default Paladin indexes the base ledger from block `0`. For a public
+chain this could take many days to index. Paladin only requires block history from the point at which Paladin tranasctions and contracts were deployed. To reduce the time to index the chain, use the `fromBlock` configuration
+option under the `blockIndexer` configuration section. For example:
+
+```
+blockIndexer:
+    fromBlock: 9024200
```
annamcallister, line 16: `latest` is also a valid option in this field which means no catchup time.

There is a consideration here that if nodes are added in later to the network, they must index at least as far back as the initial deployment of the registry contract, otherwise they will not be able to index events which give them the details of other nodes in the network.

If using a specific block number then guidance for this can be expressed quite simply as use the same `fromBlock` value for all nodes. If using `latest` it's a bit more complex as later nodes cannot use this value- they must use a specific block number.

annamcallister (approved):

doc-site/docs/getting-started/l1-and-l2-chains.md
matthew1001: Yeah agreed - I've tried to capture the issue of needing to index from where Paladin contracts were first deployed, and I think setting a number makes it a little clearer that you need to think about the value for those reasons

PR #814, opened 2025-08-26
RC flow fixes

dwertent:


.github/workflows/test-v9-to-v10.yaml
```diff
@@ -140,6 +140,9 @@ jobs:
           PALADIN_SDK_VERSION: ${{ needs.validate-tags.outputs.previous_tag }}
           RUN_COMMANDS: start,verify
         run: |
+          # checkout the to tag v0.10.0-rc.7 (before the main branch was updated)
+          git checkout v0.10.0-rc.7
```
dwertent, line 144: We can’t run the `main` branch against the released SDKs because of changes introduced there. However, at tag `v0.10.0-rc.7`, the examples were still compatible with the released SDKs. That’s why we check out that version; this ensures that whatever users ran with v9 will continue to work with v10.

dwertent:
https://github.com/LF-Decentralized-Trust-labs/paladin/actions/runs/17269053749

PR #821, opened 2025-09-03
Submit domain transactions from configured identity

annamcallister:
Provide the option for a domain to be configured with a fixed identity. When this is set, all public transactions that result from domain actions that would normally use a randomly generated identity to sign a public transaction will use this fixed identity instead.

It can be configured this via the helm charts by adding the following section to the values file:
```yaml
domains:
  noto:
    fixedSigningIdentity: noto.default # or any identity you choose
  pente:
    fixedSigningIdentity: ...
  zeto:
   fixedSigningIdentity: ...
 ```
Noting that:
* The fixed signing identity will be resolve to a different signing address on each Paladin node (e.g. `noto.fixed@node1` will resolve to a different signing address than `noto.fixed@node2`)
* To know which addresses to fund ahead of using the domains you should resolve the fixed identities against all nodes e.g.
```bash
curl --location '127.0.0.1:31548' --header 'Content-Type: application/json' --data-raw '{
      "jsonrpc": "2.0",
      "id": "1",
      "method": "ptx_resolveVerifier",
      "params": ["noto.fixed@node1", "ecdsa:secp256k1", "eth_address"]
    }'
```
* You may choose to use the same fixed identity for all three domains to minimise how many addresses you need to fund
* Some public transactions are signed by an identity specified in the from field of the private transaction that created them (e.g. a Noto mint or transfer) - the behaviour of these cannot and has not changed. Resolving these identities and funding the resolved address before using them gives the cleanest experience here.

### DIFF
--- a/core/go/internal/domainmgr/domain.go
+++ b/core/go/internal/domainmgr/domain.go
@@ -52,12 +52,13 @@ type domain struct {
 	ctx       context.Context
 	cancelCtx context.CancelFunc
 
-	conf            *pldconf.DomainConfig
-	defaultGasLimit pldtypes.HexUint64
-	dm              *domainManager
-	name            string
-	api             components.DomainManagerToDomain
-	registryAddress *pldtypes.EthAddress
+	conf                   *pldconf.DomainConfig
+	defaultGasLimit        pldtypes.HexUint64
+	dm                     *domainManager
+	name                   string
+	api                    components.DomainManagerToDomain
+	registryAddress        *pldtypes.EthAddress
+	defaultSigningIdentity string
 
 	stateLock          sync.Mutex
 	initialized        atomic.Bool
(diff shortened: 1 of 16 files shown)


core/go/internal/domainmgr/domain.go
```diff
@@ -52,12 +52,13 @@ type domain struct {
...
-	name            string
-	api             components.DomainManagerToDomain
-	registryAddress *pldtypes.EthAddress
+	conf                   *pldconf.DomainConfig
+	defaultGasLimit        pldtypes.HexUint64
+	dm                     *domainManager
+	name                   string
+	api                    components.DomainManagerToDomain
+	registryAddress        *pldtypes.EthAddress
+	defaultSigningIdentity string
```
peterbroadhurst, line 61: I think like `fixedSigningIdentity` would be a better spelling if that's ok

PR #823, opened 2025-09-04
Fix evm-registry label

annamcallister:
The label was inconsistent which resulted in the deployed nodes having invalid registries


matthew1001 (approved):
LGTM

PR #829, opened 2025-09-10
New guide docs section with certificate and peering doc updates

annamcallister:
* Adds a new "guides" docs section
* Moves the L1/L2 page into that section
* Adds a new page on certificate management
* Adds a new pages on peering debug

I've left a few comments inline in the docs about where we could make improvement- these comments won't show in the rendered docs

I've made a couple of small operator changes
* add block number to CR transaction status - I was having to write lengthy instructions to work around not having this
* mount CA if available- custom CAs just aren't going to work without this. This wasn't possible until https://github.com/LF-Decentralized-Trust-labs/paladin/pull/828 was delivered

But otherwise these docs are really a reflection of where we are and I think help highlight areas for future enhancements.


core/go/internal/publictxmgr/in_flight_transaction_stage_controller_test_retrieve_gas_test.go
```diff
@@ -130,13 +130,11 @@ func TestProduceLatestInFlightStageContextRetrieveGas(t *testing.T) {
 	it.stateManager.GetCurrentGeneration(ctx).AddPersistenceOutput(ctx, InFlightTxStageRetrieveGasPrice, time.Now(), nil)
 	rsc.StageOutput.GasPriceOutput.Err = nil
 	rsc.StageErrored = false
-	// it.stateManager.GetCurrentGeneration(ctx).SetValidatedTransactionHashMatchState(ctx, true) // TODO AM???
```
annamcallister, line 133: found this while searching for TODOs- should have been removed in an earlier PR

PR #834, opened 2025-09-15
Add logging option to include `component` in log entries

matthew1001:
The existing `role` log field used by most components helps to identify the routine issuing the log call (`block_indexer`, `public_tx_manager` etc) but doesn't help identify the component/package the log entry is defined within.

It is occasionally useful to include the component where the log entry is defined, and to filter logs for component-specific entries regardless of the role issuing the call.

This PR adds `WithComponent(ctx, component)` to the common logging package.

Example use:

```
log.L(log.WithComponent(ctx, "registry_manager")).Debugf("<log message>")
```

which results in a log such as:

```
[2025-09-15T11:59:35.545Z] DEBUG <log message> role=transports_manager component=registry_manager
```


annamcallister (approved):
Approving the main change while noting that coverage needs to increase before this can be merged

matthew1001:
I've added the necessary test to restore 100% coverage, just waiting for CI to run

PR #837, opened 2025-09-15
add RevertReason to error message in testbed

jimthematrix:
For the testbed only, the error message is enhanced
- from: `assemble result was REVERT` 
- to: `assemble result was REVERT (reason: io.kaleido.paladin.toolkit.PluginInstance$ErrorResponseException: PD011634: ABI encoding request failed: FF22062: Negative numeric value is invalid for component`


matthew1001 (approved):
LGTM

PR #840, opened 2025-09-23
Example polling timeouts

annamcallister:
Define a default and long poll timeout plus a poll interval for the examples and use these constants across the tests.

The default I have bumped from 5/10 secs to 30 secs to resolve https://github.com/LF-Decentralized-Trust-labs/paladin/issues/806.

The long I have set at 120 secs, which is largely what was in use across the tests.

Where other time periods were used I have generally bumped them up, for the simplicity of only having 2 types of timeout.

I have put a couple of erc20 public transaction deploys/mints back to the default from 120 secs. I think these were raised for debug but if these genuinely need 120 seconds then we need to be investigating what is going wrong, not just increasing the timeout.

I've cleaned up a couple of unused imports too, and reordered the wait in a loop so that the error message is more accurate.

### DIFF
--- a/core/go/build.gradle
+++ b/core/go/build.gradle
@@ -348,7 +348,7 @@ task unitTestPostgres(type: UnitTests, dependsOn: [
     mustRunAfter 'unitTestSQLite' // these tests cannot run concurrently
 }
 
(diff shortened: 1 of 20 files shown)


matthew1001 (approved):
LGTM

PR #841, opened 2025-09-23
Updates to pre-release workflow, including removal of temporary `0.9`->`0.10` upgrade test behaviour

dwertent:
# Overview

This PR introduces two changes:

~1. **Override Paladin SDK before creating a tag (release)**
   Ensures that when a user checks out a tag, the SDK version aligns with that tag.~

2. **Use local SDK on `main`**
   Prevents unnecessary `git` events from being triggered during local builds (instead of #785 ).

3. **Fixing RC testing**


.github/actions/override-sdk-version/action.yaml
```diff
@@ -0,0 +1,30 @@
+name: Override SDK tag
```
matthew1001, line 1: Is this action used in this PR or was does it need to be shifted over to https://github.com/LF-Decentralized-Trust-labs/paladin/pull/846 along with the other SDK override changes?

.github/workflows/release.yaml
```diff
@@ -5,6 +5,10 @@
 # This workflow (phase 2) creates the final release after RC testing has been completed.
 name: Release
 
+permissions:
```
matthew1001, line 8: Are some of these only required for the override SDK PR https://github.com/LF-Decentralized-Trust-labs/paladin/pull/846? If so it would be good to move them there as part of that review rather than leave them here.

.github/workflows/test-rollout-compatibility.yaml
```diff
@@ -156,26 +149,21 @@ jobs:
           
           # We check out the new tag here primarily to validate the new chart installation
           # with the corresponding version of the validation script.
-          git checkout main
+          git checkout ${{ github.sha }}
```
matthew1001, line 152: Is this also remnants from the override SDK changes we're deferring?

examples/bond/package.json
```diff
@@ -20,7 +20,7 @@
     "typescript": "^5.6.3"
   },
   "dependencies": {
-    "@lfdecentralizedtrust-labs/paladin-sdk": "^0.9.0",
+    "@lfdecentralizedtrust-labs/paladin-sdk": "file:../../sdk/typescript",
```
matthew1001, line 23: I can't remember if we said we'd leave `v0.9.0` for this release and maybe change to something else later?

examples/common/src/config.ts
```diff
@@ -79,7 +80,10 @@ function loadConfigFromFile(configPath?: string): NodeConnection[] {
   
   if (configPath) {
     // Use the provided config path
-    filePath = path.resolve(configPath);
+    // filePath = path.resolve(configPath);
```
matthew1001, line 83: Might be some tidy up to do here

matthew1001 (commented):
I think there are a few changes that need discussing before we go ahead and merge, particularly github action permissions but also anything left over from the sdk-override changes now targeted for PR https://github.com/LF-Decentralized-Trust-labs/paladin/pull/846

examples/bond/package.json
dwertent: We agreed to leave this change as-is to avoid creating too much disruption in the PR. It has no impact on end users, since they are required to override the SDK version to the latest.

matthew1001 (approved):
LGTM

PR #844, opened 2025-09-25
Revert "run component tests against a chain which does not have free gas"

dwertent:
Reverts LF-Decentralized-Trust-labs/paladin#832


annamcallister:
Closing as we've ruled this PR out from being the cause of the latest build break

PR #849, opened 2025-09-29
Use fixed address for initialising privacy group, pin to postgres 17.6

annamcallister:
The fixed address for initialisting a privacy group was missed when the domain fixedSigningAddress feature was added.

Pinning to a specific postgres version for stability, but also because we know that 17.6 to latest requires migration action so we will need to plan for this.


EnriqueL8 (approved):
Looks good! Chatted with Anna on these missing changes

PR #851, opened 2025-10-01
Provide a mechanism for a transport to return non-recoverable errors to the caller

matthew1001:
Fixes https://github.com/LF-Decentralized-Trust-labs/paladin/issues/850


matthew1001:
Thanks for the comments @annamcallister , I think I've addressed them in the latest commits

annamcallister (approved):

PR #875, opened 2025-10-23
Add config doc generation

annamcallister:
Adds a reflection based approach to building configuration documentation including defaults. This has been entirely developed with AI agents, and has been commented in such a way to facilitate any future development being done by agents. As such the comment/code style is not typically what I’d deliver when expecting humans to be the main consumers.

The reflection based approach is different to what other projects have done to achieve a similar end (e.g. `firefly-common`), but this is the approach that required the smallest number of changes to the existing code. It is arguably a retrofit rather than how we might have designed this from scratch; however, it is also an approach which is fairly straight forward to add new configuration values into.

For this reflection based approach to work I have had to define a full default `PaladinConfig` instance which has meant adding in some defaults which weren’t previously defined. The main example of this is that the `retry` package always falls back to its own defaults if a user hasn’t provided them in config. This meant the relevant default config definitions which included a retry section didn’t need to define defaults for the code to work. But they do now if we want the defaults to appear in the documentation. 

It has also meant adding in some intermediary structs providing the link between the root of the `PaladinConfig` and the multiple levels deep nested structs where the defaults are actually defined.  Where `core` code references these defaults, I have moved it to referencing the higher level intermediate structs where possible to avoid exposing too many different variables from the `pldconf` package, with the risk that the connection between defaults used in code and defaults used for documentation gets severed.

I have also done some general tidy up around how we declare configuration:
* Avoid `ManagerManager` in names
* Avoid initialising structs with a pointer, only to immediately dereference them
* Remove variables claiming to hold defaults that are actually unused
* Move defaults that were defined in the core package into the config package ﻿

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -1,4 +1,4 @@
-## Adding New Configuration Properties
+## Adding new configuration properties
 
 ### 1. Define the property
 
@@ -14,7 +14,7 @@ type DBConfig struct {
 
 `PaladinConfig` in [`config.go`](./pkg/pldconf/config.go) is the root of the Paladin configuration struct.
 
-### 2. Add Default Values if needed
+### 2. Add default values if needed
 
 Define default values inside the `PaladinConfigDefaults` variable:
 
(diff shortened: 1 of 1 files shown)


matthew1001 (approved):
There's obviously quite a lot of churn in the PR. I think if it documents accurately our current config and doesn't break existing behaviour and/or defaults then it seems good to me.

Issue #884, opened 2025-10-30
New locations for publishing docker images and npm package.


annamcallister:
### What would you like to be added?

Docker images are currently pushed to `lfdecentralizedtrustlabs/paladin(-operator)` and `ghcr.io/lf-decentralized-trust-labs/paladin(-operator)`.

SDK npm package is published to https://www.npmjs.com/package/@lfdecentralizedtrust-labs/paladin-sdk

We need to choose a new scope for these images and package.

### Why is this needed?

Paladin is no longer a lab project.

ryjones:
[Do you want to publish here?](https://www.npmjs.com/org/lfdecentralizedtrust) do you want a new org (lfdt-paladin)?

dwertent:
I’d wait for the rest of the team to weigh in on this.
What about the image registries (Docker and GHR)?

PR #887, opened 2025-11-04
Package rename after org move

annamcallister:
Ran go mod tidy a fair few times to do this  and seem to have updated some dependencies so some additional changes have resulted from that


annamcallister:
`.github/settings.yml` was a miss but the others are related to #884 and not in scope of this PR

PR #888, opened 2025-11-05
Create MAINTAINERS.md

matthew1001:
This is a required file for LFDT projects. I have used the LFDT sample maintainers file and added an initial set of maintainers.

Currently this is **not** the full list currently in the https://github.com/LFDT-Paladin/governance/blob/main/config.yaml config file. The intention is to provoke some discussion about who is a maintainer now that the two labs have become an incubating project.

I have proposed that we have a single list of maintainers for both Paladin and Zeto, but that the maintainers file identifies which of the 2 repos (or both) the user is a committer for. 

There is a PR for a minimal Zeto redirect maintainers file here: https://github.com/LFDT-Paladin/zeto/pull/169


dwertent (approved):
LGTM

PR #895, opened 2025-11-10
Paladin Technical Charter

matthew1001:
Every LFDT project requires a charter which outlines the scope and operating model of the project.

A short summary of the "default" charter that has been proposed by the LF is:
 - The project has a steering committee (TSC)
   - By default this is the active maintainers (see https://github.com/LFDT-Paladin/paladin/pull/888), but it could diverge from that list if the project had a reason to do so
 - Unless we specify something different, the project will follow the [default LFDT TAC code of conduct](https://lf-decentralized-trust.github.io/governance/governing-documents/code-of-conduct/)
 - Becoming a maintainer involves the approval of the majority of existing maintainers
   - Although this is a simplification of the actual rules outlined in https://github.com/LFDT-Paladin/paladin/pull/888)
 - Changes to the charter will require 2/3 of the TSC to agree
 - Contributors & maintainers will act in a professional way for the good of the community, respect others' trademarks, etc.
   - Read the PR in detail to see specifics

I have made 1 alteration to the default LF charter, which is simply to outline that the specific rules for becoming a maintainer are outlined in `MAINTAINERS.md`, and that the statement in the charter "involves the approval of the majority of existing maintainers" is a simplification of the actual process.


matthew1001:
On recommendation from LF legal I've reverted the fall-back code-of-conduct mentioned by the charter so it still points at the LF Projects CoC, not the LFDT one. Since we have a specific CoC in `.github` now it's not applicable anyway.

PR #900, opened 2025-11-11
New npm package scope for examples

annamcallister:
Follows on from #896 now that the v0.13.0 release has published the sdk to the new npm scope.

I've taken this opportunity after spotting that a couple of the examples installed with high vulnerabilities to do a dependency update across the board in the examples.


matthew1001 (approved):
Just to unblock the release, and as it looks like Anna's updated it to reflect @dwertent 's comments, I'm adding an approval

PR #912, opened 2025-11-20
Revert 872 privacy group access

annamcallister:
Reverts https://github.com/LFDT-Paladin/paladin/pull/872 which changed intended behaviour.


matthew1001 (approved):
Looks like it cleanly reverts the changes in https://github.com/LFDT-Paladin/paladin/pull/872. Subsequent discussions possibly to have with other maintainers on how sub-node tenancy might be introduced to Paladin, but this behaviour was intentional while we don't have proper multi-tenancy

Issue #701, opened 2025-06-15
Tokens should support minting/burning locked tokens


awrichar:
### What would you like to be added?

Both Noto and Zeto currently support locking tokens (via the "lock" method) and then releasing them to the same or a new recipient (via a method that Noto calls "unlock" and Zeto calls "transferLocked").

I propose that we also need additional methods:
- mintLocked - Create new states that are locked. Same permission characteristics as mint. Minted states can be unlocked/transferred via current behaviors, or can be burned (see below) to "undo" the mint. Useful for on-ramp and other conditions where a mint operation needs to be verified or coordinated with other calls before becoming valid.
- burnLocked - Burn locked states. Same permission characteristics as burn. Useful for off-ramp and other conditions where a burn operation needs to be verified or coordinated with other calls before becoming valid. 

Noto's "prepare unlock" function will also need some consideration for how you prepare a burn as opposed to a transfer.

### Why is this needed?

For instruments like tokenized deposits, it's common to manage supply with mint and burn operations that must be coordinated across multiple tokens/contracts. Adding these operations to the existing lock concept will make them fully programmable.

awrichar:
I started exploring this in Noto as part of https://github.com/LFDT-Paladin/paladin/pull/813

awrichar:
After thinking further, I'm not positive if my description of `mintLocked` is correct. What I think you might want is a way to say "if this unlock operation is executed, new tokens should come into existence". What I've described is actually that locked tokens come into existence first, and then get moved as part of the unlock...

So we may not need `mintLocked` at all, but just a way to do a version of "prepare unlock", which prepares a lock ID and some potential output states (but no input states at all).

annamcallister:
Closing as this has been added for Noto. A new issue should be raised for adding it as an enhancement to Zeto if needed

PR #918, opened 2025-11-21
noto: allow minting and burning tokens via locks

awrichar:
See #701 

Currently Noto supports a "lock" pattern in which locked tokens can be transferred to one or more recipients upon unlock. This PR extends that behavior with the ability to trigger mint/burn operations on unlock.

Changes to the private (user-facing) Noto ABI:
- Existing `lock` method is aliased to `createLock` (with the latter being preferred going forward)
- Existing `prepareUnlock` is unchanged, and allows preparing an existing lock with a set of recipients that will receive the value upon unlock
- New `createMintLock` creates a lock that is already "prepared" with a set of states that will only come into existence upon unlock
- New `prepareBurnUnlock` is similar to `prepareUnlock`, but prepares an existing lock with a burn operation (no recipients) that will be executed upon unlock

Changes to the public Noto ABI:
- Storage variables are rearranged to identify locks primarily by lock ID instead of by locked input state ID (crucial because "mint locks" have no input states at all, but this also simplifies the logic somewhat)
- Lock ID is now generated on-chain instead of off-chain
- Some method/event signatures around locking are changed to accommodate the new behaviors and address minor gaps (mainly passing txId and lockId more consistently)

---

This represents the next major variant of the Noto contract, so the "variant" parameter of Noto has been updated from `0` to `1`. The Noto domain logic in Paladin will read the variant that is reported when Noto is constructed, and will support either version. All attempts have been made to ensure both variants work, but this represents a "first" for Noto and should be carefully considered.

The contract is _not_ arranged to allow seamless migration from a V0 Noto contract to V1, due to the breaking changes in the storage layout noted above. Direct migration would be a significant challenge - but if needed, it could be added via a special upgrade path to be built at a later time, since the contracts are UUPS upgradeable.

---

Because the private (user-facing) Noto ABI had only minor changes (mostly additions), existing code written against Paladin should not need updating in most cases. The only backwards-incompatible change is in the `delegateLock` method, which has dropped a parameter to become simpler. When running `delegateLock` against a V0 Noto contract, the `unlock` parameter must be included, whereas when running against a V1 Noto contract, it must be omitted.

---

Final note: there are now _many_ aspects of Noto that are versioned independently, including:
- the [Noto public ABI](https://github.com/LFDT-Paladin/paladin/blob/1c93e9b589dd07316ce8064ec7a8179cbbf9db81/solidity/contracts/domains/noto/Noto.sol#L44) and [Noto private ABI](https://github.com/LFDT-Paladin/paladin/blob/6426bdfa76cfc9b00dd7eb6a5509e58ea969f8e9/solidity/contracts/domains/interfaces/INotoPrivate_V0.sol#L5-L6) (these are linked together)
- the [Noto factory ABI](https://github.com/LFDT-Paladin/paladin/blob/6426bdfa76cfc9b00dd7eb6a5509e58ea969f8e9/solidity/contracts/domains/noto/NotoFactory.sol#L10)
- the formatting of [encoded domain config data](https://github.com/LFDT-Paladin/paladin/blob/6426bdfa76cfc9b00dd7eb6a5509e58ea969f8e9/domains/noto/pkg/types/config.go#L29) in the `data` param of the factory `deploy`
- the formatting of [encoded transaction data](https://github.com/LFDT-Paladin/paladin/blob/6426bdfa76cfc9b00dd7eb6a5509e58ea969f8e9/domains/noto/pkg/types/config.go#L56) in the `data` param of all Noto methods
- the state schemas for [private transaction data](https://github.com/LFDT-Paladin/paladin/blob/6426bdfa76cfc9b00dd7eb6a5509e58ea969f8e9/domains/noto/pkg/types/states.go#L147) and [lock info](https://github.com/LFDT-Paladin/paladin/blob/eb5e2a7a737974dd2b293c0e9c35f445f269793c/domains/noto/pkg/types/states.go#L138) "info" states

I believe they are all handled appropriately, but just noting for the record that it can be a bit confusing to keep straight because V0 and V1 of all these things are not guaranteed to be synchronized in any way.

### DIFF
--- a/examples/bond/src/index.ts
+++ b/examples/bond/src/index.ts
@@ -378,7 +378,6 @@ async function main(): Promise<boolean> {
     .using(paladin3)
     .delegateLock(investor, {
       lockId: cashLockId,
-      unlock: cashUnlockParams,
       delegate: atomAddress,
       data: "0x",
     })
@@ -391,7 +390,6 @@ async function main(): Promise<boolean> {
     .using(paladin2)
     .delegateLock(bondCustodian, {
       lockId: bondLockId,
-      unlock: assetUnlockParams,
       delegate: atomAddress,
       data: "0x",
     })
--- a/examples/swap/src/index.ts
+++ b/examples/swap/src/index.ts
@@ -297,7 +297,6 @@ async function main(): Promise<boolean> {
     .using(paladin2)
     .delegateLock(investor1, {
       lockId,
-      unlock: assetUnlockParams,
       delegate: atom.address,
       data: "0x",
     })


solidity/contracts/domains/noto/Noto.sol
peterbroadhurst: Spent quite a bit of time reading and parsing this.

The updated spelling seems like a nice step forwards. I wasn't able to identify any holes in the logic.

I wonder if we could have a bit of discussion in the future about the why of needing to supply in the exact details of the inputs/outputs on the lock is necessary. Given they must be locked.

One potential edge case I noticed, was if you do a `lock`, then you do an `unlock`, the on-chain doesn't force the two to match (assuming you didn't do a prepare). This makes sense from an on-chain efficiency of storage perspective, as we have to store the locked state of the states once in the map, so storing the whole Txn inputs/outputs again would be inefficient. However, it does take a little bit of thinking to understand that's the case.

solidity/contracts/domains/noto/Noto.sol
```diff
@@ -287,10 +318,27 @@ contract Noto is EIP712Upgradeable, UUPSUpgradeable, INoto, INotoErrors {
         bytes calldata signature,
```
jimthematrix: I don't think this has changed in this PR, but want to spell out in my words the lock lifecycle for my understanding, along with questions/observations. this is also somewhat related to the question from @peterbroadhurst above.

- create: when creating a lock, the lockId is calculated based on the lock parameters, and no delegate is specified
  - the same inputs can be locked multiple times, as long as I have different outputs or lockedOutputs, since the only validation is on the calculated lockId. this seems to be a problem?
- prepare unlock (optional): a hash is provided that commits the intended way to spend the lock, including the portion of the locked states to be spent (partial spend is allowed). this unlocked hash will be checked against the later unlock operation for consistency. a subsequent unlock operation not conforming to the committed spending parameters will be rejected
  - if not prepared, a lock can be spent in any way the sender wants. what this means is until a prepare is called, a lock is not a commitment yet
- delegate (optional): if not delegated, the only the notary can spend the lock, assuming the notary has validated in "offchain" logic that the owner of the lock has submitted the request.
  - should the check for the owner vs. unlock request submitter be done onchain, since the lock's owner is already captured in the onchain LockInfo state? not sure what the current criteria for onchain vs. offchain enforcement is, but intuitively it seems onchain enforcement is better if both options are available without requiring to track additional onchain states
- unlock: can be done by the notary if not yet delegated, or must be called by the delegate. if prepared, then the spending must match the intended way by the preparation. clear logic here and no further questions.

Issue #676, opened 2025-06-02
Noto/Zeto should perform on-chain enforcement of TX ID uniqueness


awrichar:
### What would you like to be added?

Similar to the work done for Pente in https://github.com/LF-Decentralized-Trust-labs/paladin/pull/472, Noto and Zeto should provide uniqueness checking for transaction IDs at the base ledger level.

Note: There have been a few offline discussions about the best/most efficient ways to do this without bloating storage. However, we should prioritize adding the protection in some form, even if efficiency discussions need to carry on further.

### Why is this needed?

This ensures the contract is not vulnerable to replay attacks (either malicious or accidental due to bugs/crashes during assembly).

jimthematrix:
Zeto is already enforcing uniqueness to protect against replay attacks, as well as a number of other types of attacks, by checking that output UTXOs are unique. This ensures each transaction is unique and replay attacks are not possible, without having to introduce a transaction ID.

In fact, given that for both Noto and Zeto the onchain state is represented in several collections of UTXOs (spent, unspent, locked, etc.), simply using unique TX IDs to prevent replay attacks is not sufficient. Zeto currently has two implementations of `validateOutputs()` based on usage of nullifiers or not:
- https://github.com/hyperledger-labs/zeto/blob/main/solidity/contracts/lib/storage/base.sol#L71
- https://github.com/hyperledger-labs/zeto/blob/main/solidity/contracts/lib/storage/nullifier.sol#L67

These types of checks should not be unique to Zeto, but common to any UTXO based token implementations. Once Noto also performs the same checks, a unique TX ID is no longer needed for replay protection, because that's already addressed by checking that the outputs are unique and new.

awrichar:
I think they are two different things - the uniqueness of the transaction ID prevents you from doing the same thing twice with different states. If Paladin was given the directive "send $100 to Jim", then the uniqueness check on the transaction ID will give final protection that Paladin won't somehow assemble twice (with different input/output combinations) and send Jim $200.

PR #922, opened 2025-11-27
Use correct noto tx id

annamcallister:
Noto domain receipts have been missing states- this looks to be because transaction ID is not actually available in the data of a `NotoUnlock` event, so the code falls back generating a random ID (now logged at WARN level), which means the states and receipt cannot be correlated.

### DIFF
--- a/domains/noto/internal/noto/events.go
+++ b/domains/noto/internal/noto/events.go
@@ -92,7 +92,11 @@ func (n *Noto) handleV0Event(ctx context.Context, ev *prototk.OnChainEvent, res 
 		log.L(ctx).Infof("Processing '%s' event in batch %s", ev.SoliditySignature, req.BatchId)
 		var unlock NotoUnlock_V0_Event
 		if err := json.Unmarshal([]byte(ev.DataJson), &unlock); err == nil {
-			n.recordTransactionInfo(ev, unlock.TxId, nil, res)
+			txData, err := n.decodeTransactionData(ctx, unlock.Data)
+			if err != nil {
+				return err
+			}
+			n.recordTransactionInfo(ev, txData.TransactionID, txData.InfoStates, res)
 			res.SpentStates = append(res.SpentStates, n.parseStatesFromEvent(unlock.TxId, unlock.LockedInputs)...)
 			res.ConfirmedStates = append(res.ConfirmedStates, n.parseStatesFromEvent(unlock.TxId, unlock.LockedOutputs)...)
 			res.ConfirmedStates = append(res.ConfirmedStates, n.parseStatesFromEvent(unlock.TxId, unlock.Outputs)...)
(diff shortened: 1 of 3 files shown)


domains/noto/internal/noto/events.go
```diff
@@ -162,7 +158,7 @@ func (n *Noto) handleV1Event(ctx context.Context, ev *prototk.OnChainEvent, res
 			}
 			// For V1, TransactionID is not in TransactionData, but Transfer event doesn't have TxId either
 			// The TransactionID will be set from decodeTransactionData fallback (random if zero)
```
awrichar: Hmm this seems wrong. Might have been a 🤖 change in my original PR. We definitely want to propagate the TxId in one place or the other for V1 (not use a random one).

domains/noto/internal/noto/events.go
```diff
@@ -224,7 +216,7 @@ func (n *Noto) handleV1Event(ctx context.Context, ev *prototk.OnChainEvent, res
 			if err != nil {
 				return err
 			}
-			n.recordTransactionInfo(ev, txData, res)
+			n.recordTransactionInfo(ev, txData.TransactionID, txData.InfoStates, res)
```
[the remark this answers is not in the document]
peterbroadhurst: The data in this case does include the TransactionID:
https://github.com/LFDT-Paladin/paladin/blob/09a6e50e51ecbab7483c90af1ed2cad4e4e844f6/domains/noto/internal/noto/handler_prepare_unlock.go#L133-L136

domains/noto/internal/noto/events.go
awrichar: A historical note - we originally encoded it only in the data. Later some methods/events ended up with a txId as a first-class parameter as part of the work #676, so in many cases we were passing it in both places. However, this was done somewhat inconsistently and without much regard for backwards compatibility since it was still during the earlier evolution of Noto.

With V1 I am attempting to standardize primarily on passing txId as a distinct parameter, except in cases where passing it via data is the only viable route (such as some of the V0 method+event pairs).

peterbroadhurst: I note this is the prepare transaction, not the unlock transaction that has been prepared.

peterbroadhurst:
@annamcallister @awrichar - after much conversation and digging, I've made the following proposal in code:

- `V0` code paths continue to decode `txData.TransactionID` for all events
    - Preserves behavior unchanged
    - Note that the warning will be output in V0

In `main` notice that the TXID is random, and there was no data - so I'm not sure I understand how this worked previously (maybe it never confirmed the transaction and V1 was needed for that - see below):
https://github.com/LFDT-Paladin/paladin/blob/80a576f4da895df7d30c70877be3f031e694b43a/domains/noto/internal/noto/receipts.go#L82-L89

- `V1` code paths use `[event].TxID` in all cases
    - Had to fix the Go structure for transfer event, as it was broken there in the definition
    - There was a comment (see discussion above) that transfer didn't have this
    - The `data` encoded transaction ID is superfluous in V1 - might need some discussion on removal

In `v1-develop` we have this code:
https://github.com/LFDT-Paladin/paladin/blob/09a6e50e51ecbab7483c90af1ed2cad4e4e844f6/domains/noto/internal/noto/receipts.go#L90-L115

Now in the V0 fallback path there is a `data` now (which there wasn't in main) - but if you trace it back, it strikes me the TXID in there isn't the `unlockTxId`, but rather the `txId` of the transaction that _prepared_ the lock.

domains/noto/internal/noto/events.go
peterbroadhurst: @awrichar - I wonder if we should have a V1 version of the encoding of the data (separate eyecatcher), which does not include the TxID parameter. As we move forwards to finalize V1 interfaces, it seems odd to have this duplication in there that is only there for V0 data decoding.

Issue #254, opened 2024-10-02
Core unit test timed out in TestRegistryRegisterFail


awrichar:
Unclear if this is a stuck test or just an accumulation of long tests that hit the 30s limit.

https://github.com/kaleido-io/paladin/actions/runs/11133626491/job/30940116525?pr=244

```
    plugin_base_test.go:90: 
        	Error Trace:	/home/runner/work/paladin/paladin/core/go/internal/plugins/plugin_base_test.go:90
        	            				/home/runner/work/paladin/paladin/core/go/internal/plugins/unittest_plugin_loader.go:92
        	            				/opt/hostedtoolcache/go/1.22.7/x64/src/runtime/asm_amd64.s:1695
        	Error:      	Received unexpected error:
        	            	EOF
        	Test:       	TestRegistryRegisterFail
coverage: 92.9% of statements
panic: test timed out after 30s
```

peterbroadhurst:
Hit again here: https://github.com/LFDT-Paladin/paladin/actions/runs/19865176022/attempts/3?pr=923

PR #924, opened 2025-12-02
Fix or debug potential for #254 intermittent failure

peterbroadhurst:
- In chain with #923 
- Potential fix for #254 intermittent failure
- Should hopefully give more explicable failure, if not root cause

### DIFF
--- a/core/go/internal/plugins/plugin_base_test.go
+++ b/core/go/internal/plugins/plugin_base_test.go
@@ -16,6 +16,7 @@ package plugins
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


core/go/internal/plugins/plugin_base_test.go
```diff
@@ -94,7 +95,9 @@ func (tp *mockPlugin[T]) Run(grpcTarget, pluginId string) {
 		if tp.sendRequest != nil {
 			req := tp.sendRequest(pluginId)
 			err := stream.Send(req)
-			require.NoError(t, err)
+			if tp.expectClose == nil {
+				require.NoError(t, err)
```
peterbroadhurst: Hypothesis is that one of these `require` statements is exiting the thread without calling `expectClose()`

core/go/internal/plugins/plugin_base_test.go
```diff
@@ -174,7 +182,7 @@ func TestPluginRequestsError(t *testing.T) {
 		return tdm, nil
 	}
 	tdm.findAvailableStates = func(ctx context.Context, req *prototk.FindAvailableStatesRequest) (*prototk.FindAvailableStatesResponse, error) {
-		return nil, fmt.Errorf("pop")
+		return nil, NewPluginError(prototk.Header_INVALID_INPUT, fmt.Errorf("pop"))
```
peterbroadhurst: this is just test coverage

PR #925, opened 2025-12-03
Manifests and `complete_only` receipt listener mode

peterbroadhurst:
## Problem statement

Token transactions distribute different sets of states to different parties:
- The sender sees the whole transaction, including the coins they chose to use, and the amount sent
- The receiver only sees the value transferred to them (not the remainder returned to the sender)

The blockchain receipts usually arrive after states have propagated to each party, but this certainly cannot be guaranteed. There are many circumstances that could cause state delivery to be delayed, due to temporary downtime of a participant Paladin node, or simply timing of delivery.

It must be possible to have a reliable stream of receipts that provide the full view of a transaction that your node is allowed to see. The "domain receipt" as Paladin terms it, which is decoded by the domain into a business transaction (a "transfer" etc.). Blockchain applications rely on this for off-chain indexing of transactions, and triggering business logic. These are critical functions for that Paladin needs to support reliably.

This means Paladin needs to have a mode of a receipt listener, where it tracks all transactions that it has indexed without full availability of the states. Then when states arrive that provide more information to potentially decode that transaction, the Paladin node can re-assess whether the transaction is now "complete" for this node, and should be delivered as an event to the applications listening.

Noting that because there are multiple states that might be required per transaction, but not all states will be distributed to all parties, there needs to be a per-transaction concept of "complete".

The model introduced is that transactions supporting partial view by different participants, must implement a `manifest` that allows the domain to check node-by-node whether all the expected states are available.

> Note that absolute order of delivery of receipts, cannot be assured on a receipt listener in combination with deferral-of-delivery of receipts until the data is available. That is because in a privacy-preserving system, you cannot know whether a transaction is "for you" until you receive some private data for it. So when delivery of that private data is delayed, the receipt is also delayed, potentially until after other receipts have been delivered that had full data available.

## Implementation summary

The following are the related parts of this PR:

1. A new `complete_only` option for receipt listeners
    - Does not deliver a transaction to the receipt listener, until state availability is "complete"
    - Allows a domains to determine what complete looks like for a transaction
    - Does not guarantee order of delivery in this mode, as deferred-receipts are delivered when states arrive
    - Only introduces domain-invocation overhead when states for a transaction are partially complete
2. A set of new APIs to help a domain implement a `manifest` approach to completion
    - Strongly encourages the use of an `info` state (the first) as the way to document which party requires which state
    - `CheckStateCompletion()` must be implemented by domains, to identify required states in incomplete transactions
    - `ValidateStates()` function for domains to generate the IDs for states in-line during assembly
    - `LookupKeyIdentifiers()` function to determine which on-chain addresses are "local" (during complete-checking)
3. A full implementation for Noto
    - Defines a new `NotoManifest` state object
    - Distributes this as the first `info` state in each transaction type Noto supports
    - Builds it based on the existing distribution of states
    - Minimizes code structure changes to the Noto package in this PR
4. Code maintenance found during implementation
    - Reinstating coverage in packages I was working in that had dropped coverage recently
    - A number of minor test stability fixes, which were blocking development for me

## Detail on Paladin handling of `complete_only`

> See previous conversations in https://github.com/LFDT-Paladin/paladin/pull/711#pullrequestreview-3534931151

A new DB record created that is very similar to a `gap`, but is different because it points at a _single missing state_ for an _individual transaction_.

- Doesn't require invoking the domain to check completion, unless we have the manifest
    - By designating that any domain supporting _partial_ completion (not 0% or 100%) *must* have an `info` state that everybody involved in the transaction expects
- Can use a join to list these incomplete receipts efficiently when a state arrives
    - By requiring the domain to always identify a **single** state that is the "next unavailable" when it is called with at least one `info` state.


core/go/build.gradle
```diff
@@ -28,6 +28,7 @@ ext {
         include "internal/**/*.go"
         include "pkg/**/*.go"
         include "componenttest/**/*.go"
+        include "noderuntests/**/*.go"
```
annamcallister:
```suggestion
        include "noderuntests/**/*.go"
```

`componenttest/` has been removed

doc-site/docs/reference/types/_includes/transactionreceiptlistener_description.md
```diff
@@ -20,7 +20,11 @@
 }
 ```
 
-Note the ability to filter on particular receipt types, and most importantly the ability to block delivery on a given contract address while a "gap" exists in the available states.
+Note the ability to filter on particular receipt types, and most importantly the ability to control delivery behavior when states are incomplete:
+
+- **`block_contract`** (default): Pauses delivery for each individual smart contract address when incomplete states are detected
```
awrichar: Really unsure what our default should be... but I guess in the absence of a clearly "better" choice, we just stick with what it's been. All of them are problematic if you don't understand what you're asking for.

toolkit/proto/protos/to_domain.proto
```diff
@@ -257,6 +257,27 @@ message WrapPrivacyGroupEVMTXResponse {
...
+  string transaction_id = 1; // the transaction ID
+  repeated EndorsableState input_states = 2; // available input states
+  repeated EndorsableState read_states = 3; // available read states
+  repeated EndorsableState output_states = 4; // available output states
+  repeated EndorsableState info_states = 5; // available info states
+  UnavailableStates unavailable_states = 6; // details of unavailable states by classification, including a could for a quick completion check
+}
+
+message CheckStateCompletionResponse {
+  optional string primary_missing_state_id = 1; // if set, then this transaction is considered incomplete until this state is available - at which point the completion check will run again with the new set of states available at that time
```
awrichar: Is there anything that makes one state more "primary" than others? Is this "I'm missing at least one state" or "I'm missing _the primary_ state"?

domains/noto/internal/noto/handler_mint.go
```diff
@@ -96,6 +105,17 @@ func (h *mintHandler) Assemble(ctx context.Context, tx *types.ParsedTransaction,
 		return nil, err
 	}
 
+	if !tx.DomainConfig.IsV0() {
```
awrichar: Is there a reason we don't include this for V0 Noto? Is there something about the new Solidity in V1 that enables manifests?

awrichar: It seems like this is a purely off-chain behavior and not tied to the ABI of Noto V0 vs V1.

domains/noto/internal/noto/states.go
```diff
@@ -199,16 +211,55 @@ type preparedLockedInputs struct {
...
 
 type preparedLockedOutputs struct {
-	coins  []*types.NotoLockedCoin
-	states []*prototk.NewState
+	distributions []identityList
+	coins         []*types.NotoLockedCoin
+	states        []*prototk.NewState
+}
+
+type identityPair struct {
```
awrichar: This is a struct we tend to need often... I've been using it in lots of apps and happened to pick the name "resolved identity". No need to change anything here, but just an observation that it seems to be a common pattern of Paladin development.

domains/noto/internal/noto/noto.go
```diff
@@ -1036,6 +1043,96 @@ func (n *Noto) WrapPrivacyGroupEVMTX(ctx context.Context, req *prototk.WrapPriva
...
+	// Now, it get's a little complex - we need to ask the Paladin node which of the addresses
+	// in the state distribution list are "ours". There's a batch API for this provided.
+	// Note we only get to this point if we're involved in the transaction in some way, and
+	// don't have the whole state set (Notary always has full set before submit).
+	// So a bit of efficient in-memory processing overhead is perfectly acceptable.
+	lookupReq := &prototk.LookupKeyIdentifiersRequest{
+		Algorithm:    algorithms.ECDSA_SECP256K1,
+		VerifierType: verifiers.ETH_ADDRESS,
+	}
+	uniqueAddresses := make(map[string]struct{})
```
awrichar: Small detail here, but why `struct{}` and not `bool`?

toolkit/proto/protos/from_domain.proto
```diff
@@ -98,6 +100,31 @@ message GetStatesByIDResponse {
   repeated StoredState states = 1;
 }
 
+message LookupKeyIdentifiersRequest {
```
awrichar: This method is called "LookupKeyIdentifiers", but it maps to a method on key manager called "ReverseKeyLookup". Any reason to not use "ReverseKeyLookup" as the name, and/or more closely align the response ("identifier" instead of "key_identifier")?

Is there a reason that algorithm/verifier_type are global to the request instead of per-verifier?

toolkit/proto/protos/from_domain.proto
```diff
@@ -98,6 +100,31 @@ message GetStatesByIDResponse {
...
+  string algorithm = 1;
+  string verifier_type = 2;
+  repeated string verifiers = 3;
+}
```
awrichar: I notice that we don't include a `state_query_context` in this request like we do in some other "from domain" methods, even though (I think) it is a database query method. Should we consider if that's needed?

core/go/internal/domainmgr/domain.go
```diff
@@ -860,6 +860,75 @@ func (d *domain) GetStatesByID(ctx context.Context, req *prototk.GetStatesByIDRe
...
+			Schema: schema.ID(),
+			Data:   pldtypes.RawJSON(s.StateDataJson),
+		}
+		if isOutput {
+			// These are marked as locked and creating in the transaction, and become available for other transaction to read
+			stateUpsert.CreatedBy = &createdByTX.ID
+		}
+		newStatesToWrite[i] = stateUpsert
+	}
+	return newStatesToWrite, nil
```
awrichar: The naming of this return is now confusing - I know this was copied from elsewhere, but in the `ValidateStates` flow we won't actually be writing these states anywhere yet.

Issue #952, opened 2025-12-04
bug: Sequencer shouldn't re-assemble transactions with chained transactions already persisted


matthew1001:
### What happened?

 - Noto call invokes a Pente hook contract
 - The Pente hook reverts (reason unrelated to this bug)

2 issues were observed:

1. The revert wasn't propagated to the original transaction, so it didn't get a receipt and the sequencer kept it in the pool of transactions to re-assemble
2. On restart, the sequencer attempted to resume the original Noto transaction. This is theoretically valid (excepting for issue `1` above), but it created a new chained transaction result in 1 Noto with 2 chained transactions in the DB.

On this last point, actually the 2nd chained transaction couldn't be entered into the DB because of idempotency key clash of `<original-noto-tx-id>_functionName`

### What did you expect to happen?

The revert should cause the original Noto TX to be confirmed (failed)

### How can we reproduce it (as minimally and precisely as possible)?

Deploy a noto with a chained Pente transaction where the Pente transaction reverts

### OS version

Nothing specific about OS

matthew1001:
https://github.com/LFDT-Paladin/paladin/pull/956 has merged, closing the issue

Issue #929, opened 2025-12-04
bug: Pass sequencer configuration to coordinator and originator as a single object


matthew1001:
### What happened?

See PR comment https://github.com/LFDT-Paladin/paladin/pull/704#discussion_r2483000007 as part of the new sequencer integration with Paladin

### What did you expect to happen?

N/A

### How can we reproduce it (as minimally and precisely as possible)?

N/A

### Anything else we need to know?

N/A

### OS version

N/A

PR #956, opened 2025-12-05
Ensure chained transactions aren't recreated after a node outage

matthew1001:
Fixes https://github.com/LFDT-Paladin/paladin/issues/952

This PR introduces 2 basic enhancements to the distributed sequencer:

1. It prevents a transaction that has already had a child/chained transaction dispatched, from assembling a new chained transaction on node restart
2. It also ensures that an originator periodically re-delegates any of its transactions that haven't successfully moved out of `State_Delegated`.

There is existing logic whereby an originator checks heartbeats from coordinators for the presence of a delegated TX, and if absent, re-delegates the TX. However, there are cases where no heartbeats will be received. A node stop/start scenario is one, where delegation fails (e.g. coordinator endpoint not yet available) and no other transactions for that domain contract are submitted to nudge the coordinator into life.

This PR introduces an additional poll via the `DelegateTimeout` configuration option (default `5s`). This means that by default, on startup and/or after a failure to delegate an originator will re-delegate after `5s`.

If a heartbeat is received in that time, the delegate timeout will be skipped because the time since last delegation time must be `> DelegateTimeout`. So a healthy coordinator with the originator's TX in its pool will prevent an originator needing to re-delegate, even if the coordinator is taking it's time to move it to assembling.

Since this PR required another config point for the sequencer I took the opportunity to address [another issue](https://github.com/LFDT-Paladin/paladin/issues/929) raised on the sequencer PR, namely to pass the config struct to the coordinator and originator instead of individual config params.

### DIFF
--- a/core/go/internal/sequencer/common/types_test.go
+++ b/core/go/internal/sequencer/common/types_test.go
@@ -1,0 +1,94 @@
+/*
+ * Copyright © 2025 Kaleido, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package common
+
+import (
+	"testing"
+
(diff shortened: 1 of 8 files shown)


core/go/internal/sequencer/originator/originating.go
```diff
@@ -27,22 +27,27 @@ import (
...
+			// only re-delegate after the delegate timeout
 			privateTransactions = append(privateTransactions, txn.PrivateTransaction)
-		} else if txn.GetCurrentState() == transaction.State_Pending {
+		}
+
+		if txn.GetCurrentState() == transaction.State_Pending {
 			privateTransactions = append(privateTransactions, txn.PrivateTransaction)
 		}
+
+		txn.UpdateLastDelegatedTime()
```
annamcallister: is this updating the last delegated time for transactions that aren't being delegated?

PR #957, opened 2025-12-08
Noto nullifiers

jimthematrix:
(should be merged after #978 and [#1148](https://github.com/LFDT-Paladin/paladin/pull/1148))

Adding the capability to spend Noto UTXOs with nullifiers to mask the spending history.

## Nullifiers

Given that the commitments for a Noto token is `EIP712Hash(amount, salt, owner_address)`, a typical nullifier scheme would be `EIP712Hash(amount, salt, owner_priv_key)`. However, in Noto we do not make use of ZKPs, which means when the nullifier submitted by the sender is verified by the notary, the sender can not leak its private key to the notary. As such, we design the nullifier hashing scheme differently, by using just `EIP712Hash(amount, salt)`. 

This should still preserve the main security guarantees of a nullifier:
- **unique** binding to the UTXO commitment: due to the uniqueness of the salt, the nullifier is guaranteed to be 1:1 bound to the UTXO commitment
- **secret** binding to the UTXO commitment: because the salt is only known by the sender and the receiver, no other parties are able to construct the nullifier

This design lacks the guarantee that only the owner of the UTXO commitment is able to construct the nullifier, because the sender also knows the secrets required to construct it (besides the receiver which is the owner). However, this is mitigated by the fact that the notary already checks that the sender must be the rightful owner of the input UTXO, thus completing the security requirements to make this design secure.

## Onchain storage

Unlike the regular Noto contract that tracks the UTXOs in a simple map for spent status, the UTXOs in this new NotoNullifiers contract are tracked inside a Sparse Merkle Tree (SMT). This way, a transaction submitter can submit these parameters to the notary, as part of the private transaction payload, for verification:

- `input UTXOs`
- `nullifiers`
- `outputUTXOs`

The job of the notary is to verify that:
- the input UTXOs are included in the SMT maintained by the smart contract
  - to do this, the notary must maintain its own SMT
- the nullifiers are the proper ones corresponding to the input UTXOs
- the input UTXOs and the output UTXOs obey the spending rules (mass conservation, etc.)

Then the notary submits the following (public) parameters to the NotoNullifiers contract:
- `nullifiers`
- `output UTXOs`
- `merkle tree root` resulted from verifying the input UTXOs against the notary's SMT (this does NOT need to be the latest root, just needs to be one of the legitimate historical root)

The job of the smart contract is to verify that:
- the root is a legitimate historical root (validity of the input UTXOs represented by the nullifiers)
  - this check is a safety net for cases when the notary makes honest mistakes and submits transactions out of order, which would attempt to consume UTXOs that have not been produced
- the nullifiers are not seen before (prevents double spending)
- the notary certificate, which is as usual, same as when not using nullifiers

### Sparse Merkle Tree supporting configurable hashing algos
This solution depends on a Sparse Merkle Tree that uses keccak256 hashing, rather than the Poseidon hashing as seen in the [current implementation](https://github.com/iden3/contracts/blob/master/contracts/lib/SmtLib.sol). The Poseidon hashing is used for its efficiency inside ZKP circuits, but it's much slower in EVM compared to the "native" hashing like keccak256.

Since for Noto uses we do not need to generate ZK proofs, we want to have an SMT implementation that's based on keccak256, making transaction processing much cheaper when the onchain SMT needs to be updated with new UTXOs.

This is accomplished by enhancing the implementations in Solidity (to use in the NotoNullifiers contract), typescript (to use in hardhat tests), and golang (to use in Paladin's domain implementation):
- Solidity: https://github.com/iden3/contracts/pull/402
- typescript: https://github.com/iden3/js-merkletree/pull/53
- golang: https://github.com/LFDT-Paladin/smt (refactored out of LFDT-Paladin/zeto/go-sdk)

### SMT Storage implementation using Paladin's state storage
This was originally implemented for the Zeto domain. It's now moved into a common golang package under `toolkit/go/pkg/smt`, to be shared by the Noto and Zeto domains.

```
              github.com/LFDT-Paladin/smt
            (Sparse Merkle Tree implementation)
                             |
                             |
                    toolkit/go/pkg/smt
                   (SMT storage plugin)
                             |
              --------------------------------
              |                              |
        Noto Domain                     Zeto Domain
```


solidity/contracts/domains/noto/NotoFactory.sol
```diff
@@ -52,12 +59,18 @@ contract NotoFactory is Ownable, IPaladinContractRegistry_V0 {
      */
     function deployImplementation(
         bytes32 transactionId,
+        string calldata implName,
```
awrichar: This reminds me that https://github.com/LFDT-Paladin/paladin/pull/964 is still open. You might want to sync up with @annamcallister.

solidity/contracts/domains/noto/NotoNullifiers.sol
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    function buildConfig(
+        bytes calldata data
+    ) external view override returns (bytes memory) {
+        return
+            _encodeConfig(
+                NotoConfig_V1({
+                    name: _name,
+                    symbol: _symbol,
```
awrichar: Maybe you should just use `name()` and `symbol()` here (like with `decimals()` below) and make them `public` instead of `external` - which avoids needing the variables to be `internal` instead of `private`. It feels slightly better to me than increasing the private variable visibility.

solidity/contracts/domains/noto/NotoNullifiers.sol
```diff
@@ -0,0 +1,169 @@
...
+        if (!_commitmentsTree.rootExists(root)) {
+            revert NotoInvalidRoot(root);
+        }
+
+        _processNullifiers(args.inputs);
+        _processOutputs(args.outputs);
+        _processLockContents(lockId, args.contents);
+
+        _processOutput(args.newLockState);
+        _lockStates[lockId] = args.newLockState;
```
awrichar: Feels like maybe this should also have an `internal` helper to avoid exposing `_lockStates` directly. Maybe a `_initLockState` near `_transitionLockState` in the base. I think it's cleanest if the parent contract doesn't expose its storage fields.

solidity/contracts/domains/noto/NotoNullifiers.sol
```diff
@@ -0,0 +1,169 @@
...
+            // record the nullifier as used
+            _nullifiers[inputNullifiers[i]] = true;
+        }
+    }
+
+    // check the existence of a UTXO in the commitments tree. we take a shortcut
+    // by checking the list of nodes by their node hash, because the commitments
+    // tree is append-only, no updates or deletions are allowed. As a result, all
+    // nodes in the list are valid leaf nodes, aka there are no orphaned nodes.
+    function existsAsUnlocked(uint256 utxo) internal view returns (bool) {
```
awrichar: nit: usually `_` prefix on internal methods

solidity/contracts/domains/noto/NotoNullifiers.sol
```diff
@@ -0,0 +1,169 @@
...
+            }
+            // record the nullifier as used
+            _nullifiers[inputNullifiers[i]] = true;
+        }
+    }
+
+    // check the existence of a UTXO in the commitments tree. we take a shortcut
+    // by checking the list of nodes by their node hash, because the commitments
+    // tree is append-only, no updates or deletions are allowed. As a result, all
+    // nodes in the list are valid leaf nodes, aka there are no orphaned nodes.
```
awrichar: nit: can we consistently use `/** */` style to match the file?

solidity/contracts/domains/noto/NotoNullifiers.sol
```diff
@@ -0,0 +1,169 @@
...
+    function existsAsUnlocked(uint256 utxo) internal view returns (bool) {
+        uint256 nodeHash = getLeafNodeHash(utxo, utxo);
+        SmtLib.Node memory node = _commitmentsTree.getNode(nodeHash);
+        return node.nodeType != SmtLib.NodeType.EMPTY;
+    }
+
+    function getLeafNodeHash(
```
awrichar: same here with `_`

solidity/contracts/domains/noto/NotoNullifiers.sol
```diff
@@ -0,0 +1,169 @@
...
+        uint256 index,
+        uint256 value
+    ) internal pure returns (uint256) {
+        uint256[3] memory params = [index, value, uint256(1)];
```
awrichar, line 180: Maybe a comment on what the params mean (especially the `1`)?

solidity/hardhat.config.ts
```diff
@@ -5,13 +5,36 @@ import { HardhatUserConfig } from "hardhat/config";
...
+          },
+          viaIR: true,
+        },
+      }
+    ]
+  },
+  networks: {
+    hardhat: {
+      // a small number of verifiers are bigger than the default 24576 bytes
+      // so we need to allow unlimited contract size to avoid test errors
```
awrichar, line 36: Will this cause practical issues on real networks?

solidity/package.json
```diff
@@ -10,7 +10,9 @@
   "author": "",
   "license": "Apache-2.0",
   "devDependencies": {
-    "@iden3/contracts": "^2.3.0",
+    "@iden3/contracts": "git+https://github.com/kaleido-io/contracts.git#keccak256",
+    "@iden3/js-merkletree": "file://./js-merkletree",
+    "@lfdecentralizedtrust/zeto-contracts": "v0.2.2",
```
awrichar, line 15: Was it intentional to also upgrade the Zeto contract version in this PR?

solidity/contracts/domains/noto/NotoNullifiers.sol
```diff
@@ -0,0 +1,169 @@
...
+
+import {SmtLib} from "@iden3/contracts/contracts/lib/SmtLib.sol";
+import {Noto} from "./Noto.sol";
+
+uint256 constant MAX_SMT_DEPTH = 64;
+
+contract NotoNullifiers is Noto {
+    using SmtLib for SmtLib.Data;
+
+    SmtLib.Data internal _commitmentsTree;
```
awrichar: Should this be private?

solidity/hardhat.config.ts
```diff
@@ -5,13 +5,36 @@ import { HardhatUserConfig } from "hardhat/config";
...
+    compilers: [
+      {
+        version: "0.8.27",
+        settings: {
+          evmVersion: "shanghai",
+          optimizer: {
+            enabled: true,
+            runs: 200,
+          },
+          viaIR: true,
```
awrichar, line 17: Note that we're enabling `viaIR` - previously we'd left it off. For a long time this was still "experimental" although it worked well. Was this needed to compile the contracts, or just a forward-looking switch (or both)?

awrichar: Maybe it's more recommended now.

solidity/contracts/domains/noto/NotoNullifiers.sol
```diff
@@ -0,0 +1,169 @@
...
+                    symbol: _symbol,
+                    decimals: decimals(),
+                    notary: notary,
+                    variant: NotoVariantV2Nullifiers,
+                    data: data
+                })
+            );
+    }
+
+    function transfer(
```
awrichar: Maybe should override the internal `_transfer` method instead, just for consistency with the other overrides.

solidity/contracts/domains/noto/NotoNullifiers.sol
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    function _updateLock(
+        NotoUpdateLockArgs memory args,
+        bytes32 spendCommitment,
+        bytes32 cancelCommitment,
+        bytes32 lockId,
+        NotoLockInfo storage lock
+    ) internal virtual override {
+        if (args.proof.length > 0) {
```
awrichar: Should `proof` be required?

solidity/contracts/domains/noto/NotoNullifiers.sol
```diff
@@ -0,0 +1,169 @@
...
+            spendCommitment,
+            cancelCommitment,
+            lockId,
+            lock
+        );
+    }
+
+    /**
+     * @dev Lock state IDs and other outputs live in the commitment tree, not _unspent.
+     *      Base Noto spends lock states via _processInput; accept tree membership there.
```
awrichar: The comment here is a little confusing. I think what I've understood is:
- Regular inputs (coins) are now stored in the merkle tree (not in a map). However, spending those input coins no longer comes through this path at all - because you've overridden the two methods that would have used this method (`_transfer` and `_createLock`) to use `_processNullifiers` instead for spending inputs (which checks the nullifiers map, not the states merkle tree).
- The only other remaining paths that use this method are those that spend/transition lock states. Lock states don't require nullifiers, so this method should verify merkle tree membership of those lock states.

awrichar: Hmm even after writing this down, I'm still struggling with questions. Is the merkle tree only used for lock states? It seems like we append all outputs to the tree, but this is the only place where we check them.

awrichar: Maybe it's two different patterns with the same tree... the regular inputs are verified by the notary and passes as a root, while lock states are verified here directly?

I think some more comments in the code might help. I like the PR description, but I think I'm just not connecting all the dots.

solidity/test/domains/noto/NotoNullifiers.ts
```diff
@@ -0,0 +1,505 @@
...
+          noto,
+          [txo1.nullifier!, txo2.nullifier!],
+          [txo3.hash!],
+          [locked1.hash!],
+          root.bigInt().toString(10),
+          randomBytes32()
+        ));
+        // only the outputs are tracked in the merkle tree.
+        // the locked outputs are tracked in the locked states map and
+        // to be consumed using the UTXO ids instead of nullifiers
```
awrichar: I think I'm struggling with the various places things are tracked.
- unlocked states -> tracked in merkle tree, spent via nullifiers which are tracked in a map
- locked states -> tracked and spent via map
- lock states (the lock itself) -> tracked and spent via map

I don't fully understand why it's broken down this way. Might need explanation directly in the Solidity contract.

solidity/contracts/domains/noto/NotoNullifiers.sol
```diff
@@ -0,0 +1,169 @@
...
+import {Noto} from "./Noto.sol";
+
+uint256 constant MAX_SMT_DEPTH = 64;
+
+contract NotoNullifiers is Noto {
+    using SmtLib for SmtLib.Data;
+
+    SmtLib.Data internal _commitmentsTree;
+
+    uint64 public constant NotoVariantV2Nullifiers = 0x0003;
```
awrichar: Hmm so if we just increase this, we'll end up with the release history being a mix of different types of contract. Need to think about options...
- Keep it: just document a version table and note that there are many variants of Noto and not a single stream.
- Use a different digit: increment the 2nd or 3rd digit for Nullifier variants instead of the last (not likely to scale if we have additional variants though).
- Move away from sequential numbers: use a hash or randomized value for each variant.

core/go/build.gradle
```diff
@@ -155,27 +155,36 @@ task protoc(type: ProtoCompile, dependsOn: [
     }
 }
 
+// Invoke mockery v3 via `go run`: GOBIN/PATH often has an unrelated v2 binary also named `mockery`,
+// which rejects this repo's v3-format YAML (pkgname/template at top level).
+def mockeryGoPkg = 'github.com/vektra/mockery/v3@v3.3.2' // keep in sync with root :installMockery
```
awrichar: This seems strange

core/go/internal/statemgr/statestore.go
```diff
@@ -181,12 +181,24 @@ func (ss *stateManager) GetTransactionStates(ctx context.Context, dbTX persisten
...
+			`ON ("states"."id" = "records"."state"
+			    OR ("records"."record_type" = 'spent'
+				    AND EXISTS (
+					    SELECT 1
+						  FROM "state_nullifiers"
+					    WHERE "state_nullifiers"."id" = "records"."state"
+              AND "state_nullifiers"."state" = "states"."id"
+					  )
+					)
+				)`,
```
awrichar, line 203: This feels significant. Why didn't we need this for Zeto nullifiers? I'm trying to decide if it's a concern. Might need a @peterbroadhurst check on this one.

domains/integration-test/noto_test.go
```diff
@@ -92,15 +112,22 @@ func toJSON(t *testing.T, v any) []byte {
...
 }
 
+func (s *notoTestSuite) TestNotoV1Nullifiers() {
+	// s.T().Skip()
+	s.testNoto("v1", "noto_nullifiers")
+}
+
 func (s *notoTestSuite) TestNotoV1() {
-	s.testNoto("v1")
+	s.T().Skip()
```
awrichar: Should we remove the "skip" annotations?

testinfra/besu_bootstrap/main.go
```diff
@@ -120,11 +120,12 @@ func main() {
...
-			ChainID:     1337,
-			LondonBlock: 0, // Enable EIP-1559 from genesis
-			CancunTime:  0,
-			ZeroBaseFee: ptrTo(zeroBaseFee),
-			MinGasPrice: minGasPrice,
+			ChainID:           1337,
+			LondonBlock:       0, // Enable EIP-1559 from genesis
+			CancunTime:        0,
+			ZeroBaseFee:       ptrTo(zeroBaseFee),
+			ContractSizeLimit: big.NewInt(24576 * 2), // to accommodate contracts in NotoNullifiers
```
awrichar: Hmm would like to avoid this if we can. Was it `NotoNullifiers` that got too big, or `NotoFactory`?

domains/noto/internal/noto/smt/merkle_tree.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+
+	pldsmt "github.com/LFDT-Paladin/paladin/toolkit/pkg/smt"
+	"github.com/LFDT-Paladin/smt/pkg/sparse-merkle-tree/core"
+	"github.com/LFDT-Paladin/smt/pkg/sparse-merkle-tree/smt"
+	"github.com/LFDT-Paladin/smt/pkg/utxo"
+	utxocore "github.com/LFDT-Paladin/smt/pkg/utxo/core"
+)
+
+const SMT_HEIGHT_UTXO = 64
```
awrichar, line 32: Maybe a note on why this number?

domains/noto/pkg/types/config.go
```diff
@@ -134,12 +136,36 @@ const (
 	NotaryModeIntHooks pldtypes.HexUint64 = 0x0001
 )
 
-var NotoVariantV2 pldtypes.HexUint64 = 0x0002 // V2 variant (latest)
-var NotoVariantV1 pldtypes.HexUint64 = 0x0001 // V1 variant
-var NotoVariantV0 pldtypes.HexUint64 = 0x0000 // V0 variant
+// NotoVariantV2Nullifiers is reported by NotoNullifiers.sol (nullifier inputs + commitment tree).
+var NotoVariantV2Nullifiers pldtypes.HexUint64 = 0x0003
```
awrichar: I noted in the Solidity as well, but this naming is getting confusing 🫠

domains/noto/pkg/types/config.go
```diff
@@ -134,12 +136,36 @@ const (
...
+
+// NotoVariantV2 is the on-chain variant reported by standard Noto.sol (flat lock ABI, coin state IDs).
+var NotoVariantV2 pldtypes.HexUint64 = 0x0002
+var NotoVariantV1 pldtypes.HexUint64 = 0x0001 // INoto_V1-shaped parameters (nested LockParams)
+var NotoVariantV0 pldtypes.HexUint64 = 0x0000 // legacy
+
+// Backward-compatible aliases used across handlers and tests
+var (
+	NotoVariantNullifier = NotoVariantV2Nullifiers
+	NotoVariantDefault   = NotoVariantV1
```
awrichar: I think `NotoVariantV2` is the default now...?

domains/noto/pkg/types/config.go
```diff
@@ -134,12 +136,36 @@ const (
...
+	NotoVariantLegacy    = NotoVariantV0
+)
 
 func (c *NotoParsedConfig) IsV2() bool {
-	return c.Variant == NotoVariantV2
+	return c.Variant == NotoVariantV2 || c.Variant == NotoVariantV2Nullifiers
+}
+
+// UsesFlatLockABI is true for modern Noto (standard or nullifiers): INoto / ILockableCapability parameter layout.
```
awrichar: Not sure what "flat lock" means - I like the idea of using these helpers, just might need to think more about how to differentiate.

domains/noto/pkg/types/states.go
```diff
@@ -20,6 +20,9 @@ import (
 	"github.com/hyperledger/firefly-signer/pkg/abi"
 )
 
+const PAYLOAD_DOMAIN_NOTO_NULLIFIER = "domain:noto:nullifier"
+const VERIFIER_DOMAIN_NOTO_NULLIFIER = "domain:noto:nullifier:verifier"
```
awrichar: Are these new constants for signing? Maybe they should be in a new file.

domains/noto/pkg/types/config.go
```diff
@@ -149,3 +175,7 @@ func (c *NotoParsedConfig) IsV1() bool {
 func (c *NotoParsedConfig) IsV0() bool {
 	return c.Variant == NotoVariantV0
 }
+
+func AlgoDomainNullifier(name string) string {
```
awrichar: Should this be with the other signing constants?

solidity/hardhat.config.ts
jimthematrix: forgot about this temporary hack while exploring ways to reduce the contract size under the EIP-170 limit. the size had been successfully reduced just under the limit (without having to do more dramatic refactoring in Solidity) so this can be removed.

solidity/package.json
jimthematrix: correct, this is the latest Zeto release in the v0.2.x stream.

solidity/hardhat.config.ts
jimthematrix: the IR based compiler pipeline is supposed to become the default according to the roadmap. it's been considered fully tested for a while now.

I went ahead and added that to 0.8.30 as well.

solidity/contracts/domains/noto/NotoNullifiers.sol
jimthematrix: agreed, the root should always be validated.

solidity/contracts/domains/noto/NotoNullifiers.sol
jimthematrix: Good question. It got me thinking that there's really no good reason for the lock states to be tracked in the merkle tree, while the locked outputs are tracked in the regular _unspent map. Plus there was a bug where the lock states were not properly spent (the `_processInput()` returns after simply checking the validity of the lock states).

Updated to have the lock states tracked "normally" as in Noto, the same way the locked outputs are handled.

solidity/test/domains/noto/NotoNullifiers.ts
jimthematrix: so the locked states and the lock themselves are tracked exactly the same way as in base Noto. this is because the locked states needs to be shown to the counterparty to be the ones actually spent by the spender, thus it wouldn't be desirable to use nullifiers to spend them (and thus hiding they were the ones being spent).

only the regular states are spent by nullifiers and thus tracked in the merkle tree.

solidity/contracts/domains/noto/NotoNullifiers.sol
jimthematrix: or maybe just a two-part identifier: `0x{variant}{version}`. So the base variant of Noto version would be `0x0002` as currently defined, and the nullifier variant would be `0x0102`.

core/go/build.gradle
jimthematrix: it's been eliminated (was a temporary workaround that should've been removed)

core/go/internal/statemgr/statestore.go
jimthematrix: pretty sure this is because Zeto hasn't implemented domain receipts. the Zeto domain supports queries via nullifiers with `pstate_queryContractNullifiers` which exercises a different query that already does the above join.

testinfra/besu_bootstrap/main.go
jimthematrix: this has been removed and reverted to the EIP-170 compliant setting

domains/noto/pkg/types/config.go
jimthematrix: wasn't my choice 😄  it's been eliminated and replaced with just using the existing `IsV2()`, as the two are equivalent

solidity/contracts/domains/noto/NotoNullifiers.sol
awrichar: That seems fine to me.

solidity/contracts/domains/noto/Noto.sol
```diff
@@ -51,15 +51,15 @@ contract Noto is EIP712Upgradeable, UUPSUpgradeable, INoto, INotoErrors {
             "Unlock(bytes32 txId,bytes32[] lockedInputs,bytes32[] outputs,bytes data)"
         );
 
-    string private _name;
-    string private _symbol;
+    string internal _name;
+    string internal _symbol;
```
awrichar: Are we able to keep all of the "private" symbols private (_name, _symbol, _lockStates) or is there a good reason to make them mutable by child contracts?

solidity/contracts/domains/noto/NotoNullifiers.sol
```diff
@@ -0,0 +1,179 @@
...
+        bytes32 txId,
+        bytes32[] memory inputs,
+        bytes32[] memory outputs,
+        bytes calldata proof,
+        bytes calldata data
+    ) internal virtual override {
+        (uint256 root, bytes memory _signature) = abi.decode(
+            proof,
+            (uint256, bytes)
+        );
```
awrichar: Could consider an internal helper for decoding the proof (since this same decode is done in many places).

testinfra/besu_bootstrap/main.go
```diff
@@ -120,11 +120,12 @@ func main() {
...
-			ChainID:     1337,
-			LondonBlock: 0, // Enable EIP-1559 from genesis
-			CancunTime:  0,
-			ZeroBaseFee: ptrTo(zeroBaseFee),
-			MinGasPrice: minGasPrice,
+			ChainID:           1337,
+			LondonBlock:       0, // Enable EIP-1559 from genesis
+			CancunTime:        0,
+			ZeroBaseFee:       ptrTo(zeroBaseFee),
+			ContractSizeLimit: big.NewInt(24576),
```
awrichar: Are we still overriding the default?

domains/noto/internal/noto/handler_burn_common.go
```diff
@@ -165,20 +179,29 @@ func (h *burnCommon) baseLedgerInvokeBurn(ctx context.Context, tx *types.ParsedT
 	functionName := "transfer"
 	var paramsJSON []byte
 
+	payload := sender.Payload
```
awrichar: nit: might want to call this local (here and in other handlers) the `proof` (since that's what we're building... either from a signing payload alone or from a signature+root)

domains/noto/internal/noto/handler_burn_common.go
```diff
@@ -165,20 +179,29 @@ func (h *burnCommon) baseLedgerInvokeBurn(ctx context.Context, tx *types.ParsedT
...
 			Signature: sender.Payload,
 			Data:      data,
 		})
 	} else {
 		paramsJSON, err = json.Marshal(&NotoTransferParams{
 			TxId:    req.Transaction.TransactionId,
-			Inputs:  endorsableStateIDs(req.InputStates),
-			Outputs: endorsableStateIDs(req.OutputStates),
-			Proof:   sender.Payload,
+			Inputs:  endorsableStateIDs(req.InputStates, tx.DomainConfig.UsesNullifiers()),
```
awrichar: Feels inconsistent that in some places we rely on the passed-in param `useNullifier` while in others we look back at the domain config.

awrichar: If we have the domain config anyway, I don't think we need to require the caller to pass in `useNullifier`.

domains/noto/internal/noto/noto.go
```diff
@@ -515,10 +520,16 @@ func (n *Noto) ConfigureDomain(ctx context.Context, req *prototk.ConfigureDomain
 	n.chainID = req.ChainId
 	n.fixedSigningIdentity = req.FixedSigningIdentity
 
+	algoName := types.AlgoDomainNullifier(n.name)
+	signingAlgos := map[string]int32{
+		algoName: 32,
```
awrichar: What is the significance of this number?

domains/noto/internal/noto/noto.go
```diff
@@ -1101,7 +1117,7 @@ func (n *Noto) encodeTransactionDataV0(ctx context.Context, transaction *prototk
 	return data, nil
 }
 
-func (n *Noto) encodeTransactionDataV1(ctx context.Context, infoStates []*prototk.EndorsableState) (pldtypes.HexBytes, error) {
+func (n *Noto) encodeTransactionDataV1(ctx context.Context, infoStates []*prototk.EndorsableState, useNullifiers bool) (pldtypes.HexBytes, error) {
```
awrichar: Is the `useNullifiers` param used?

domains/noto/internal/noto/noto.go
```diff
@@ -1207,11 +1223,35 @@ func mapPrepareTransactionType(transactionType pldapi.TransactionType) prototk.P
...
+		err := json.Unmarshal(req.Payload, &coin)
+		log.L(ctx).Debugf("unmarshaled coin: %+v\n", coin)
+		if err == nil {
+			hashBytes, err = calculateNullifier(coin)
+		}
+		if err != nil {
+			return nil, i18n.WrapError(ctx, err, msgs.MsgNullifierGenerationFailed)
+		}
+		log.L(ctx).Infof("generated nullifier (string): %s\n", hashBytes.String())
+		log.L(ctx).Infof("generated nullifier (bytes): %x\n", hashBytes.Bytes())
```
awrichar: Should this be info or debug?

domains/noto/internal/noto/states.go
```diff
@@ -579,10 +585,24 @@ func encodedStateIDs(states []*pldapi.StateEncoded) []string {
...
+		} else {
+			// Use the nullifier as the ID
+			var coin types.NotoCoin
+			var hashBytes *pldtypes.Bytes32
+			err := json.Unmarshal([]byte(state.StateDataJson), &coin)
+			if err == nil {
+				hashBytes, err = calculateNullifier(&coin)
+			}
+			if err != nil {
+				return nil
```
awrichar, line 603: Should this error be surfaced more explicitly? Seems to be thrown away.

domains/noto/internal/noto/states.go
```diff
@@ -737,6 +757,67 @@ func (n *Noto) getAccountBalance(ctx context.Context, stateQueryContext string,
 	return len(states), totalBalance, false, false, nil
 }
 
+func (n *Noto) encodeRootAndSignature(ctx context.Context, txContractAddress, stateQueryContext string, payload []byte) ([]byte, error) {
+	// for nullifier variants, the "signature" parameter includes both the signature and the root
```
awrichar, line 762: I think it's called "proof" now instead of "signature"

domains/noto/internal/noto/handler_lock.go
```diff
@@ -58,25 +58,50 @@ func (h *lockHandler) checkAllowed(ctx context.Context, tx *types.ParsedTransact
...
 }
 
 func (h *lockHandler) Assemble(ctx context.Context, tx *types.ParsedTransaction, req *prototk.AssembleTransactionRequest) (*prototk.AssembleTransactionResponse, error) {
 	params := tx.Params.(*types.LockParams)
 
+	// there are special handling in terms of using nullifiers in transactions involving locks.
+	// when locking assets, the input states will have nullifier specs, but the locked output states
+	// will NOT have nullifier specs. This is because the locked outputs are spent by UTXO ID rather
+	// than nullifiers. This is by design to allow easier tracking of locked outputs on-chain throughout
+	// their short life cycle.
```
awrichar, line 95: I spent some extra time thinking on this - just wanted to note that I agree with the design.

domains/noto/internal/noto/handler_lock.go
```diff
@@ -301,28 +343,29 @@ func (h *lockHandler) baseLedgerInvoke(ctx context.Context, tx *types.ParsedTran
 				Data: data,
 			})
 		}
-	} else {
+	} else if tx.DomainConfig.IsV2() {
```
awrichar, line 346: In any of these handlers where we've removed the `else` (and have only `else if`), we should probably have an `else` that throws an error.

domains/noto/internal/noto/handler_mint.go
```diff
@@ -83,6 +84,30 @@ func (h *mintHandler) Assemble(ctx context.Context, tx *types.ParsedTransaction,
...
+					PayloadType:  types.PAYLOAD_DOMAIN_NOTO_NULLIFIER,
+				},
+			}
+			if toID.identifier != notaryID.identifier {
+				newState.NullifierSpecs = append(newState.NullifierSpecs, &prototk.NullifierSpec{
+					Party:        toID.identifier,
+					Algorithm:    types.AlgoDomainNullifier(h.noto.name),
+					VerifierType: types.VERIFIER_DOMAIN_NOTO_NULLIFIER,
+					PayloadType:  types.PAYLOAD_DOMAIN_NOTO_NULLIFIER,
+				})
```
awrichar, line 113: I'm not sure I understand the generation of two nullifier specs. What does it mean for one state to have two nullifiers?

domains/noto/internal/noto/events.go
```diff
@@ -155,6 +177,19 @@ func (n *Noto) handleV1Event(ctx context.Context, ev *prototk.OnChainEvent, res
 	default:
 		log.L(ctx).Infof("Skipping '%s' event in batch %s", ev.SoliditySignature, req.BatchId)
 	}
+
+	// Handle new states representing new SMT nodes
+	if useNullifier {
+		newStatesForSMT, err := smtForStates.Storage.GetNewStates(ctx)
+		if err != nil {
+			log.L(ctx).Errorf("Failed to get new SMT states for tree %s: %s", smtForStates.Name, err)
+			return nil
```
awrichar, line 186: Is it OK to ignore this error?

solidity/contracts/domains/noto/Noto.sol
jimthematrix: reverted to all being "private"

domains/noto/pkg/types/config.go
```diff
@@ -134,12 +134,31 @@ const (
 	NotaryModeIntHooks pldtypes.HexUint64 = 0x0001
 )
 
-var NotoVariantV2 pldtypes.HexUint64 = 0x0002 // V2 variant (latest)
-var NotoVariantV1 pldtypes.HexUint64 = 0x0001 // V1 variant
-var NotoVariantV0 pldtypes.HexUint64 = 0x0000 // V0 variant
+// NotoVariantV2Nullifiers is reported by NotoNullifiers.sol (nullifier inputs + commitment tree).
+var NotoVariantV2Nullifiers pldtypes.HexUint64 = 0x0003
+
+// NotoVariantV2 is the on-chain variant reported by standard Noto.sol (flat lock ABI, coin state IDs).
```
awrichar: I think the "flat lock" description needs to be removed/changed

testinfra/besu_bootstrap/main.go
jimthematrix: it's equivalent to the default, (EIP-170 limit being 24576). guess removing it would be less confusing.

domains/noto/internal/noto/noto.go
jimthematrix: it's a placeholder as for the "Sign" method we just generate the nullifier and doesn't need the key length. added inline comment.

domains/noto/internal/noto/noto.go
jimthematrix: not used, and cleaned up.

domains/noto/internal/noto/noto.go
jimthematrix: cleaned up both

domains/noto/internal/noto/states.go
jimthematrix: added a log here.

domains/noto/internal/noto/handler_mint.go
jimthematrix: added the following inline comments to this code section to clarify the two nullifier specs for output states that are to be owned by a receiver that's not the notary:

```golang
		// for new output states, while we create them, we add the corresponding nullifier to the new state,
		// which will be persisted in the state DB. This allows us to track which states have been spent,
		// because the spending transactions will include the nullifier IDs, rather than the state IDs, in
		// the receipt.
		for _, newState := range outputStates.states {
			// Here the new output state could be for the minter (notaryID) or the receiver (toID).
			// regardless of the owner, the notary always knows about the nullifier. So we always
			// add the nullifier spec for the notary.
			newState.NullifierSpecs = []*prototk.NullifierSpec{
				{
					Party:        notaryID.identifier,
					Algorithm:    types.AlgoDomainNullifier(h.noto.name),
					VerifierType: types.VERIFIER_DOMAIN_NOTO_NULLIFIER,
					PayloadType:  types.PAYLOAD_DOMAIN_NOTO_NULLIFIER,
				},
			}
			// In addition, Paladin also puts the responsibility to generate the nullifier for the states
			// to be owned by the receiver, on the minter. So if the receiver is not the notary, we add
			// another nullifier spec with the distribution to the receiver.
			if toID.identifier != notaryID.identifier {
				newState.NullifierSpecs = append(newState.NullifierSpecs, &prototk.NullifierSpec{
					Party:        toID.identifier,
					Algorithm:    types.AlgoDomainNullifier(h.noto.name),
					VerifierType: types.VERIFIER_DOMAIN_NOTO_NULLIFIER,
					PayloadType:  types.PAYLOAD_DOMAIN_NOTO_NULLIFIER,
				})
			}
		}
```

domains/noto/internal/noto/events.go
jimthematrix: the result of these errors, which should be super-rare as they are all from parsing hex strings, is that an onchain event will fail to be reflected in the merkle tree in the state db (and thus failed to be marked as "Confirmed"). meaning at this point the offchain merkle tree becomes out of sync with the onchain merkle tree maintained by the NotoNullifier contract. future transactions will not recognize these new states as spendable (won't be selected into a Tx's inputs list). in addition, the offchain tree's root may become unrecognizable to the smart contract, causing transactions to fail.

in summary, I think it's OK to simply log the (super-rare) error situation as errors for diagnosis purposes, and it's not a safety issue that would lead to double-spending.

solidity/contracts/domains/noto/NotoNullifiers.sol
jimthematrix: updated as discussed above:

```golang
// for the variant field in the NotoConfig, we adopt a two-byte convention:
// - the lower 8 bits are for the variant number (V0, V1, V2 etc.)
// - the upper 8 bits are for the variant type (standard, nullifiers etc.)
//   - the variant type is 0 for the standard Noto.sol variant
//   - the variant type is 1 for the nullifiers variant
const (
	// legacy
	NotoVariantV0 pldtypes.HexUint64 = 0x0000
	// INoto_V1-shaped parameters (nested LockParams)
	NotoVariantV1 pldtypes.HexUint64 = 0x0001
	// NotoVariantV2 is for the standard Noto.sol (ILockableCapability ABI).
	NotoVariantV2 pldtypes.HexUint64 = 0x0002

	// NotoVariantV2Nullifiers is for the NotoNullifiers.sol (nullifier inputs + commitment tree).
	NotoVariantV2Nullifiers pldtypes.HexUint64 = 0x0102
)
```

domains/noto/internal/noto/handler_transfer_common.go
```diff
@@ -157,26 +185,37 @@ func (h *transferCommon) baseLedgerInvokeTransfer(ctx context.Context, tx *types
 		return nil, err
 	}
 
+	payload := signature.Payload
```
awrichar: payload -> proof

awrichar (approved):
There are some very minor open comments still, but all major items have been addressed. Leaving approval, and you can decide if any of the remaining small items need fixing.

PR #959, opened 2025-12-09
Noto V0 testing

annamcallister:
Brings the changes from https://github.com/LFDT-Paladin/paladin/pull/958 as a base for improved Noto test coverage plus a modification to make the improved tests run against both V0 and V1 Noto.


domains/noto/internal/noto/receipts.go
```diff
@@ -103,8 +103,12 @@ func (n *Noto) BuildReceipt(ctx context.Context, req *prototk.BuildReceiptReques
 			paramsJSON, err = json.Marshal(receipt.LockInfo.UnlockParams)
 		} else {
 			interfaceABI = n.getInterfaceABI(types.NotoVariantLegacy)
+			unlockTxId := pldtypes.Bytes32UUIDFirst16(uuid.New()).String()
+			if receipt.LockInfo != nil && receipt.LockInfo.UnlockTxId != nil && !receipt.LockInfo.UnlockTxId.IsZero() {
```
annamcallister: This is the only non-test code change that isn't also included in #958 

The tests showed that `receipt.LockInfo.UnlockTxId` is set for V0 Notos with the new code, so we should not be using a random transaction ID here if it will be different to the UnlockTxID we've reported in the lock info.

PR #962, opened 2025-12-09
domainmgr: expose some domain/contract config on RPCs

### LINKED ISSUE #612: New RPC for getting config of a domain

Paladin should support an RPC call to domain manager to:
- look up a domain instance by smart contract address, and get its config

While adding this, it's possible these other RPC queries to domain manager might be useful:
- list the configured domains
- look up a domain by name, with details

### Why is this needed?

Currently applications built against Paladin must "remember" all details about configured domains - such as who is the notary for a Noto domain. But it should be possible to look this up and verify it at any point.

awrichar:
Resolves #612 (unless anyone thinks we need to expose any further info)

### DIFF
--- a/.github/WORKFLOWS.md
+++ b/.github/WORKFLOWS.md
@@ -54,13 +54,13 @@ Once changes are merged into the `main` branch, workflows prepare the project fo
   Detects documentation updates and publishes the latest content to the documentation site.
 
 ## Release Time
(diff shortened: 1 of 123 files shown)


peterbroadhurst (approved):
Looks great. I have moved the base over to `v1-develop` and addressed the changes for that.

Issue #928, opened 2025-12-04
bug: Review public and private TX submission between nodes


matthew1001:
### What happened?

See PR review comments:

 - https://github.com/LFDT-Paladin/paladin/pull/704#discussion_r2586110365
 - https://github.com/LFDT-Paladin/paladin/pull/704#discussion_r2586052338
 - https://github.com/LFDT-Paladin/paladin/pull/704#discussion_r2586184483

This issue has been raised to discuss and capture decisions around distribution public transactions, public submissions, and private transactions between nodes

### What did you expect to happen?

N/A

### How can we reproduce it (as minimally and precisely as possible)?

N/A

### Anything else we need to know?

N/A

### OS version

N/A

PR #963, opened 2025-12-09
Distribute public TX submissions and sequencing activity to transaction originators

matthew1001:
## Overview

This PR addresses task https://github.com/LFDT-Paladin/paladin/issues/928 that has had a reasonable amount of discussion.

That task was raised to decide how Paladin should distribute different information about public transactions and Paladin transaction sequencing, to other nodes, particularly with regards to aiding problem diagnosis and giving originating nodes more detail about the progress of their transactions when they are being coordinated by a different node.

For more detail of the discussions leading up to this PR see the write-up in #928 

### Summary of the changes

1. New `sequencer_activities` DB table
   - Initially only storing 2 activity types - `dispatch` and `chained_dispatch`
2. New `dispatcher` column in `public_txns` table
   - An originator will not attempt to progress public transactions dispatched by another node.
3. Distributed sequencer coordinator sends sequencer activity back to the originator to store in the new table
4. Distributed sequencer coordinator no longer persists a copy of the originator's transaction
5. Update to `ptx_getTransactionFull` to return the following where appropriate:
   - sequencer activity (returned in JSON/RPC calls to the originator)
   - dispatches (returned in JSON/RPC calls to the coordinator)
   - chained transactions (returned in JSON/RPC calls to the coordinator who created the chained transaction)
   - (note this JSON/RPC call already returns public transactions)
6. Update to `ptx_getTransactionReceiptFull` to return the following where appropriate:
   - public transactions
      - Needed to give the coordinator a way to retrieve public transactions that relate to a receipt, since it no longer persists the full transaction itself
   - dispatches
   - chained transactions
7. New data flow diagrams in the documentation clarifying what data is available on which nodes in both a simple and a chained TX scenario:

image

8. I've taken the opportunity to bring overall codebase coverage up from 93% to 95% while working on the PR. Some are unrelated to the sequencer but are hopefully not a distraction when reviewing it.

### User experience as a result of the changes

1. For the node that originated a transaction
   - For PD a user may need to see how many public transactions have been attempted, allowing them to correlate the transaction hashes with on-chain transactions.
      - The existing `ptx_getTransactionFull` already returns public transactions, so distributing them & persisting at the originator already makes them available through JSON/RPC
   - For tracking if a specific transaction is being/has been coordinated, the user needs to see at least dispatch related sequencing activity, which indicates that an attempt has been made to deliver a base ledger transaction
      - `ptx_getTransactionFull` has been extended to return all sequencing activity for a transaction
5. For the node that coordinated the transaction
   - The ability to have visibility of at least some sequencer-related information for a given TX ID, without needing to retain and have access to logs, allows a node administrator to aid in PD of a user's transaction, especially where a transaction is not progressing.
   - Specifically this PR makes the `dispatch` table and the `chained_private_txns` table available via `ptx_getTransactionFull` and `ptx_getTransactionReceiptFull` for a given transaction.
   - Sequencing activity isn't stored at the coordinator, but could be in the future when activity other than dispatch is to be retained.
   - As a convenience (and since the coordinator does not persist the Paladin transaction itself), `ptx_getTransactionReceiptFull` has been extended to return public transaction data as well as the receipt. The "non-full" variant does not return the public transactions.

### Other internals changes

Some parts of the codebase pass DB-variants of public transactions around (e.g. `DBPubTxnSubmission`). Since these structures do not contain all public transaction data, some have been extended with non-gorm fields to allow public TX data to be retrieved in the distributed sequencer without additional DB lookups (see https://github.com/LFDT-Paladin/paladin/pull/963/files#diff-90d922f506dca6c4261a005ddc372fd5dd79457615e6fa70aafe056ed9bf7315)

### DIFF
--- a/core/go/db/migrations/sqlite/000024_create_sequencer_activity_table.up.sql
+++ b/core/go/db/migrations/sqlite/000024_create_sequencer_activity_table.up.sql
@@ -1,9 +1,8 @@
 CREATE TABLE sequencer_activities (
-  "id"                        BIGINT          NOT NULL,
+  "id"                        INTEGER         PRIMARY KEY AUTOINCREMENT,
   "subject_id"                TEXT            NOT NULL,
   "timestamp"                 BIGINT          NOT NULL,
   "transaction_id"            UUID            NOT NULL,
   "activity_type"             TEXT            NOT NULL,
-  "submitting_node"           TEXT            NOT NULL,
-  PRIMARY KEY ("id")
+  "submitting_node"           TEXT            NOT NULL
 );


common/go/pkg/pldmsgs/en_descriptions.go
```diff
@@ -106,6 +106,7 @@ var (
 	PublicTxRevertData                     = pdm("PublicTx.revertData", "The revert data (optional)")
 	PublicTxSubmissions                    = pdm("PublicTx.submissions", "The submission data (optional)")
 	PublicTxActivity                       = pdm("PublicTx.activity", "The transaction activity records (optional)")
+	PublicTxDispatcher                     = pdm("PublicTx.dispatcher", "The dispatcher that submitted this public transaction")
```
peterbroadhurst: Note to come back to what string this is.
Is it the Paladin identifier the submitting node uses to identify the signing key?

core/go/db/migrations/postgres/000024_create_sequencer_activity_table.up.sql
```diff
@@ -0,0 +1,13 @@
+BEGIN;
+
+CREATE TABLE sequencer_activities (
+  "id"                        BIGINT          GENERATED ALWAYS AS IDENTITY,
+  "remote_id"                 TEXT            NOT NULL,
```
peterbroadhurst: Interested to come back to what "remote" means here

core/go/internal/publictxmgr/in_flight_transaction_state_generation.go
```diff
@@ -296,9 +296,43 @@ func (v *inFlightTransactionStateGeneration) PersistTxState(ctx context.Context)
 	}
 
 	if rsc.StageOutputsToBePersisted.TxUpdates != nil {
-
 		newSubmission := rsc.StageOutputsToBePersisted.TxUpdates.NewValues.NewSubmission
 		if newSubmission != nil {
+			log.L(ctx).Debugf("PersistTxState TXID %s: Queueing new submission", newSubmission.PrivateTXID)
+
+			if newSubmission.Binding == nil && rsc.InMemoryTx != nil {
```
peterbroadhurst: A comment here would be helpful for what `Binding` and `InMemoryTx` mean.

I know there's a lot of historical complexity in this package, so it's pretty complex to describe why this particular function is using a piece of memory written by another, and why it's safe to do this check+set processing.... So it might be complex to explain.... but given you have it paged in right now, it's likely to help future us to page back in again.

core/go/internal/sequencer/transport_client.go
```diff
@@ -480,76 +477,6 @@ func (sMgr *sequencerManager) handleEndorsementRequest(ctx context.Context, mess
 		return
 	}
 
-	// If this TX ID doesn't exist in the "transactions" DB, insert here.
```
peterbroadhurst: Great to see this block going 👍

core/go/internal/txmgr/persisted_receipt.go
```diff
@@ -412,7 +413,50 @@ func (tm *txManager) buildFullReceipt(ctx context.Context, receipt *pldapi.Trans
 			}
 		}
 	}
-	return fullReceipt, nil
+
+	fullReceipt, err = tm.mergeDispatches(ctx, tm.p.NOTX(), []uuid.UUID{fullReceipt.ID}, []*pldapi.TransactionReceiptFull{fullReceipt})
```
peterbroadhurst: Think you should change now with all these calls nested, to the DB TXID being a parameter, then push it in from above - rather than lots of individual `tm.p.NOTX()` calls.

Then later if we decide we need to include this in a DB TX, there's a lot less risk of missing one of the sub-reads.

core/go/internal/txmgr/transaction_submission.go
```diff
@@ -370,61 +370,6 @@ func (tm *txManager) PrepareChainedPrivateTransaction(ctx context.Context, dbTX
 	return chained, err
 }
 
-func (tm *txManager) InsertRemoteTransaction(ctx context.Context, dbTX persistence.DBTX, txi *components.ValidatedTransaction, ignoreConflicts bool) (int64, error) {
```
peterbroadhurst: 🎉  again great to see this go

sdk/go/pkg/pldapi/transaction.go
```diff
@@ -131,10 +131,12 @@ type TransactionHistory struct {
...
-	History   []*TransactionHistory   `docstruct:"TransactionFull" json:"history,omitempty"`   // list of values previously provided for this transaction
+	DependsOn         []uuid.UUID             `docstruct:"TransactionFull" json:"dependsOn,omitempty"`         // transactions registered as dependencies when the transaction was created
+	Receipt           *TransactionReceiptData `docstruct:"TransactionFull" json:"receipt"`                     // available if the transaction has reached a final state
+	Public            []*PublicTx             `docstruct:"TransactionFull" json:"public"`                      // list of public transactions associated
+	History           []*TransactionHistory   `docstruct:"TransactionFull" json:"history,omitempty"`           // list of values previously provided for this transaction
+	SequencerActivity []*SequencerActivity    `docstruct:"TransactionFull" json:"sequencerActivity,omitempty"` // list of sequencing activity for this transactions
+	Dispatches        []*Dispatch             `docstruct:"TransactionFull" json:"dispatches,omitempty"`        // list of dispatches for this transaction
```
peterbroadhurst: Can you clarify the difference in purpose and relationship between the `sequencerActivity` array and the `dispatches` array?

I think we need to strongly justify to ourselves every time the list of top-level constructs for a user grows, and validate we aren't introducing more than we need to (as it's already quite a high data and cognitive load to receive/read this payload.

peterbroadhurst (changes requested):
This PR is great 🙌 

I've gone through in detail, and you'll find a number of comments which are very much in the weeds and suggestions in nature.

I did mark it as requested changes, as mixed in there are a couple of important items I think we harness this point in time for the PR merge, to validate we confident of:
- Consistently adding/naming of fields
- Ensuring the origination information of data is set (or checked) based on the transport
- In-memory data locking in public Tx manager

PR #964, opened 2025-12-10
Upgradeable domain factories and v2 noto factory

annamcallister:
* Add a new V2 noto factory which fixes a bug with the `deployImplementation` function. It now has an `implementationName` parameter instead of using the Noto name as the implementation name
* Add new versions of all the domain factories which are suitable to be deployed behind ERC1967 proxies. Update tests and the quick start deployment to deploy both the factories and and their proxies.
  * `NotoFactory` became too large to be deployed after making it upgradeable, so the default Noto implementation now needs to be deployed separately to the factory.


doc-site/docs/getting-started/installation-advanced.md
```diff
@@ -174,37 +174,37 @@ If you wish to use an existing seed phrase, store a file containing this seed ph
...
-```
-NAME                                  STATUS      DOMAIN_REGISTRY                              DEPLOYMENT      LIBRARY
-paladindomain.core.paladin.io/noto    Available   0xd93630936d854fb718b89537cce4acc97fd50463   noto-factory    /app/domains/libnoto.so
-paladindomain.core.paladin.io/pente   Available   0x48c11bbb7caa77329d53b89235fec64733a24ca1   pente-factory   /app/domains/pente.jar
-paladindomain.core.paladin.io/zeto    Available   0xc29ed8a902ff787445bdabee9ae5e2380089959d   zeto-factory    /app/domains/libzeto.so
+```bash
+NAME                                  STATUS      DOMAIN_REGISTRY                              DEPLOYMENT            LIBRARY
+paladindomain.core.paladin.io/noto    Available   0xa1de12834d636299d28d974a2742f203daa1c131   noto-factory-proxy    /app/domains/libnoto.so
+paladindomain.core.paladin.io/pente   Available   0xa146d815872f0b6f83e0b608d07b70102860764b   pente-factory-proxy   /app/domains/pente.jar
+paladindomain.core.paladin.io/zeto    Available   0x270b25f87da05af06fc99eebdc39dffc29cdf0e1   zeto-factory          /app/domains/libzeto.so
```
awrichar: I know this is just an example, but why is zeto different here?

doc-site/docs/getting-started/installation-advanced.md
```diff
@@ -225,11 +225,11 @@ Here is a full example of the `values-attach.yaml`:
 ```yaml
 mode: attach
 smartContractsReferences:
-  notoFactory:
+  notoFactorProxy:
```
awrichar:
```suggestion
  notoFactoryProxy:
```

solidity/contracts/domains/noto/NotoFactory.sol
```diff
@@ -1,18 +1,53 @@
...
+    IPaladinContractRegistry_V0
+{
+    /// @custom:storage-location erc7201:paladin.storage.NotoFactory
+    struct NotoFactoryStorage {
+        mapping(string => address) implementations;
+    }
+
+    // keccak256(abi.encode(uint256(keccak256("paladin.storage.NotoFactory")) - 1)) & ~bytes32(uint256(0xff))
+    bytes32 private constant NOTO_FACTORY_STORAGE_LOCATION =
+        0xe67f96c01195ce35b5ff7e8a6398509283372467daaa337070e1893bae2d8600;
```
awrichar: In theory we don't need to namespace the storage, if this is a concrete class that we don't expect others to extend (I think namespacing is largely needed to decouple storage layout from inheritance order - so it's less of an issue in child/leaf contracts). But I guess there's no harm in doing it anyway.

solidity/contracts/domains/noto/NotoFactory.sol
```diff
@@ -44,20 +81,23 @@ contract NotoFactory is Ownable, IPaladinContractRegistry_V0 {
...
         bytes32 transactionId,
+        string calldata implementationName,
         string calldata name,
         string calldata symbol,
         address notary,
         bytes calldata data
     ) external {
-        _deploy(implementations[name], transactionId, name, symbol, notary, data);
+        NotoFactoryStorage storage $ = _getNotoFactoryStorage();
+        _deploy($.implementations[implementationName], transactionId, name, symbol, notary, data);
```
awrichar: Should we sanity-check that the implementation is defined?

awrichar (approved):
Few minor things noted inline, but this looks good

solidity/contracts/domains/noto/NotoFactory.sol
annamcallister: Removed the namespacing

solidity/contracts/domains/noto/NotoFactory.sol
annamcallister: Added a revert if the implementation doesn't exist

PR #967, opened 2025-12-11
noto: add tests for delegate lock

awrichar:
This includes a test demonstrating the current incorrect behavior when trying to delegate a lock created with "createMintLock".


domains/noto/internal/noto/handler_delegate_lock_test.go
```diff
@@ -0,0 +1,499 @@
...
+			Verifier:     senderKey.Address.String(),
+		},
+	}
+
+	assembleRes, err := n.AssembleTransaction(ctx, &prototk.AssembleTransactionRequest{
+		Transaction:       tx,
+		ResolvedVerifiers: verifiers,
+	})
+
+	// Currently expecting REVERT result (this is wrong!)
```
awrichar: Note that the test reverts here - but we need to find a way to allow this pattern (delegating a lock that has no input states).

PR #973, opened 2025-12-16
add print/clean up disk space steps to workflows

annamcallister:
Some of our workflows have been using most of the available disk space on the `ubuntu-latest` GHA runner, meaning that the recent increase in disk usage by preinstalled tools on the runner has been causing workflows to fail with no disk space left on device errors.

This PR adds a clean up disk space action to any work flow that runs gradle `build`, `test` or `e2e` tasks . Not all of these have been failing, but I expect some of these are getting close to the limit so it feels safer to get out ahead of this.

On some other workflows which include downloads I have included just a print disk usage step at the start and the end. These are workflows that I do not think are at risk of running out of space with small increases in preinstalled runner tools, so do not justify the extra time it takes to clear disk space. However, it costs very little to do the print, and we might appreciate this data in future.

There is a separate conversation about could we be reducing the amount of disk space we need per workflow, but since we haven't got worse in this regard, this PR allows us to continue to have at least as much disk space as we have had previously, and prints out disk usage in a way that we can monitor consumption over time.

PR also includes commit to work around https://github.com/helm-unittest/helm-unittest/issues/790.

### DIFF
--- a/operator/Makefile
+++ b/operator/Makefile
@@ -201,7 +201,7 @@ helm-template: ## Install operator using Helm.
 # install helm unittest
 .PHONY: helm-unittest-install
 helm-unittest-install: ## Install helm unittest.
-	$(HELM) plugin install https://github.com/helm-unittest/helm-unittest.git || true
+	$(HELM) plugin install https://github.com/helm-unittest/helm-unittest.git --version 1.0.3 || true
 
 .PHONY: helm-unittest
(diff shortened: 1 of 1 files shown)


matthew1001 (approved):
Changes look good to me @annamcallister 

We ought to make a note to come back and un-pin the `helm-unittest` version at some point

PR #978, opened 2025-12-17
Noto V1 smart contract refactor and new domain API features

peterbroadhurst:
### Noto refactor summary

This PR introduces a significant refactor of Noto as part of its V1 upgrade, which is included in the Paladin V1 release stream.

V1 state machine

The on-chain smart contract interface and implementation of locking have been revised to be simpler, and more extensible, centered around an `ILockableCapability` interface that is implementation agnostic.

This clean separation between a generic `ILockableCapability` interface, and a `INoto` implementation of this interface paves the way to:

1. Standardize Noto, Zeto and ERC20 derived lockable coins to a common on-chain interface
2. Simplify the Atom smart contract to this interface
3. Promote wider interoperability of confidential tokens in the Ethereum space

A comparison with the Noto V0 on-chain locking system is provided in the documentation part of this PR:

https://github.com/LFDT-Paladin/paladin/blob/d5a611f5b99ca318bb1d7447e565901abf76b73f/doc-site/docs/architecture/noto.md#locking-and-atomic-settlement

### `ILockableCapability` high-level design

The `ILockableCapability` interface is deliberately very generic and extensible in nature. It does not imply anything about the implementation of the underlying smart contract, meaning it is intended to be used for:

- Transparent tokens
    - Compliant with ERC-20, ERC-721, ERC-1155 etc.
    - With support for locking (no one standard exists for this)
- Confidential tokens using any proof system
    - Client-originated ZKPs (per Zeto)
    - Server-originated notary certificates (per Noto)
    - Server-originated MPC encrypted inputs (per common use of FHE)

To achieve this level of flexibility the interface designates different data payloads, with unique architecturally significant uses, but must use `bytes` as the encoding for the implementation payloads.

- Input data for the operations
   - `createLock`/`updateLock`/`delegateLock`/`spendLock`/`cancelLock`
- Data stored on the lock on-chain
- Lock IDs (fixed 32 bytes)
- Pass-through data to events
   - Should be transparent to on-chain - consistent with ERC-5750

There are then a set of generic events that provide lifecycle information on the locking system in a consistent way across all implementations:

- `LockUpdated` / `LockDelegated` / `LockSpent` / `LockCancelled`

These events cannot assume the whole input data should be emitted in the event (for a system like a ZKP system this would be a large data payload), and the interface encourages separation of concerns with ERC-5750 data. As such implementation (including Noto) should define additional events containing their own implementation-specific fields.

Please see the comments in the interface for more detail:

https://github.com/LFDT-Paladin/paladin/blob/d5a611f5b99ca318bb1d7447e565901abf76b73f/solidity/contracts/domains/interfaces/ILockableCapability.sol#L36

### New Noto API features

This PR brings some important new features to the Noto domain private smart contract interface available to applications through the Paladin API.

https://github.com/LFDT-Paladin/paladin/blob/d5a611f5b99ca318bb1d7447e565901abf76b73f/solidity/contracts/domains/interfaces/INotoPrivate.sol

- `prepareMintUnlock` / `prepareBurnUnlock`
    - These allow mint and burn operations to be prepared for an existing lock, operating similarly to `prepareUnlock`
- `createTransferLock` / `createMintLock` /  `createBurnLock`
    - These function combine the `lock` and `prepareXYZUnlock` functions together to a single on-chain transaction
    - This avoids the latency of two blockchain operations, and simplifies application logic

### New hooks

Corresponding hooks have been added for the new operations.

https://github.com/LFDT-Paladin/paladin/blob/d5a611f5b99ca318bb1d7447e565901abf76b73f/solidity/contracts/domains/interfaces/INotoHooks.sol

- `onPrepareMintUnlock` / `onPrepareBurnUnlock`
- `onPrepareTransferLock` / `onPrepareMintLock` / `onPrepareBurnLock`

### New lock state

The biggest architectural change to the on/off-chain integration of Noto in this PR, is the introduction of a new on-chain tracked input/output state for the lock separate to the locked-coins.

This was critical for `mint` locks, because there is no locked coins in the lock to track the ownership. So without this new state, it would be impossible for the notary to validate ownership of an existing lock in order to "spend" the lock (and perform the mint).

The introduction of this new state has a number of other benefits though as well. Particularly in ensuring the on/off-chain logic is tracking a single unspent lock state at any point in time. This consistency in the way both lock and coin states are managed in Noto V1 reduces the likelihood of the Notary being convinced by applications to submit transactions that require the on-chain source of truth to intercept and reject the transaction (as these scenarios drive off-chain unwinding of state).

### Cancel operations

In Noto V1 each lock now as three exit options after delegation has been performed, where the 3rd is new:

1. `delegate` the lock - back to the owner where it can be unlocked with any valid output (or to another smart contract)
2. "spend" the lock - by performing _exactly_ the coin operations prepared in the lock
3. "cancel" the lock - returning a pre-prepared set of coins back to the original owner

In both 2 & 3 the on-chain delegate smart contract must pass the exact parameters matching the `spendHash`/`cancelHash` stored on the lock.

The Noto domain receipt `lockInfo` structure now contains `cancelFunction`/`cancelParams`/`cancelCall`, in addition to the existing `unlockFunction`/`unlockParams`/`unlockCall`.

### Spend / cancel transaction ID - `spendTxId`

A part of Noto V1 that improves on the locking system in V0 is the ability to deterministically know the transaction ID that will confirm the prepared lock.  This is important for correlation in the app layer.

This `spendTxId` is tracked in the lock state on and off-chain, and is established as unique at the time the lock is prepared (if a lock were to be prepared multiple times, the transaction ID would be new each time).

The same ID is used regardless of if the `spend` or `cancel` completion paths are taken for the lock.

### Spend / cancel transaction data

The data of the transfer/mint/burn operation is part of the spend/cancel hash. As with all Noto transactions, this data on-chain is an ABI encoded structure containing a list of info transaction states. This provides the intention that correlation data can (in addition to the `spendTxId`) can be supplied in the application to be emitted as part of the event when the final mint/burn/transfer transaction completes.

> ~TODO: There is no way at the end of this PR to specify non-empty data on the transaction. If/when this is introduced, it will require a new `info` state to be emitted during the prepare, and the has of this needs to be included in the `txData` that is used in the spend/cancel hash.~ Addressed as part of review

New `unlockData` optional parameter is added to `prepareUnlock` to allow the data for the prepared unlock to be provided, and this is be prepared into the `spendHash`/`cancelHash`.

Also after discussion with @awrichar the distribution of both the data for `prepareUnlock` and `unlock` is distributed to all token recipients in the manifest.

### Upgrade / migration considerations

Compatibility with Noto V0 domains is preserved and regression testing of V0 compatibility enhanced in this PR. However, there is no direct upgrade path for on-chain smart contracts provided in this release. Value should be migrated to a new smart contract instance.

Noto V1 includes features to facilitate in-place migration for future releases of the domain after V1.

> There is no migration path for Noto tokens deployed from earlier revisions on the `v1-develop` pre-release branch.

The external API for the domain on Paladin has remained backwards compatible. A number of new features are only available on the V1 release (described below), but no change should be required to applications to begin using Noto V1.

If your environment was created using the getting started guide and the simple supplied Kubernetes operator functions for deploying the Noto factory smart contracts, you are recommended to **delete+recreate your environment** to begin using the V1 smart contract factory.

If you have a more sophisticated long-lived Paladin ecosystem you will need to deploy a Noto V1 factory, and configure that factory as a separate domain on all nodes to begin using Noto V1.

> If you have a mature Paladin ecosystem you are recommended to talk to the maintainers.

### Noto V1 history

This PR is a larger refactor that builds on learnings from a number streams of work:

- Supersedes:
    - #813 
    - #967 
- Builds on:
    - #918
    - #922
    - #923
    - #925 - chained for independent review


domains/noto/internal/noto/handler_prepare_unlock.go
```diff
@@ -130,49 +238,48 @@ func (h *prepareUnlockHandler) baseLedgerInvoke(ctx context.Context, tx *types.P
...
+		lockParams, err = h.buildPrepareUnlockParams(ctx, tx, lockTransition, sender.Payload, lockedInputs, spendOutputs, cancelOutputs, req.InfoStates)
+		if err == nil {
+			interfaceABI = h.noto.getInterfaceABI(types.NotoVariantDefault)
+			functionName = "updateLock"
+			params := lockParams
+			paramsJSON, err = json.Marshal(params)
+		}
+	default:
+		var unlockHash ethtypes.HexBytes0xPrefix
+		unlockHash, err = h.noto.unlockHashFromIDs_V0(ctx, tx.ContractAddress, endorsableStateIDs(lockedInputs), endorsableStateIDs(lockedOutputs), endorsableStateIDs(spendOutputs), inParams.Data)
```
awrichar: This isn't totally related to the PR - but I was reminded again today that it's weird how we pass in a "data" parameter to prepareUnlock and it gets used in two ways: 
- it is encoded into an "info" state with the prepareUnlock transaction
- it is also included in the computation of the unlockHash, meaning that the same piece of data has to be passed later when you unlock

These could/should actually be two separate pieces of "data". I can't tell if that's changed in the new landscape here, but it seems like another thing we might want to roll in with the other changes.

solidity/contracts/domains/interfaces/ILockableCapability.sol
```diff
@@ -0,0 +1,300 @@
...
+     * Emits a {LockCreated} event.
+     */
+    function createLock(
+        bytes calldata createInputs,
+        LockParams calldata params,
+        bytes calldata data
+    ) external returns (bytes32 lockId);
+
+    /**
+     * @dev Update the current options for a lock (non-normative method aligned with ILockableCapability recommendations).
```
awrichar: I think this was moved from a base-class up to this interface - now it is prescribed by ILockableCapability, so this comment can be adjusted.

solidity/contracts/domains/interfaces/ILockableCapability.sol
```diff
@@ -0,0 +1,300 @@
...
+     *
+     * @param lockId The identifier of the lock.
+     * @return info  The information about the lock.
+     */
+    function getLock(
+        bytes32 lockId
+    ) external view returns (LockInfo memory info);
+
+    // ------------------------------------------------------------------------
+    // Non-normative examples
```
awrichar: Are these still useful, or did we align on a standardized set of methods now?

solidity/contracts/domains/interfaces/INoto.sol
```diff
@@ -1,67 +1,116 @@
...
-        bytes signature,
+        bytes32[] lockedOutputs,
+        bytes proof,
         bytes data
     );
 
-    event NotoLock(
-        bytes32 txId,
-        bytes32 lockId,
+    // The Noto event for creation of a lock contains the inputs,outputs an lockedOutputs
```
awrichar: typo: and

solidity/contracts/domains/noto/Noto.sol
```diff
@@ -286,219 +314,305 @@ contract Noto is EIP712Upgradeable, UUPSUpgradeable, INoto, INotoErrors {
...
-        LockInfo storage lockInfo = _locks[lockId];
+        bytes calldata updateInputs,
+        LockParams calldata params,
+        bytes calldata data
+    ) external virtual override onlyNotary {
+        NotoLockOperation memory lockOp = abi.decode(updateInputs, (NotoLockOperation));
+        if (lockOp.lockedOutputs.length > 0) {
+            revert NotoLockedOutputsInUpdate(lockId);            
+        }
+        _updateLock(lockOp, params, lockId, data);
```
awrichar: Shouldn't we ensure the lock exists?

solidity/contracts/domains/noto/Noto.sol
```diff
@@ -286,219 +314,305 @@ contract Noto is EIP712Upgradeable, UUPSUpgradeable, INoto, INotoErrors {
...
-        bytes calldata signature,
+        bytes calldata spendInputs,
         bytes calldata data
-    ) external virtual override onlyNotary txIdNotUsed(txId) {
-        LockInfo storage lockInfo = _locks[lockId];
-        if (lockInfo.delegate != address(0)) {
-            revert NotoAlreadyPrepared(unlockHash);
+    ) external override lockActive(lockId) onlySpender(lockId) {
+        LockInfo storage lock = _locks[lockId];
+        NotoUnlockOperation memory unlockOp = abi.decode(spendInputs, (NotoUnlockOperation));
```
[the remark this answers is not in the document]
awrichar: I'm also trying to work out why we have to pass the states in again and check them against what is stored. Also, wouldn't we be spending the lock state itself, in addition to the locked coin states?

solidity/contracts/domains/interfaces/ILockableCapability.sol
```diff
@@ -0,0 +1,300 @@
...
+     */
+    event LockDelegated(
+        bytes32 indexed lockId,
+        address indexed from,
+        address indexed to,
+        bytes data
+    );
+
+    // Fields of the lock that can be set on create or update
+    struct LockParams {
```
awrichar: Can we follow the Solidity style guide for ordering? structs/types -> vars (n/a for interfaces) -> events -> errors -> modifiers -> functions

solidity/contracts/domains/interfaces/INoto.sol
```diff
@@ -1,67 +1,116 @@
...
+        bytes proof; // recorded signature for the lock operation
+    }
+
+    // The structure definition for a Noto unlock operation, which can be hashed
+    // in order to construct a spendHash or a cancelHash
+    struct NotoUnlockOperation {
+        bytes32 txId;
+        bytes32[] inputs;
+        bytes32[] outputs;
+        bytes data; // this is the inner-data of the prepared transaction (not the unlock)
```
awrichar: Isn't the prepared transaction an unlock?

solidity/contracts/domains/noto/Noto.sol
```diff
@@ -286,219 +314,305 @@ contract Noto is EIP712Upgradeable, UUPSUpgradeable, INoto, INotoErrors {
...
+        lock.spendHash = params.spendHash;
+        lock.cancelHash = params.cancelHash;
+
+        if (params.options.length > 0) {
+            _setLockOptions(lockId, params.options);
+        }
+
+        emit LockUpdated(
             lockId,
-            msg.sender,
```
awrichar: Should we consider keeping the "sender" in the base event signature(s)?

domains/noto/internal/noto/noto.go
```diff
@@ -1170,14 +1288,55 @@ func (n *Noto) computeLockId(ctx context.Context, contractAddress *pldtypes.EthA
...
+	if state.SchemaId == n.lockInfoSchemaV0.Id {
+		lock, err := n.unmarshalLockV0(lockStates[0].StateDataJson)
 		if err != nil {
-			return nil, err
+			return nil, nil, nil, err
 		}
-		return lock, nil
+		return &lock.LockID, nil /* no spendTxId in V0 lock info */, lock.Delegate, nil
+	}
+	lock, err := n.unmarshalLockV1(lockStates[0].StateDataJson)
```
awrichar: Do we ever end up in this branch? That is, we're inside `extractLockInfoV0` but we find V1 lock states?

domains/noto/internal/noto/noto.go
```diff
@@ -139,37 +152,189 @@ type NotoDeployParams struct {
...
+	Data     pldtypes.HexBytes    `json:"data"`
+}
+
+type LockStates struct {
+	Inputs        []pldtypes.Bytes32 `json:"inputs"`
+	Outputs       []pldtypes.Bytes32 `json:"outputs"`
+	LockedOutputs []pldtypes.Bytes32 `json:"lockedOutputs"`
+}
+
+// INoto.NotoLockUpdated event JSON schema - describes the UTXO transaction that accompanies a lock create
```
awrichar: should be NotoLockCreated

domains/noto/internal/noto/noto.go
```diff
@@ -139,37 +152,189 @@ type NotoDeployParams struct {
...
+	LockID       pldtypes.Bytes32     `json:"lockId"`
+	Operator     *pldtypes.EthAddress `json:"operator"`
+	LockContents []pldtypes.Bytes32   `json:"lockContents"`
+	Inputs       []pldtypes.Bytes32   `json:"inputs"`
+	Outputs      []pldtypes.Bytes32   `json:"outputs"`
+	Proof        pldtypes.HexBytes    `json:"proof"`
+	Data         pldtypes.HexBytes    `json:"data"`
+}
+
+// INoto.LockDelegated event JSON schema
```
awrichar: should be NotoLockDelegated

solidity/contracts/domains/interfaces/ILockableCapability.sol
peterbroadhurst: Using https://docs.soliditylang.org/en/latest/style-guide.html#order-of-layout to work through the file

solidity/contracts/domains/interfaces/INoto.sol
peterbroadhurst: Changed comment to:

```go
bytes data; // this is the inner-data of the prepared transaction (not the data from the unlock function call)
```

domains/noto/internal/noto/handler_delegate_lock.go
```diff
@@ -75,8 +74,17 @@ func (h *delegateLockHandler) Assemble(ctx context.Context, tx *types.ParsedTran
...
+	// Load the existing lock
+	var existingLock *loadedLockInfo
+	if !tx.DomainConfig.IsV0() {
+		existingLock, err = h.noto.loadLockInfoV1(ctx, req.StateQueryContext, params.LockID)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Requester must own at least one locked coin state to show ownership of the lock
```
awrichar: This is no longer the case in V1, right?

domains/noto/internal/noto/handler_lock.go
```diff
@@ -203,11 +229,22 @@ func (h *lockHandler) Endorse(ctx context.Context, tx *types.ParsedTransaction,
 		return nil, err
 	}
 
+	if !tx.DomainConfig.IsV0() {
+		lockID, err := h.noto.computeLockIDForLockTX(ctx, tx, notaryID)
+		if err == nil {
+			// TODO: Support preparation of target operation directly during lock (avoiding extra call)
```
awrichar: Is this now provided in the form of createTransferLock etc?

domains/noto/internal/noto/handler_unlock.go
```diff
@@ -97,79 +98,98 @@ func (h *unlockCommon) init(ctx context.Context, tx *types.ParsedTransaction, pa
 	}, nil
 }
 
-func (h *unlockCommon) assembleStates(ctx context.Context, tx *types.ParsedTransaction, params *types.UnlockParams, unlockTxId *pldtypes.Bytes32, req *prototk.AssembleTransactionRequest) (*prototk.AssembleTransactionResponse, *unlockStates, error) {
+func (h *unlockCommon) assembleStates(ctx context.Context, tx *types.ParsedTransaction, spendTxId *pldtypes.Bytes32, params *types.UnlockParams, req *prototk.AssembleTransactionRequest) (*prototk.AssembleTransactionResponse, *manifestBuilder, *unlockStates, error) {
```
awrichar: Should we move the "common" logic to a separate file?

solidity/contracts/domains/noto/Noto.sol
peterbroadhurst: I added a `lockActive(lockId)` guard

solidity/contracts/domains/noto/Noto.sol
```diff
@@ -286,219 +314,305 @@ contract Noto is EIP712Upgradeable, UUPSUpgradeable, INoto, INotoErrors {
...
-            params.signature,
-            params.data
+            lock,
+            data
         );
     }
 
-    function _validateUnlock(
+    function _setLockOptions(
         bytes32 lockId,
```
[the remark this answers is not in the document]
peterbroadhurst: Updated this in a few places

solidity/contracts/domains/noto/Noto.sol
```diff
@@ -286,219 +314,305 @@ contract Noto is EIP712Upgradeable, UUPSUpgradeable, INoto, INotoErrors {
...
+     * Emits a {LockUpdated} event.
      */
-    function unlock(
-        bytes32 txId,
+    function updateLock(
         bytes32 lockId,
-        UnlockParams memory params
-    ) external virtual override txIdNotUsed(txId) {
-        LockInfo storage lockInfo = _locks[lockId];
+        bytes calldata updateInputs,
```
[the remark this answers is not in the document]
peterbroadhurst: > I do wonder if separate structs for create/update (instead of reusing NotoLockOperation) might be slightly cleaner. It feels odd to reuse this struct but require that you must not set one of the fields in it.

Implemented

solidity/contracts/domains/noto/Noto.sol
```diff
@@ -286,219 +314,305 @@ contract Noto is EIP712Upgradeable, UUPSUpgradeable, INoto, INotoErrors {
...
-        lockInfo.delegate = delegate;
-        bytes32 unlockHash = lockInfo.unlockHash;
+
+        useTxId(delegateOp.txId);
+
+        address previousSpender = lock.spender;
+        lock.spender = newSpender;
+
+        _processInputs(delegateOp.inputs);
+        _processOutputs(delegateOp.outputs);
```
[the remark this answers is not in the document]
peterbroadhurst: The `inputs`/`outputs` of Noto operations are states that are spent/created as part of this operation. The lock state for example.

domains/noto/internal/noto/handler_lock.go
awrichar: Is the "decode" below just a sanity check?

domains/noto/internal/noto/handler_prepare_unlock.go
peterbroadhurst: Added in https://github.com/LFDT-Paladin/paladin/pull/978/changes/c9b01986ab503db41b91c48f8be3b2e1e4065d83, and updated the header comment to reflect this.

Also after our discussion @awrichar I updated the distribution of the state data in that commit to include all token recipients, as well as the sender/notary.

peterbroadhurst:
Fixes #1041

domains/noto/internal/noto/noto.go
```diff
@@ -710,24 +808,13 @@ func validateTransactionCommon[T any](
...
-		if domainConfig.IsV0() {
-			abi = types.NotoV0ABI.Functions()[functionABI.Name]
-		} else {
-			abi = types.NotoABI.Functions()[functionABI.Name]
-		}
-	} else {
-		abi = types.NotoABI.Functions()[functionABI.Name]
-	}
+	// If there is an exact signature match, use that directly
+	abi := types.NotoABIFunctionsBySignature[tx.FunctionSignature]
```
peterbroadhurst: I made this change to support polymorphism more generally (we have two versions of `prepareUnlock` now, as well as two versions of `delegateLock`).

On face value this looks like a change of behavior, because you could just match on the `name` before... but if you look below there's a `MsgUnexpectedFunctionSignature` in the case `tx.FunctionSignature != signature`

solidity/contracts/domains/interfaces/INotoPrivate.sol
```diff
@@ -44,25 +52,43 @@ interface INotoPrivate {
 
     function createLock(uint256 amount, bytes calldata data) external;
 
-    function createMintLock(
+    function unlock(
+        bytes32 lockId,
+        string calldata from,
         UnlockRecipient[] calldata recipients,
```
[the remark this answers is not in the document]
peterbroadhurst: Propose we postpone batch support in regular transfers to a different PR, which doesn't necessarily need coupling to the V0->V1 breaking migration, as a batch transfer function could be additive

domains/integration-test/noto_test.go
```diff
@@ -599,25 +615,51 @@ func (s *notoTestSuite) TestNotoCreateMintLock() {
...
+	lockedCoins := findAvailableCoins[types.NotoLockedCoinState](t, ctx, rpc, notoDomain.Name(), notoDomain.LockedCoinSchemaID(), "pstate_queryContractStates", noto.Address, nil)
+	require.Len(t, lockedCoins, 0)
+
+	balanceOfResult := noto.BalanceOf(ctx, &types.BalanceOfParam{Account: recipient1Name}).SignAndCall(notaryName).Wait()
+	assert.Equal(t, "0", balanceOfResult["totalBalance"].(string), "Balance of recipient1 should be 0")
+
+	log.L(ctx).Infof("Prepare mint unlock for the locked 50")
+	rpcerr = rpc.CallRPC(ctx, &invokeResult, "testbed_invoke", &pldapi.TransactionInput{
+		TransactionBase: pldapi.TransactionBase{
+			From:     recipient1Name,
```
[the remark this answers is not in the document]
peterbroadhurst: ... but this test was left disabled - I'd missed that - thank you!

domains/integration-test/noto_test.go
```diff
@@ -628,82 +670,54 @@ func (s *notoTestSuite) TestNotoCreateMintLock() {
 	}, true)
 	require.NoError(t, rpcerr)
 
-	var mintLockReceipt types.NotoDomainReceipt
-	err := json.Unmarshal(invokeResult.DomainReceipt, &mintLockReceipt)
+	var burnUnlockReceipt types.NotoDomainReceipt
+	err = json.Unmarshal(invokeResult.DomainReceipt, &burnUnlockReceipt)
```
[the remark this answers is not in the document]
peterbroadhurst: Cleaned this up with the test fix

solidity/contracts/domains/noto/Noto.sol
peterbroadhurst: I've added it as `operator` in the events where it was missing

domains/noto/internal/noto/noto.go
peterbroadhurst: No, we shouldn't. Removing.

domains/noto/internal/noto/handler_lock.go
peterbroadhurst: Comment has been removed.

> Is the "decode" below just a sanity check?

Yes, it's the endorser check of the validity of the lock transition.

domains/noto/internal/noto/events.go
```diff
@@ -69,71 +69,74 @@ func (n *Noto) handleV1Event(ctx context.Context, ev *prototk.OnChainEvent, res
...
+			txData, err := n.decodeTransactionDataV1(ctx, lockSpent.TxData)
 			if err != nil {
 				return err
 			}
-			if domainConfig.IsNotary &&
-				domainConfig.NotaryMode == types.NotaryModeHooks.Enum() &&
-				!domainConfig.Options.Hooks.PublicAddress.Equals(unlock.Sender) {
-				err = n.handleNotaryPrivateUnlockV1(ctx, req.StateQueryContext, domainConfig, &unlock)
+			n.recordTransactionInfo(ev, lockSpent.TxId, txData.InfoStates, res)
+			res.SpentStates = append(res.SpentStates, n.parseStatesFromEvent(lockSpent.TxId, lockSpent.Inputs)...)
```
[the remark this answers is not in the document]
peterbroadhurst: This is a great bit of feedback, but it cannot be addressed without more on-chain work (see 🐔 🥚 discussion on this in earlier code discussion).

I have a proposal to:
- Add a `lockStateId` parameter into the `options` so the on-chain, has knowledge of the latest off-chain
- Treat the stored value of this as an implicit `input` to `spendLock`/`cancelLock`
   - It cannot be an explicit input, or that would affect the `spendHash`/`cancelHash`

This is the smallest thing I can see to do that ensures the lock state is consumed on-chain, and off-chain.

We could make this more sophisticated in the future (such as doing more on-chain validation of the `inputs`/`outputs` of the `createLock`/`updateLock`/`delegateLock` functions that supply this state), but I do not see any necessity to do that now as it's fundamentally the notary's responsibility to track these states.

solidity/contracts/domains/interfaces/INotoHooks.sol
```diff
@@ -81,6 +81,33 @@ interface INotoHooks is IPenteExternalCall {
...
+    ) external;
+
+    function onCreateBurnLock(
+        address sender,
+        bytes32 lockId,
+        address from,
+        uint256 amount,
+        bytes calldata data,
+        PreparedTransaction calldata prepared
+    ) external;
```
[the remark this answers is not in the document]
awrichar: OK... was thinking `from`, `amount`, and `recipients` would just be zeroed when not needed, but potentially it adds headaches rather than simplifying.

solidity/contracts/domains/noto/Noto.sol
```diff
@@ -286,219 +314,379 @@ contract Noto is EIP712Upgradeable, UUPSUpgradeable, INoto, INotoErrors {
...
-            data
-        );
+        // The current lockStateId is an implicit input of the function. It cannot be passed in
+        // as the inputs were fixed at the time the spendHash/cancelHash were determined, before delegation.
+        allInputs = new bytes32[](unlockOp.inputs.length + 1);
+        uint i = 0;
+        for (; i < unlockOp.inputs.length; i++) {
+            allInputs[i] = unlockOp.inputs[i]; // the locked inputs
+        }
+        allInputs[i] = options.lockStateId; // plus the lockStateId
```
awrichar: Seems this `allInputs` list is only needed for event emission. We could consider having the Noto events emit the `lockStateId` as an explicitly separate field, in order to avoid building another in-memory array here.

sdk/typescript/src/domains/noto.ts
```diff
@@ -390,13 +382,33 @@ export class NotoInstance {
     );
   }
 
-  unlockAsDelegate(from: PaladinVerifier, data: NotoUnlockPublicParams) {
+  spendLock(
```
awrichar: I think the "public" methods (and encoding utils) here might all need updating for the final signatures in Noto.sol.

domains/noto/internal/noto/handler_delegate_lock.go
```diff
@@ -93,14 +105,33 @@ func (h *delegateLockHandler) Assemble(ctx context.Context, tx *types.ParsedTran
...
+	}
 	if err != nil {
 		return nil, err
 	}
-	infoStates = append(infoStates, lockState)
 
-	// This approval may leak the requesting signature on-chain, as all the inputs are visible on-chain
-	// TODO: possibly we should be signing a different payload here
+	// This approval may leak the requesting signing identity on-chain, if the data is empty/static.
+	// As apart from the 'data' (which is held off-chain in an info-state) all other parameters are written directly.
```
awrichar: Nice distinction here

domains/noto/internal/noto/handler_delegate_lock.go
```diff
@@ -75,8 +74,21 @@ func (h *delegateLockHandler) Assemble(ctx context.Context, tx *types.ParsedTran
...
+		minAmount = big.NewInt(1)
+	} else {
+		existingLock, err = h.noto.loadLockInfoV1(ctx, req.StateQueryContext, params.LockID)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Prepare the locked inputs.
+	lockedInputs, revert, err := h.noto.prepareLockedInputs(ctx, req.StateQueryContext, params.LockID, senderID.address, minAmount, false)
```
awrichar: We technically don't require any locked inputs in V1, right? Since the lock itself should be sufficient? If you think there's still a good reason to include at least one locked coin as a "read" state, it's OK - but I'm feeling like it isn't needed.

domains/noto/internal/noto/receipts.go
```diff
@@ -82,58 +71,67 @@ func (n *Noto) BuildReceipt(ctx context.Context, req *prototk.BuildReceiptReques
...
-	if receipt.LockInfo != nil && (len(receipt.States.PreparedOutputs) > 0 || len(receipt.States.ReadLockedInputs) > 0) {
-		// For prepareUnlock, createMintLock, and prepareBurnUnlock transactions, include the encoded "unlock"
-		// call that can be used to unlock the coins.
-		var interfaceABI abi.ABI
+	// For prepareUnlock, createMintLock, and prepareBurnUnlock transactions, include the encoded "unlock"
+	// call that can be used to unlock the coins.
+	if variant == types.NotoVariantDefault && len(receipt.States.UpdatedLockInfo) > 0 {
+		receipt.LockInfo, err = n.receiptLockInfoV1(ctx, req)
+	} else if variant == types.NotoVariantLegacy {
+
```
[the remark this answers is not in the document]
peterbroadhurst: 👍 - actioned in https://github.com/LFDT-Paladin/paladin/pull/1053

domains/noto/internal/noto/events.go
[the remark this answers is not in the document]
peterbroadhurst: This got fully resolved and worked through via multiple comments in this PR and an initial set of changes which eventually lead to the cleaner spelling in #1053

domains/noto/internal/noto/handler_delegate_lock.go
peterbroadhurst: Agreed. Made this path V0 only and removed the `readStates` for V1 in #1053

sdk/typescript/src/domains/noto.ts
peterbroadhurst: Reconciling the SDK for all the new features of V1 beyond those used in the examples has been deferred out of this PR, and isn't in #1053

peterbroadhurst: @awrichar covered this in https://github.com/LFDT-Paladin/paladin/pull/1057

Issue #983, opened 2026-01-05
bug: Heartbeats should cause both the originator and coordinator to remove finalised (`confirmed` or `reverted`) to be removed from memory


matthew1001:
### What happened?

After `heartbeatInterval` heartbeats containing a given TX in a final state (confirmed or reverted) the coordinator should remove its in-memory state machine for that transaction and stop communicating it to other coordinators and the originator.

After `heartbeatInterval` heartbeats containing a given TX in a final state (confirmed or reverted) the originator should remove its in-memory state machine for that transaction

### What did you expect to happen?

Transactions shouldn't be tracked in-memory after they reach a final state

### How can we reproduce it (as minimally and precisely as possible)?

Run a number of transactions through the `v1-develop` branch of Paladin and the number of TX logs in the coordinator builds up indefinitely.

In the originator there aren't the same logs - but the memory consumption will slowly grow.

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

PR #984, opened 2026-01-06
Clean up transactions in terminal state

annamcallister:
This PR contains fixes for https://github.com/LFDT-Paladin/paladin/issues/983. It is probably most easily reviewed as two separate commits since the changes for coordinator/originator are conceptually separate but overlap in code, resulting in some misleading diffs where new code is in the same place as older removed code.

### Clean up transactions in coordinator
diff here - https://github.com/LFDT-Paladin/paladin/pull/984/commits/cd6999f70f5871637f35819b42d85863ef3f8b95

The previous use of heartbeat events (for transaction cleanup) was
* A common heartbeat interval event existed but was not emitted anywhere
* A coordinator transaction specific heartbeat interval event was emitted when a heartbeat containing a confirmed transaction was received from a coordinator
* A coordinator would only send heartbeats to other nodes, meaning the coordinator itself would never emit this event for transactions it is coordinating and `heartbeatIntervalsSinceStateChange` was never incremented (resulting in transactions never exceeding their grace period and being cleaned up)

The new behaviour is
* There is a single common heartbeat interval event
* When a coordinator sends heartbeats to other nodes, it also emits this event for itself, and propagates it to all transactions (I have left a TODO on approach in the code here- we don't need to address it to fix this bug)
* `heartbeatIntervalsSinceStateChange` is incremented by the transaction state machine when this event is received, so transactions are now cleaned up

### Clean up transactions in originator
diff here- https://github.com/LFDT-Paladin/paladin/pull/984/commits/77bcd409a68f5a1ab3e2845777dff9a6eaae7e24

Transactions that reach the terminal states of `State_Confirmed` or `State_Reverted` now transition into `State_Final` where they are cleaned up from memory. There is no heartbeating for this- the transition occurs immediately as I don't believe there is any reason not to, but it is implemented in such a way that guards could be added in later if needed.

One fall out from this change is around the behaviour of lost assemble reverted messages. When an originator receives an assemble request, if the assemble reverts, it sends a response back to the coordinator, but also directly updates its own state machine, moving the transaction into a terminal state. Previously if the coordinator did not receive the assemble reverted response, it would send a new assemble request to the originator, and because the originator kept the transaction in memory indefinitely, it could resend the revert response back. Once the transaction is removed from memory, with no further changes the behaviour would be to ignore the assemble request, meaning that on the coordinator, the transaction would move between pooled and assembling states indefinitely- i.e. a different memory leak.

To address this I've added a new unknown transaction message type which the originator can send to the coordinator if it receives a request for a transaction it does not know about. This means that the coordinator can distinguish between "the message never reached the originator/the originator was not able to handle the message in a timely fashion" and "the originator is ignoring this transaction because they don't know about it". If the originator does not know about the transaction anymore, then it would be safe for the coordinator to also clean the transaction up in memory.

### DIFF
--- a/core/go/internal/sequencer/coordinator/state_machine.go
+++ b/core/go/internal/sequencer/coordinator/state_machine.go
@@ -459,11 +459,6 @@ func (c *coordinator) heartbeatLoop(ctx context.Context) {
 				c.heartbeatCtx = nil
 				c.heartbeatCancel = nil
 				return
-			case <-ctx.Done(): // TODO AM: don't think this is needed?
-				log.L(ctx).Infof("Cancelled heartbeat loop for %s", c.contractAddress.String())
-				c.heartbeatCtx = nil
-				c.heartbeatCancel = nil
-				return
 			}
 		}
 	}


core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -92,6 +92,7 @@ func init() {
 		State_Idle: {
 			OnTransitionTo: action_Idle,
 			Events: map[EventType]EventHandler{
+				common.Event_HeartbeatInterval: {},
```
matthew1001: After discussing this in person I think we've both agreed on the conclusion that `Event_HeartbeatInterval` should be specific to the coordinator. It should be used as an internal state machine event to cause the coordinator to do whatever it needs to send a heartbeat.

So if we make that change here, I think the only valid state to handle that event would be `State_Active`, and the event would no longer be defined in `common`.

I think that means that all of the other coordinator states should no longer accept `Event_HeartbeatInterval`.

core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -458,6 +460,8 @@ func (c *coordinator) heartbeatLoop(ctx context.Context) {
 				if err != nil {
 					log.L(ctx).Errorf("error sending heartbeat: %v", err)
 				}
+				// Queue heartbeat interval event to self so transactions can track grace periods
```
matthew1001: I think this is correct as per our agreed use of the interval event, but it should mean that we remove calls to `c.sendHeartbeat` from this function because the interval event causes the state machine to call it. (It already does handle interval event - but before this PR didn't ever receive it).

We'll also need to queue an initial event to replace the initial heartbeat in the heartbeatLoop.

core/go/internal/sequencer/coordinator/transaction/events.go
```diff
@@ -307,15 +307,19 @@ func (*StateTransitionEvent) TypeString() string {
...
 
-// events emitted by the transaction state machine whenever a state transition occurs
-type HeartbeatIntervalEvent struct {
+// TransactionUnknownByOriginatorEvent is sent by an originator when it receives a message
+// for a transaction it doesn't recognize. The most likely cause is that the transaction reached
+// a terminal state (e.g. reverted during assembly) but the response was lost, and the transaction
+// has since been removed from memory on the originator after cleanup.
+type TransactionUnknownByOriginatorEvent struct {
 	BaseCoordinatorEvent
+	AssembleRequestID uuid.UUID
```
matthew1001: I don't think this is an assemble request ID. But maybe we don't need an ID at all by virtue of the fact that `BaseCoordinatorEvent` already has the TX ID, and I'm not sure we need to correlate it with a request.

core/go/internal/sequencer/originator/state_machine.go
```diff
@@ -38,7 +38,7 @@ const (
 
 const (
 	Event_HeartbeatInterval                EventType = iota + 300 // the heartbeat interval has passed since the last time a heartbeat was received or the last time this event was received
-	Event_HeartbeatReceived                                       // a heartbeat message was received from the current active coordinator
+	Event_HeartbeatReceived                                       // a heartbeat message was received from the current active coordinatorgit
```
matthew1001:
```suggestion
	Event_HeartbeatReceived                                       // a heartbeat message was received from the current active coordinator
```

core/go/internal/sequencer/transport/transport_writer.go
```diff
@@ -623,6 +624,41 @@ func (tw *transportWriter) SendDispatched(ctx context.Context, transactionOrigin
 	return err
 }
 
+// SendTransactionUnknown is sent by an originator when it receives a message for a transaction
```
matthew1001:
```suggestion
// SendTransactionUnknown is called by an originator when it receives a message for a transaction
```

core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -129,8 +127,7 @@ func init() {
 		},
 		State_Standby: {
 			Events: map[EventType]EventHandler{
-				common.Event_HeartbeatInterval: {},
```
matthew1001: This feels better, not expecting heartbeat interval in most states.

matthew1001 (commented):
The latest changes look good, I think there's just one area to agree on before approving which is the guard in-flight function.

matthew1001 (approved):

PR #992, opened 2026-01-21
keymanager: add keymgr_sign RPC

awrichar:
This adds a new `keymgr_sign` RPC call which directly wraps the raw signing mechanism, allowing clients to request signing via Paladin rather than needing to access the underlying keystore.

The RPC can be disabled via config if needed, but is enabled by default.

### DIFF
--- a/common/go/pkg/pldmsgs/en_descriptions.go
+++ b/common/go/pkg/pldmsgs/en_descriptions.go
@@ -492,11 +492,12 @@ var (
 	KeyManagerInlineConfigWallets         = pdm("KeyManagerInlineConfig.wallets", "List of wallet configurations")
 	KeyManagerInlineConfigIdentifierCache = pdm("KeyManagerInlineConfig.identifierCache", "Identifier cache configuration")
 	KeyManagerInlineConfigVerifierCache   = pdm("KeyManagerInlineConfig.verifierCache", "Verifier cache configuration")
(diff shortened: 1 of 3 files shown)


common/go/pkg/pldmsgs/en_descriptions.go
```diff
@@ -492,6 +492,7 @@ var (
 	KeyManagerInlineConfigWallets         = pdm("KeyManagerInlineConfig.wallets", "List of wallet configurations")
 	KeyManagerInlineConfigIdentifierCache = pdm("KeyManagerInlineConfig.identifierCache", "Identifier cache configuration")
 	KeyManagerInlineConfigVerifierCache   = pdm("KeyManagerInlineConfig.verifierCache", "Verifier cache configuration")
+	KeyManagerInlineConfigDisableSignRPC  = pdm("KeyManagerInlineConfig.disableSignRPC", "True to disable the signing RPC")
```
peterbroadhurst:
```suggestion
KeyManagerInlineConfigDisableSignRPC  = pdm("KeyManagerInlineConfig.disableSignRPC", "True to disable the keymgr_sign JSON/RPC command, in order to prevent external applications from requesting arbitrary signing using the keys accessible to the node")
````

peterbroadhurst (approved):
Looks great. I've made one suggestion to expand the wording on the config option.

PR #1010, opened 2026-02-12
sdk: track subscription name to ID mapping

awrichar:
Events returned from a subscription will only include the subscription ID, which is transient and not very useful to clients if they use a single websocket for multiple subscriptions.

Track the subscription ID -> name mapping for all active websocket subscriptions, and provide a utility for lookup.

Also includes some general cleanup of websocket connect/reconnect/close logic.

### DIFF
--- a/sdk/typescript/src/interfaces/websocket.ts
+++ b/sdk/typescript/src/interfaces/websocket.ts
@@ -29,7 +29,7 @@ export interface WebSocketClientOptions<TMessageTypes extends string> {
   // Initial delay (ms) before first reconnect. With reconnectMaxDelay, doubles each attempt up to the max.
   reconnectDelay?: number;
   // If set, use exponential backoff up to a maximum delay.
-  reconnectMaxDelay?: number;
+  reconnectBackoffMaxDelay?: number;
   afterConnect?: WebSocketConnectCallback;
   socketOptions?: WebSocket.ClientOptions | http.ClientRequestArgs;
 }
--- a/sdk/typescript/src/websocket.ts
+++ b/sdk/typescript/src/websocket.ts
@@ -175,8 +175,8 @@ abstract class PaladinWebSocketClientBase<
         this.socket.removeAllListeners();
         try {
           this.socket.close();
-        } catch {
-          // ignore
+        } catch (e: any) {
+          this.logger.warn(`Failed to close websocket: ${e.message}`);
         }
         this.socket = undefined;
       }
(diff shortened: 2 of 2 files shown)


awrichar:
Unit test failure seems to be (completely coincidentally) related to a race condition in the websocket handling in the Go core.

It seems that a websocket subscription might start delivering events _before sending the response containing the subscription ID_. This race condition has probably always been possible, but it's particularly relevant to this PR. It means that if you're relying on being able to map the subscription ID (included with each event) back to a listener that you created, you might get events before you've been assigned a subscription ID. That should not happen, as it means the subscription ID is meaningless while processing that batch - in particular, it means that an application calling `getSubscriptionName()` (added in this PR) from the context of this first batch would get an `undefined` response.

I believe the websocket code needs to be refactored to guarantee that the response to the initial subscription request is always sent _before_ any events are sent out on the websocket.

sdk/typescript/src/interfaces/websocket.ts
```diff
@@ -25,7 +26,10 @@ export interface WebSocketClientOptions<TMessageTypes extends string> {
   subscriptions?: WebSocketSubscription<TMessageTypes>[];
   logger?: Logger;
   heartbeatInterval?: number;
+  // Initial delay (ms) before first reconnect. With reconnectMaxDelay, doubles each attempt up to the max.
   reconnectDelay?: number;
+  // If set, use exponential backoff up to a maximum delay.
+  reconnectMaxDelay?: number;
```
matthew1001: `reconnectBackoffMaxDelay`? It feels a little odd to just call it `reconnectMaxDelay` with no reference to the fact that it also enables backoff behaviour.

sdk/typescript/src/websocket.ts
```diff
@@ -132,18 +162,37 @@ abstract class PaladinWebSocketClientBase<
...
+    this.logger.error(`Websocket closed: ${msg}`);
+    if (this.options.reconnectDelay === -1) {
+      // Reconnection disabled - just clean up
+      this.clearPingTimers();
+      if (this.socket) {
+        this.socket.removeAllListeners();
+        try {
+          this.socket.close();
+        } catch {
+          // ignore
```
matthew1001:
```suggestion
          this.logger.warn(`Failed to close websocket`);
```

to match the logging of the other `socket.close()` catch?

matthew1001 (approved):

PR #1014, opened 2026-02-14
privatetxnmgr: use per-request channels rather than a global channel

awrichar:
The assemble coordinator is currently prone to leak requests and stall when tracking multiple requests in flight. Switch to a map of requests so that each one can track completion/timeout individually.

---

Built with Cursor Plan
Model = GPT-5.3 Codex

## Goal

Move from a single shared `commit chan string` to per-request completion channels, keyed by `requestID`, to eliminate cross-request interference and blocking risks in completion signaling.

## Implementation Plan

1. Introduce inflight request state in `[/Users/andrew/workspace/paladin/core/go/internal/privatetxnmgr/assemble_coordinator.go](/Users/andrew/workspace/paladin/core/go/internal/privatetxnmgr/assemble_coordinator.go)`:
  - Add `inflight map[string]chan struct{}` and a mutex (`sync.Mutex`).
  - Create helper methods:
    - `registerInflight(requestID string) chan struct{}`
    - `completeInflight(requestID string) (found bool)`
    - `removeInflight(requestID string)`
2. Replace global commit signaling:
  - Remove `commit chan string` from coordinator state.
  - In `Start()`, register inflight immediately before `processLocal/processRemote`.
  - In `Complete(requestID)`, lookup+close only the matching done channel; never block.
3. Rework wait logic:
  - Change `waitForDone(requestID, doneCh)` to wait on:
    - `<-doneCh`
    - `<-timeoutTimer.C`
    - `<-ac.stopProcess`
    - `<-ac.ctx.Done()`
  - Add `defer` timer stop/drain cleanup.
  - Ensure inflight entry is removed on every exit path.
4. Handle late/unknown completions safely:
  - If `Complete()` receives a request ID not in `inflight`, log as late/unknown and return.
  - This avoids blocking and removes dependence on channel buffer sizing.
5. Keep behavior scope tight:
  - Do not alter transaction-flow semantics in this pass beyond completion signaling reliability.
  - Keep timeout-to-failure-event work as a follow-up task if needed.

## Test Plan

- Add/extend tests under `core/go/internal/privatetxnmgr` to cover:
  - `Complete()` for unknown request ID is non-blocking and harmless.
  - Matching `Complete()` unblocks only the waiting request.
  - Late `Complete()` after timeout does not interfere with next request.
  - `Stop()` while waiting exits promptly (without waiting full request timeout).
  - Timer cleanup path is exercised on success and timeout.

## Scope

- Primary files:
  - `[/Users/andrew/workspace/paladin/core/go/internal/privatetxnmgr/assemble_coordinator.go](/Users/andrew/workspace/paladin/core/go/internal/privatetxnmgr/assemble_coordinator.go)`
  - `[/Users/andrew/workspace/paladin/core/go/internal/privatetxnmgr/assemble_coordinator_test.go](/Users/andrew/workspace/paladin/core/go/internal/privatetxnmgr/assemble_coordinator_test.go)` (new or extended)

### DIFF
--- a/core/go/internal/privatetxnmgr/assemble_coordinator.go
+++ b/core/go/internal/privatetxnmgr/assemble_coordinator.go
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024 Kaleido, Inc.
+ * Copyright © 2026 Kaleido, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
@@ -17,6 +17,7 @@ package privatetxnmgr
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"github.com/LFDT-Paladin/paladin/common/go/pkg/log"
@@ -30,8 +31,10 @@ type assembleCoordinator struct {
 	ctx                  context.Context
 	nodeName             string
 	requests             chan *assembleRequest
-	stopProcess          chan bool
-	commit               chan string
+	stopProcess          chan struct{}
+	stopProcessOnce      sync.Once
+	inflightLock         sync.Mutex
+	inflight             map[string]chan struct{}
 	components           components.AllComponents
 	domainAPI            components.DomainSmartContract
 	domainContext        components.DomainContext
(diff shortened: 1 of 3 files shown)


awrichar:
Closed in favor of #1015

PR #1018, opened 2026-02-16
Avoid failures ever overwriting a success receipt

peterbroadhurst:
There's only one function where we write receipts, which is driven:

- When we're indexing blocks and find a success
- When we're indexing blocks and find a failure, for transaction that originated locally
- When we've received a remote notification of a failure for a transaction we originated, from the coordinator

This PR updates that function so that it protect against cases previously we might not have recorded a success:

- When both success and failure are in the same batch (local indexer, or remotely received)
- When success has previously been persisted, and we're about to write failure

I could not find any way to avoid an additional query before insert for these receipts, because we can only have one receipt per transaction, and the failure might have been inserted before the success. So in that exceptional situation we need to:

1. Delete the failure
2. Insert the success

### DIFF
--- a/core/go/internal/txmgr/persisted_receipt.go
+++ b/core/go/internal/txmgr/persisted_receipt.go
@@ -174,7 +174,7 @@ func (tm *txManager) FinalizeTransactions(ctx context.Context, dbTX persistence.
 			var duplicateReceipts []*transactionReceipt
 			err = dbTX.DB().Table("transaction_receipts").
 				WithContext(ctx).
-				Where("`transaction` IN ?", transactionIDs).
+				Where(`"transaction" IN ?`, transactionIDs).
 				Find(&duplicateReceipts).
 				Error
 			if len(duplicateReceipts) > 0 {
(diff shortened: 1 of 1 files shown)


core/go/internal/txmgr/persisted_receipt.go
```diff
@@ -159,7 +165,52 @@ func (tm *txManager) FinalizeTransactions(ctx context.Context, dbTX persistence.
...
 		// in transaction A will be lower than transaction B (not guaranteed otherwise).
 		err := tm.p.TakeNamedLock(ctx, dbTX, "transaction_receipts")
+
+		// Failures must not override success, so when we have failure we check for previously persisted success
+		// (it's a shame we need to hold the table lock here, but we do this on multiple threads due to detecting
+		// failures remotely and successes locally)
+		var receiptsToDelete []uuid.UUID
 		if err == nil {
+			var duplicateReceipts []*transactionReceipt
+			err = dbTX.DB().Table("transaction_receipts").
```
peterbroadhurst: I couldn't find a way to avoid this query in all cases - because we might have previously written a failure, which we need to override with success.

Override in this case means delete+recreate, as the new receipt needs to be a new notification to the app listener stream.

awrichar: Is there any way we could use an "on conflict" clause to detect duplicates, rather than needing the query every time?

awrichar (approved):
Looks good. Agreed the query isn't ideal (left one question there) - but overall this is better than overwriting receipts.

PR #1019, opened 2026-02-17
Ensure websocket events aren't sent until subscription is set up

awrichar:
Websocket subscriptions now begin in a "pending" state and are only moved to the "active" list after the initial websocket response that notifies the subscriber of their subscription ID.

This ensures that no events are delivered before the subscription ID is known to the subscriber.

### DIFF
--- a/toolkit/go/pkg/rpcserver/rpchandler_test.go
+++ b/toolkit/go/pkg/rpcserver/rpchandler_test.go
@@ -18,18 +18,49 @@ package rpcserver
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"testing"
 	"testing/iotest"
+	"time"
 
 	"github.com/LFDT-Paladin/paladin/config/pkg/pldconf"
 	"github.com/LFDT-Paladin/paladin/sdk/go/pkg/pldtypes"
 	"github.com/LFDT-Paladin/paladin/sdk/go/pkg/rpcclient"
 	"github.com/go-resty/resty/v2"
+	"github.com/gorilla/websocket"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+type asyncHandlerForTest struct {
+	postSendDone chan struct{}
+}
+
+func (h *asyncHandlerForTest) StartMethod() string {
+	return "ut_asyncStart"
+}
+
+func (h *asyncHandlerForTest) LifecycleMethods() []string {
+	return nil
+}
+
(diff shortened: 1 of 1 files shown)


core/go/internal/groupmgr/message_listener.go
```diff
@@ -495,29 +500,48 @@ func (l *messageListener) addReceiver(r components.PrivacyGroupMessageReceiver)
...
-			}
+	l.receivers = l.removeReceiverFromList(l.receivers, rid)
+	l.pendingReceivers = l.removeReceiverFromList(l.pendingReceivers, rid)
+}
+
+func (l *messageListener) removeReceiverFromList(receivers []*registeredMessageReceiver, rid uuid.UUID) []*registeredMessageReceiver {
+	if len(receivers) == 0 {
+		return receivers
+	}
+	newReceivers := make([]*registeredMessageReceiver, 0, len(receivers))
```
matthew1001: `len(receivers)-1` ? I guess it's possible the UUID isn't in the original list though?

core/go/internal/groupmgr/rpc_eventstreams.go
```diff
@@ -72,31 +72,36 @@ func (es *rpcEventStreams) HandleStart(ctx context.Context, req *rpcclient.RPCRe
...
+	afterSend = func() {
+		sub.pgmrc.SetActive()
 	}
 
 	return sub, &rpcclient.RPCResponse{
 		JSONRpc: "2.0",
 		ID:      req.ID,
 		Result:  pldtypes.JSONString(ctrl.ID()),
-	}
+	}, afterSend
```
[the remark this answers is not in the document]
awrichar: I _believe_ it's removing the window, because we now push the subscription response to the websocket (which will succeed or fail) before making the subscription an active candidate for delivering subscription messages. But would appreciate confirmation on that reasoning.

toolkit/go/pkg/rpcserver/websockets.go
```diff
@@ -172,6 +172,9 @@ func (c *webSocketConnection) handleMessage(payload []byte) {
 	if r.sendRes {
 		c.sendMessage(r.res)
```
awrichar: @matthew1001 here is where we send the response, and below is the "postSend" callback. Here it should block on sending the message unless the context is cancelled.

awrichar: Currently we invoke `postSend` even if `sendRes` is nil... maybe need to think in theory about whether you'd have one without the other (although for our actual implementation, "start" is the only async call and it provides both)

PR #1029, opened 2026-02-19
Repool coordinator transactions after all stage based timeouts

annamcallister:
Extend stage based timeouts to `State_EndorsementGathering` and `State_ConfirmingDispatchable` - these previously had  request timeouts to trigger a retry, but if the retry did not result in a response from the request, there was no path to transition out of this state.

I have chosen to make the state based timeout common for all states (replacing `assembleTimeout`). I have also made the state timeout tied to entry into the state, rather than first request time, after observing transactions in a system under high load get stuck in the state because it hadn't yet managed to make a first request when the state timeout fired.

Ensure that repooling causes all dependents to roll back correctly. On entering the pool a transaction now (in addition to existing reset behaviour)
* emits a dependency repooled event to all its dependencies (which causes them to repool, making this recursive through the dependency tree)
* removes any in memory state locks it has created

I have also done some tidy up of `Event_DependencyReverted` as it looked like this was frequently being handled as though it were a base ledger revert, whereas it is actually an assembly revert, where the transaction today has no dependencies at this point. The repool logic described above handles repooling dependents in case of a base ledger revert.

Alongside tidying `Event_DependencyReverted` I have also done some tidying and documenting around `State_PreAssemblyBlocked` in anticipation of the FIFO work, as it is closely related to the event handling in this PR.

### DIFF
--- a/core/go/noderuntests/componenttest/component_test.go
+++ b/core/go/noderuntests/componenttest/component_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/LFDT-Paladin/paladin/config/pkg/confutil"
 	testutils "github.com/LFDT-Paladin/paladin/core/noderuntests/pkg"
 	"github.com/LFDT-Paladin/paladin/core/noderuntests/pkg/domains"
+	"github.com/google/uuid"
 	"github.com/hyperledger/firefly-signer/pkg/abi"
 
 	"github.com/LFDT-Paladin/paladin/sdk/go/pkg/pldapi"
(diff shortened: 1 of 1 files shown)


core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -131,7 +146,7 @@ var stateDefinitionsMap = StateDefinitions{
 		},
 	},
 	State_Assembling: {
-		OnTransitionTo: action_SendAssembleRequest,
+		OnTransitionTo: action_OnTransitionToAssembling,
```
matthew1001: We're sort of losing any descriptive benefits of looking through the state machine and understanding at a glance what a transition is going to do. Maybe that's fine if we're going to do multiple things inside the function, but in other places we've moved to an array of actions to take, which makes the state machine much clearer to read at a glance. Just reflecting that this feels like a change that should be a conscious one, rather than an unconcsious one.

matthew1001 (approved):
Left a couple of comments but I'm happy to defer anything necessary to future PRs if you prefer.

PR #1030, opened 2026-02-19
Update transactions based on heartbeat contents

annamcallister:
This was already done partially but some updates required confirmation to come for the specific transaction, (e.g. nonce assigned, confirmed etc).

Allowing all state transitions to happen as a result of the heartbeat improves resiliency under high load where some transport messages might get lost or delayed.

Because a transaction may go through multiple states within a heartbeat interval, the originator transaction state machine needs to be able to skip any number of intermediate states. This has the effect of making the state machine definition more verbose, but I think this full specification is preferable to forcing a replay of events when we apply a heartbeat to fill in any potential gaps.


annamcallister:
Closing to deliver a subset of these changes under #1052

PR #1031, opened 2026-02-19
Internal priority queues

annamcallister:
Use the internal priority queue for queueing state transitions back from transactions to originator/coordinator and ensure that they are set by defualt to a size large enough not to cause deadlocks.

We have discussed that this is probably not the right pattern going forwards, and that relying on a queue size which has to increase with load, is a bad smell, but this PR provides an intermediate step before we can resolve this to make sure the system can function under load.


matthew1001 (approved):
Looks good and keeps the originator in keeping with the coordinator approach to internal events.

PR #1044, opened 2026-02-25
Grace period for removing locks

annamcallister:
There is a timing window where if in memory locks are removed on the coordinator immediately after transaction confirmation, an assembly request may be sent to another node which hasn't yet indexed the state spends, resulting in double spend.

Holding on to the in memory locks for a period of time (currently defaulted to one heart beat interval) removes this window in most cases. There is some duplication if in memory locks are held on top of equivalent DB state but this does cause incorrect selection. The retention period is configurable but if another node is sufficiently behind in its indexing that it requires multiple heartbeat intervals, there is likely wider problems with the system.

As mentioned in comments, this could be handled more elegantly with block height at a later point.


core/go/internal/sequencer/coordinator/transaction/confirming.go
```diff
@@ -35,7 +35,12 @@ func action_Confirmed(ctx context.Context, t *CoordinatorTransaction, event comm
 
 func action_NotifyDependantsOfConfirmation(ctx context.Context, txn *CoordinatorTransaction, _ common.Event) error {
 	log.L(ctx).Debugf("action_NotifyOfConfirmation - notifying dependents of confirmation for transaction %s", txn.pt.ID.String())
-	txn.engineIntegration.ResetTransactions(ctx, txn.pt.ID)
+	txn.confirmedLocksReleased = false
```
matthew1001: Is this not the default value at this point anyway, or do we need to be able to reset from `true` to `false` in some cases?

matthew1001 (approved):
LGTM, minor comment

PR #1048, opened 2026-02-25
Validate that contents of receipts response matches block repsonse

annamcallister:
Extra validation in response to an issue we saw where one Besu node reported fewer receipts for a block than the other nodes, resulting in one Paladin node permanently missing a state spend.

It's not certain that this fix would have stopped the issue as we don't how many transactions the getBlockByNumber response returned, but having identified this area as a place where discrepancies can be critical, it makes sense to increase any validation we can so that we don't persist blocks we know we can't trust.

The error log will be large but an error of this severity will be sufficiently rare and severe enough to justify it.

### DIFF
--- a/core/go/pkg/blockindexer/block_indexer_test.go
+++ b/core/go/pkg/blockindexer/block_indexer_test.go
@@ -1391,7 +1391,7 @@ func TestHydrateBlockReceiptTxHashMismatchReturnsMissing(t *testing.T) {
 	batch.wg.Add(1)
 
 	bi.hydrateBlock(ctx, batch, 0)
-	assert.Regexp(t, "PD011317", batch.receiptResults[0])
+	assert.Regexp(t, "PD011316", batch.receiptResults[0])
 	batch.wg.Wait()
 }
 
@@ -1442,7 +1442,7 @@ func TestHydrateBlockReceiptDuplicateCausesMissing(t *testing.T) {
 	batch.wg.Add(1)
 
 	bi.hydrateBlock(ctx, batch, 0)
-	assert.Regexp(t, "PD011317", batch.receiptResults[0])
+	assert.Regexp(t, "PD011316", batch.receiptResults[0])
 	batch.wg.Wait()
 }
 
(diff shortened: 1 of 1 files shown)


core/go/internal/msgs/en_errors.go
```diff
@@ -204,17 +204,21 @@ var (
...
+	MsgBlockIndexerNoBlocksIndexed            = pde("PD011308", "No confirmed blocks have yet been indexed")
+	MsgBlockIndexerTransactionReverted        = pde("PD011309", "Transaction reverted: %s")
+	MsgBlockIndexerConfirmedBlockNotFound     = pde("PD011310", "Block %s (%d) not found on retrieval after detection and requested number of confirmations")
+	MsgBlockIndexerLimitRequired              = pde("PD011311", "limit is required on all queries")
+	MsgBlockIndexerEventStreamNotFound        = pde("PD011312", "Event stream not found: %s")
+	MsgBlockIndexerReceiptCountMismatch       = pde("PD011313", "Receipt integrity check failed for block %s (%d): expected %d receipts from block transactions, received %d")
+	MsgBlockIndexerReceiptBlockMismatch       = pde("PD011315", "Receipt integrity check failed for block %s (%d): receipt for tx %s references block %s (%d)")
+	MsgBlockIndexerReceiptIntegrityNilReceipt = pde("PD011316", "Receipt integrity check failed for block %s (%d): nil receipt entry")
+	MsgBlockIndexerReceiptMissingTxHash       = pde("PD011317", "Receipt integrity check failed for block %s (%d): missing transaction hash from receipts %s")
```
matthew1001:
```suggestion
	MsgBlockIndexerReceiptBlockMismatch       = pde("PD011314", "Receipt integrity check failed for block %s (%d): receipt for tx %s references block %s (%d)")
	MsgBlockIndexerReceiptIntegrityNilReceipt = pde("PD011315", "Receipt integrity check failed for block %s (%d): nil receipt entry")
	MsgBlockIndexerReceiptMissingTxHash       = pde("PD011316", "Receipt integrity check failed for block %s (%d): missing transaction hash from receipts %s")
```

PR #1049, opened 2026-02-25
Remove reverse links on transaction forget

annamcallister:
These changes should become obsolete with the new grapher but it fixes an issue where if a transaction was confirmed and removed from the grapher before its dependent was ready for dispatch (e.g. because of slowness in endorsement gathering), it's dependent could never pass its readiness checks as its prereq was not found.


matthew1001 (approved):
Added another commit that removes the no-op lines we discussed in person, and adding my approval to it since it basically leaves a subset of your original changes

PR #1050, opened 2026-02-25
Collapse dispatched states, try queue for public tx manager events

annamcallister:
There is currently a deadlock situation where if public transaction manager blocks trying to write events back into the sequencer, the sequencer can never progress because no transactions ever get submitted or confirmed.

This PR makes events from the public transaction manager purely informational by
* using TryQueue- so that the event gets dropped if the queue is full
* collapsing the states that relied on these events for their transitions to just dispatched state

As a result it depends on https://github.com/LFDT-Paladin/paladin/pull/1030 to move originator transactions through missing states if the events that would have caused events such as nonce assigned to be propagated back to the originator are dropped.

We have discussed that this is a temporary resolution to ensure we are reliable until we can move towards a model which has substates and queueing behaviour encoded on the event.

### DIFF
--- a/core/go/internal/sequencer/coordinator/transaction/assembling_test.go
+++ b/core/go/internal/sequencer/coordinator/transaction/assembling_test.go
@@ -452,7 +452,6 @@ func Test_isNotAssembled_AssembledStates(t *testing.T) {
 		State_Confirming_Dispatchable,
 		State_Ready_For_Dispatch,
 		State_Dispatched,
-		State_Dispatched,
 		State_Confirmed,
 	}
 


core/go/internal/sequencer/coordinator/transaction/assembling_test.go
```diff
@@ -452,7 +452,7 @@ func Test_isNotAssembled_AssembledStates(t *testing.T) {
 		State_Confirming_Dispatchable,
 		State_Ready_For_Dispatch,
 		State_Dispatched,
-		State_Submitted,
+		State_Dispatched,
```
matthew1001: Dup?

PR #1054, opened 2026-02-27
Make the grapher the source of truth for dependencies

matthew1001:
Paladin transactions are assembled based on a combination of:

 1. The states an originator knows about
 2. Any potential state changes from unconfirmed transactions (potentially created states, and locks of existing states)

The domain context is intended to provide a "snapshot" of context that a domain smart contract can use as the basis for domain logic. When a node is asked to assemble a transaction it already has `1`, and is given `2` by the coordinating node in the assemble request.

## Context

Up to now, the coordinator has used a long-lived domain context to track potential state changes and locks. The new distributed sequencer introduced a basic `grapher` component for tracking potential state changes and locks, but it wasn't properly integrated into the sequencer:

 - It wasn't used to create the point-in-time snapshot of potential states and locks, even though it tracked them
 - It wasn't used to track dependencies between transactions, requiring it to separately update individual transaction objects by adding or removing dependencies it could infer from the locked and minted states.

## PR aims

- Move the grapher from the `transaction` package to its own `grapher` package, for code cleanliness and better mocking & test coverage 
- Remove the use of a long-lived domain context just for tracking point-in-time states ahead of the chain
    - Since the grapher tracks potential and locks states, it provides the snapshots to send in assemble requests
 - Complete the grapher implementation to ensure new states and locks are tracked correctly, and removed appropriately when transactions confirm or revert
 - Add TX-to-TX dependency tracking to the grapher
    - If TX2 locks state produced by TX1, it has TX1 as a pre-req and and TX1 has TX2 as a dependant 
    - This eliminates the need for transactions to track their own dependencies
 - While I was tracking metrics I took the opportunity to add `pgroup_*` RPC metrics

## Notes on the design

 - The grapher has from day one used the following terms, and while there has been some discussion about this (primarily for the term `minter`) my current opinion is that they work well within the confines of the grapher package:
    - `minter` - the transaction that produces some state
    - `lock` - a lock on minted state
 - The grapher is thread safe for all calls to its interface functions
 - Nothing outside the grapher asserts post-assemble dependencies
    - When state is minted, locked, or forgotten, the grapher is responsible for making sure the dependency chains are correct
 - In support of both this PR, and the chained-TX dependencies and pre-assemble dependencies, I've added another new package `dependencytracker`. This is a generic package for tracking inter-TX dependencies, using a new interface `DependencyChain`, which it defines 3 instances of (pre-assemble, post-assemble, chained-TX)
   - The new package might evolve a little over time, but I think the basic structure is sound. I chopped and changed a little in terms of how much paladin-awareness the `dependencytracker` package had and landed on:
      - `DependencyTracker` is Paladin-aware (knows about the 3 types of dependency)
      - `DependencyChain` is completely generic
 - The `post-assemble` dependency chain is managed by the grapher. The grapher is given the dependency tracker to retrieve and update the chain when needed. This is currently enforced except by code conventions.
 - Explicit (application-defined) dependencies are not currently managed by the dependency tracker

## WIP notes while closing out the PR

 - I owe some docs updates which aren't yet in the PR
 - I believe the `DomainSmartContract` function `LockStates()` is no longer required, and I'll remove unless we think there's a reason not to


core/go/internal/domainmgr/private_smart_contract.go
```diff
@@ -303,10 +303,11 @@ func (dc *domainContract) WritePotentialStates(dCtx components.DomainContext, re
 		log.L(dCtx.Ctx()).Debugf("WritePotentialStates: Writing %+v info potentialstates", len(postAssembly.InfoStatesPotential))
 		postAssembly.InfoStates, err = dc.upsertPotentialStates(dCtx, readTX, tx, postAssembly.InfoStatesPotential, false)
 	}
-	log.L(dCtx.Ctx()).Debugf("WritePotentialStates: %d post assembly output states", len(postAssembly.OutputStates))
-	log.L(dCtx.Ctx()).Debugf("WritePotentialStates: %d post assembly info states", len(postAssembly.InfoStates))
 	return err
+}
 
+func (dc *domainContract) MapPotentialStates(dCtx components.DomainContext, potentialStates []*prototk.NewState, outputStates bool, createdByTX *components.PrivateTransaction) (stateUpserts []*components.StateUpsert, err error) {
+	return dc.d.mapPotentialStates(dCtx, potentialStates, outputStates, createdByTX)
```
annamcallister, line 310: It's not new to this PR but reading this code I realise I can't obviously tell what formats we're mapping between in this function and I think a comment would be helpful, especially now its exported outside the package.

core/go/internal/groupmgr/groupmgr_rpc.go
```diff
@@ -60,33 +60,38 @@ func (gm *groupManager) rpcCreateGroup() rpcserver.RPCHandler {
 			group, err = gm.CreateGroup(ctx, dbTX, &spec)
 			return err
 		})
+		gm.metrics.IncRpc("createGroup")
```
annamcallister, line 63: As more of these get added in future PRs, we should consider making it more generic on the RPC server.

core/go/internal/sequencer/common/engine_integration.go
```diff
@@ -170,9 +135,8 @@ func (e *engineIntegration) AssembleAndSign(ctx context.Context, transactionID u
 	// if our block height is ahead of the coordinator, there is a small chance that we we assemble a transaction that the coordinator will not be able to
 	// endorse yet but it is better to wait around on the endorsement flow than to wait around on the assemble flow which is single threaded per domain
 
-	// TODO - we're not actually policing anything based on block height differences?
-
-	err := e.delegateDomainContext.ImportSnapshot(ctx, stateLocksJSON)
+	delegateDomainContext := e.components.StateManager().NewDomainContext(ctx, e.domainSmartContract.Domain(), e.domainSmartContract.Address())
```
annamcallister: I can't see that this domain context ever gets closed?

core/go/internal/sequencer/coordinator/transaction/assembling.go
```diff
@@ -83,16 +83,28 @@ func (t *coordinatorTransaction) applyPostAssembly(ctx context.Context, postAsse
...
-	return t.calculatePostAssembleDependencies(ctx)
+
+	// Add a lock for every output we create
+	t.grapher.LockMintsOnCreate(ctx, createLocks, postAssembly.OutputStates, t.pt.ID)
+
+	// Add a lock for every read state
+	t.grapher.LockMintsOnRead(ctx, postAssembly.ReadStates, t.pt.ID)
+
+	// Add a lock for every input state to prevent other transactions using them
+	t.grapher.LockMintsOnSpend(ctx, postAssembly.InputStates, t.pt.ID)
```
annamcallister: I can't see that these grapher functions are ever called anywhere else - could these be captured as a single function in the grapher instead?

core/go/internal/sequencer/coordinator/transaction/assembling.go
```diff
@@ -105,18 +117,20 @@ func (t *coordinatorTransaction) sendAssembleRequest(ctx context.Context) error
 	// and nudge the request every requestTimeout event to implement the short retry.
 	// The state machine will deal with the longer state timeout via timeout guards.
 	t.pendingAssembleRequest = common.NewIdempotentRequest(ctx, t.clock, t.requestTimeout, func(ctx context.Context, idempotencyKey uuid.UUID) error {
-		stateLocks, err := t.engineIntegration.GetStateLocks(ctx)
+		grapherStatesAndLocks, err := t.grapher.ExportMints(ctx)
 		if err != nil {
-			log.L(ctx).Errorf("failed to get engine state locks: %s", err)
+			log.L(ctx).Errorf("failed to export grapher state locks: %s", err)
```
annamcallister, line 119: The function name and log don't tie up. Should the grapher function be called `ExportStateLocks`?

core/go/internal/sequencer/coordinator/transaction/assembling.go
```diff
@@ -105,18 +117,20 @@ func (t *coordinatorTransaction) sendAssembleRequest(ctx context.Context) error
...
 			return err
 		}
+		log.L(ctx).Debugf("grapher state locks: %s", string(grapherStatesAndLocks))
```
annamcallister, line 122: Is this log the reason they're being exported as a byte array?

In all other cases we're passing structs to the transport writer, and letting the transport writer do the json serialization.

core/go/internal/sequencer/coordinator/transaction/assembling.go
```diff
@@ -138,9 +152,9 @@ func (t *coordinatorTransaction) notifyPreAssembleDependentOfSelection(ctx conte
 	if t.preAssemblePrereqOf == nil {
 		return nil
 	}
-	dependent := t.grapher.TransactionByID(ctx, *t.preAssemblePrereqOf)
+	dependent := t.getCoordinatorTransaction(ctx, *t.preAssemblePrereqOf)
```
annamcallister, line 152: Now that we're only accessing other transactions to handle events, I wonder if rather than a getter for the full transaction, there should be a function passed into the transaction which can be used to call `handleEvent` on a given transaction by ID?

core/go/internal/sequencer/coordinator/transaction/ready_for_dispatch.go
```diff
@@ -73,10 +73,16 @@ func (t *coordinatorTransaction) hasDependenciesNotReady(ctx context.Context) bo
 	// We already calculated the dependencies when we got assembled and there is no way we could have picked up new dependencies without a re-assemble
 	// some of them might have been confirmed and removed from our list to avoid a memory leak so this is not necessarily the complete list of dependencies
 	// but it should contain all the ones that are not ready for dispatch
-	for _, dependencyID := range t.dependencies.DependsOn {
-		dependency := t.grapher.TransactionByID(ctx, dependencyID)
+
+	dependencies := t.grapher.GetDependencies(ctx, t.pt.ID)
+	if t.pt.PreAssembly != nil && t.pt.PreAssembly.Dependencies != nil && t.pt.PreAssembly.Dependencies.DependsOn != nil {
```
annamcallister: I have finally taken steps to remove `PreAssembly.Dependencies` in #1146, but I don't think it is necessary to add them back in here ahead of merging with that code.

core/go/internal/sequencer/coordinator/transaction/ready_for_dispatch.go
```diff
@@ -107,10 +113,10 @@ func (t *coordinatorTransaction) notifyDependentsOfReadiness(ctx context.Context
 
 	//this function is called when the transaction enters the ready for dispatch state
 	// and we have a duty to inform all the transactions that are dependent on us that we are ready in case they are otherwise ready and are blocked waiting for us
-	for _, dependentId := range t.dependencies.PrereqOf {
-		dependent := t.grapher.TransactionByID(ctx, dependentId)
+	for _, dependentID := range t.grapher.GetDependants(ctx, t.pt.ID) {
```
annamcallister: The more I read this repeated code pattern, it makes me wonder whether my earlier comment about exposing a handleEvent function rather than a get transaction function from the coordinator should go further and include getting the dependents from the grapher - so from the transaction all you say is ask my dependents to handle this event

core/go/internal/sequencer/coordinator/transaction/transaction.go
```diff
@@ -79,8 +77,7 @@ type coordinatorTransaction struct {
 	pendingEndorsementRequests   map[string]map[string]*common.IdempotentRequest //map of attestationRequest names to a map of parties to a struct containing information about the active pending request
 	pendingPreDispatchRequest    *common.IdempotentRequest
 
-	// Transaction dependencies - used for tracking assembly and dispatch order
-	dependencies         *pldapi.TransactionDependencies
+	// Transaction pre-assemble dependencies - used for tracking pre-assembly and dispatch order
```
annamcallister, line 80:
```suggestion
	// Transaction pre-assemble dependencies - used for tracking pre-assembly order
```

core/go/internal/sequencer/coordinator/coordinating.go
```diff
@@ -49,7 +49,10 @@ func action_TransactionsDelegated(ctx context.Context, c *coordinator, event com
 }
 
 func (c *coordinator) newCoordinatorTransaction(ctx context.Context, originator string, originatorNode string, nodeName string, pt *components.PrivateTransaction, coordinatorSigningIdentity string, preAssembleDependsOn *uuid.UUID) transaction.CoordinatorTransaction {
-	return transaction.NewTransaction(ctx, originator, originatorNode, nodeName, pt, coordinatorSigningIdentity, preAssembleDependsOn, c.transportWriter, c.clock, c.queueEventInternal, c.engineIntegration, c.syncPoints, c.components, c.domainAPI, c.dCtx, c.requestTimeout, c.stateTimeout, c.closingGracePeriod, c.confirmedLockRetentionGracePeriod, c.baseLedgerRevertRetryThreshold, c.assembleErrorRetryThreshhold, c.grapher, c.metrics)
+	return transaction.NewTransaction(ctx, originator, originatorNode, nodeName, pt, coordinatorSigningIdentity, preAssembleDependsOn, c.transportWriter, c.clock, c.queueEventInternal, func(ctx context.Context, id uuid.UUID) transaction.CoordinatorTransaction {
+		// MRW TODO unsafe placeholder
```
annamcallister, line 53: Does switching this to be a function that calls `HandleEvent` on transactions (which is protected with a lock) fix this unsafe TODO?

core/go/internal/sequencer/sequencer_lifecycle.go
```diff
@@ -184,22 +179,18 @@ func (sMgr *sequencerManager) loadSequencer(ctx context.Context, dbTX persistenc
...
-			delegateDomainContext := sMgr.components.StateManager().NewDomainContext(sMgr.ctx, domainAPI.Domain(), contractAddr)
-
 			seqCtx, cancelCtx := context.WithCancel(log.WithComponent(sMgr.ctx, "sequencer"))
 			seqCtx = log.WithLogField(seqCtx, "contract", contractAddr.String())
 
 			// Create a transport writer for the sequencer to communicate with sequencers on other peers
 			transportWriter := transport.NewTransportWriter(seqCtx, &contractAddr, sMgr.nodeName, sMgr.components.TransportManager(), sMgr.HandlePaladinMsg)
 
-			sMgr.engineIntegration = common.NewEngineIntegration(seqCtx, sMgr.components, sMgr.nodeName, domainAPI, dCtx, delegateDomainContext, sMgr)
+			sMgr.engineIntegration = common.NewEngineIntegration(seqCtx, sMgr.components, sMgr.nodeName, domainAPI, dCtx, sMgr)
```
annamcallister: This isn't new- but this seems odd that `sMgr.engineIntegration` is created on the sequencer manager every time a sequencer is loaded, with the domain context for that sequencer.

Does the sequencer manager ever use this this property- or is it only passed into coordinators?

core/go/internal/txmgr/persisted_receipt.go
```diff
@@ -115,7 +115,6 @@ func (tm *txManager) FinalizeTransactions(ctx context.Context, dbTX persistence.
 			Indexed:         pldtypes.TimestampNow(),
 			ContractAddress: ri.ContractAddress,
 		}
-		log.L(ctx).Debugf("FinalizeTransactions: created receipt object %v, receipt type %+v", receipt, ri.ReceiptType)
```
annamcallister, line 118: what is the motivation for removing this log?

annamcallister (commented):
A first review of every thing _except_ the new grapher.

I think we might want to have a follow on conversation about `DependenciesMustWait`  as it becomes the only remaining place where one transaction calls another without a lock

core/go/internal/sequencer/coordinator/transaction/ready_for_dispatch.go
matthew1001: Yeah I think this was a merge mistake, just noticed in reviewing the other PR :)

core/go/internal/sequencer/common/engine_integration.go
matthew1001: Good spot, added a commit to close and I've renamed it to `dCtx` in keeping with other throw-away domain contexts. It was only `delegateDomainContext` to distinguish between them, but that's no longer required.

core/go/internal/sequencer/coordinator/transaction/assembling.go
matthew1001: I've collapsed `LockMintsOnRead` and `LockMintsOnSpend` into a single function call `LockMintsOnReadAndSpend`. I've left `LockMintsOnCreate` alone as it has a different function signature.

core/go/internal/sequencer/sequencer_lifecycle.go
matthew1001: Agreed, all in favour of removing unnecessary fields. Gone in the latest commit

core/go/internal/txmgr/persisted_receipt.go
matthew1001: For some reason when reviewing the PR files it looked like I'd added this, and I wasn't sure it was helpful so removed it. I'll reinstate it.

core/go/internal/sequencer/coordinator/dependencytracker/dependency_tracker.go
```diff
@@ -0,0 +1,309 @@
...
+//   - pre-assembly dependencies. These are determined the first time transactions are received, to maintain FIFO order
+//     of delegated tranasctions
+//   - post-assembly dependencies. These are determine by the grapher package. The grapher updates the PostAssembly chain in
+//     the dependency tracker directly, based on calls to the grapher to mint/lock/clear states.
+//   - chained dependencies. These are determined when transaction dispatches result in a new chained transactions, where each of
+//     the chained transactions has a dependency link with the chained transactions of previous parent transactions.
+
+// The dependency tracker also implements a discrete interface DependencyChain which each over the above types of dependency is
+// represented by. Some dependencies are graph-like (e.g. TX3 pre-reqs TX1 and TX2). Some are linked-lists, and the DependecyChain
+// constructor allows the caller to specify this, to allow for stricter checks are chain-manipulation time. The DependencyChain
```
annamcallister:
```suggestion
// The dependency tracker also implements a discrete interface DependencyChain which each of the above types of dependency is
// represented by. Some dependencies are graph-like (e.g. TX3 pre-reqs TX1 and TX2). Some are linked-lists, and the DependencyChain
// constructor allows the caller to specify this, to allow for stricter checks at chain-manipulation time. The DependencyChain
```

core/go/internal/sequencer/coordinator/dependencytracker/dependency_tracker.go
```diff
@@ -0,0 +1,309 @@
...
+	GetChainedDeps() ChainedTransactionDependencyChain
+}
+
+// DependencyChain records, for each transaction ID, which other transactions it depends on
+// and which transactions depend on it. The interface (currently) has one method for adding dependencies,
+// via AddPrerequisites() which updates all the pre-req and depends-on lists. There are getters for the
+// individual lists, and functions to remove from just one of the lists.
+type DependencyChain interface {
+	AddPrerequisites(txID uuid.UUID, prereq ...uuid.UUID) // Updates both the pre-req-of and depends-on lists
+	ClearPrerequisites(txID uuid.UUID)                    // Remove all depends-on entries this transaction for the transaction, updating their pre-req-of chains as well
```
annamcallister:
```suggestion
	ClearPrerequisites(txID uuid.UUID)                    // Remove all depends-on entries for this transaction for the transaction, updating their pre-req-of chains as well
```

core/go/internal/sequencer/coordinator/dependencytracker/dependency_tracker.go
```diff
@@ -0,0 +1,309 @@
...
+		cd.unassembledDependencies[txID][unassembledDependencyID] = struct{}{}
+	}
+}
+
+func (cd *chainedDependencies) DeleteUnassembledDependencies(txID uuid.UUID, unassembledDependencyIDs ...uuid.UUID) {
+	cd.mu.Lock()
+	defer cd.mu.Unlock()
+	for _, unassembledDependencyID := range unassembledDependencyIDs {
+		delete(cd.unassembledDependencies[txID], unassembledDependencyID)
+	}
```
annamcallister, line 169: I think there should be an additional step after this for loop to do 
```
if `len(cd.unassembledDependencies[txID]) == 0 {
  delete(cd.unassembledDependencies, txID)
}
```
Otherwise we're slowly leaking empty map entries in a long running coordinator.

core/go/internal/sequencer/coordinator/dependencytracker/dependency_tracker.go
```diff
@@ -0,0 +1,309 @@
...
+}
+
+// AddPrerequisites records that txID depends on each prerequisite (prerequisite IDs must
+// complete before txID). Updates both sides of each edge.
+func (d *dependencyChain) AddPrerequisites(txID uuid.UUID, prereq ...uuid.UUID) {
+	if d.singleChainOnly {
+		if len(prereq) > 1 {
+			panic("singleChainOnly dependency chain does not support multiple prerequisites")
+		}
+	}
```
annamcallister: This looks like it stops more than one prereq from being added in a single call, but doesn't prevent the function from being called twice to add a second prereq

core/go/internal/sequencer/coordinator/dependencytracker/dependency_tracker.go
annamcallister: I can see that the `Delete` function on a `dependencyChain` fully removes a transaction from `nodes`. But there's no `Delete` function defined on `chainedDependencies` which combines both these types of delete.

We might also be missing a `Delete` function on the `dependencyTracker` which can call `Delete` on all types of dependency chain and can be called at the point where a transaction is removed from the coordinator to be completely confident that nothing has been left behind.

core/go/internal/sequencer/coordinator/transaction/assembling.go
```diff
@@ -142,71 +154,26 @@ func action_NotifyDependentsOfSelection(ctx context.Context, txn *coordinatorTra
 
 func (t *coordinatorTransaction) notifyDependentsOfSelection(ctx context.Context) error {
 	var dependentIDs []uuid.UUID
-	if t.dependencies.PreAssemble.PrereqOf != nil {
-		dependentIDs = append(dependentIDs, *t.dependencies.PreAssemble.PrereqOf)
-	}
-	dependentIDs = append(dependentIDs, t.dependencies.Chained.PrereqOf...)
+	// Get transactions this is a pre-req, including chained dependencies
```
annamcallister:
```suggestion
	// Get transactions this is a pre-req of, including chained dependencies
```

core/go/internal/sequencer/coordinator/transaction/confirming.go
```diff
@@ -148,22 +148,15 @@ func action_NotifyDependantsOfRevertedConfirmation(ctx context.Context, txn *coo
...
-			})
-			if err != nil {
-				log.L(ctx).Errorf("error notifying dependent transaction %s of revert of transaction %s: %s", dependent.GetPrivateTransaction().ID, t.pt.ID, err)
-				return err
-			}
-		}
+	dependents := t.dependencyTracker.GetPostAssemblyDeps().GetDependents(t.pt.ID)
+	chainedDependents := t.dependencyTracker.GetChainedDeps().GetDependents(t.pt.ID)
+	for _, dependentId := range append(dependents, chainedDependents...) {
+		t.queueEventForCoordinator(ctx, &DependencyConfirmedRevertedEvent{
```
annamcallister: This is a really significant change to how unwind will happen  because it moves from a cascading synchronous `handleEvent` to a queued event via the coordinator. We could probably make this work given enough testing, but I don't know that we want to bite that off right now when we've already put a lot of work into hardening the synchronous cascade.

I think the intent of my original review comment was to avoid giving one transaction full access to another, so to expose only a function which allowed one transaction to access `HandleEvent` on another transaction.

core/go/internal/sequencer/coordinator/transaction/pooled.go
```diff
@@ -29,14 +29,14 @@ func (t *coordinatorTransaction) hasDependenciesNotAssembled(ctx context.Context
...
-	if t.dependencies.PreAssemble.DependsOn != nil {
+	if len(t.dependencyTracker.GetPreassemblyDeps().GetPrerequisites(t.pt.ID)) > 0 {
 		return true
 	}
 	return t.hasUnassembledChainedDependencies(ctx)
 }
 
 func (t *coordinatorTransaction) hasUnassembledChainedDependencies(_ context.Context) bool {
-	return len(t.dependencies.Chained.Unassembled) > 0
+	return len(t.dependencyTracker.GetChainedDeps().GetUnassembledDependencies(t.pt.ID)) > 0
```
annamcallister: Adding `Has` functions to the tracker interface would avoid the copying needed to safely return these values, for them to only then have their length checked.

core/go/internal/sequencer/coordinator/transaction/transaction_test_utils.go
```diff
@@ -348,6 +359,110 @@ func (b *TransactionBuilderForTesting) NumberOfRequiredEndorsers(num int) *Trans
...
+			ct.queueEventForCoordinator = deliver
+		}
+	}
+}
+
+// WireCoordinatorQueueDeliveryFrom sets queueEventForCoordinator only on from, delivering transaction.Event
+// values to targets in byID in a new goroutine so from can release its state machine lock before the target
+// runs (the target may call getCoordinatorTransactionState on from and would deadlock on synchronous delivery).
+// The returned function blocks until all such deliveries started from this wiring have completed; call it
+// after from's HandleEvent returns before asserting on the target.
```
annamcallister: This feels concerning - if it is needed in tests, why wouldn't the same problems it avoids occur in real usage.

core/go/internal/sequencer/coordinator/transaction/transaction_test_utils.go
```diff
@@ -744,9 +844,9 @@ func (b *TransactionBuilderForTesting) Build() (*coordinatorTransaction, *transa
 	txn.revertCount = b.revertCount
 	txn.assembleErrorCount = b.assembleErrorCount
 
-	if b.dependencies != nil {
-		txn.dependencies = *b.dependencies
-	}
+	// if b.dependencies != nil {
+	// 	txn.dependencies = *b.dependencies
+	// }
```
annamcallister:
```suggestion
```

core/go/internal/sequencer/coordinator/coordinating.go
```diff
@@ -48,8 +48,14 @@ func action_TransactionsDelegated(ctx context.Context, c *coordinator, event com
...
+func (c *coordinator) newCoordinatorTransaction(ctx context.Context, originator string, originatorNode string, nodeName string, pt *components.PrivateTransaction, coordinatorSigningIdentity string) transaction.CoordinatorTransaction {
+	return transaction.NewTransaction(ctx, originator, originatorNode, nodeName, pt, coordinatorSigningIdentity, c.transportWriter, c.clock, c.queueEventInternal, func(ctx context.Context, id uuid.UUID) (transaction.State, bool) {
+		txn := c.transactionsByID[id]
+		if txn == nil {
+			return transaction.State(0), false
+		}
+		return txn.GetCurrentState(), true
+	}, c.engineIntegration, c.syncPoints, c.components, c.domainAPI, c.dCtx, c.requestTimeout, c.stateTimeout, c.closingGracePeriod, c.confirmedLockRetentionGracePeriod, c.baseLedgerRevertRetryThreshold, c.assembleErrorRetryThreshhold, c.grapher, c.dependencyTracker, c.metrics)
```
annamcallister: this function is difficult to read defined inline- could it be defined on the coordinator and passed in

core/go/internal/sequencer/coordinator/coordinating.go
```diff
@@ -309,11 +314,10 @@ func action_CleanUpTransaction(ctx context.Context, c *coordinator, event common
...
 	c.metrics.DecCoordinatingTransactions()
-	err := c.grapher.Forget(ctx, e.TransactionID)
-	if err != nil {
-		log.L(ctx).Errorf("error forgetting transaction %s: %v", e.TransactionID.String(), err)
-	}
-	c.chainedChildStore.ForgetChainedChild(e.TransactionID)
+	c.grapher.Forget(e.TransactionID)
+
+	// MRW TODO - Clean up the chained-child grapher instance here
+	// c.chainedChildStore.ForgetChainedChild(e.TransactionID)
```
annamcallister: I think this TODO actually ties back to my suggestion of a top level function to make the dependency tracker completely remove a transaction?

core/go/internal/sequencer/coordinator/coordinating.go
annamcallister: does this need to be handled via an event anymore or can it be part of the single `AddPrerequisites` call below?

core/go/internal/sequencer/coordinator/coordinating_test.go
```diff
@@ -276,9 +277,9 @@ func Test_action_CleanUpTransaction_GrapherForgetError_LogsButReturnsNil(t *test
 	txn, _ := transaction.NewTransactionBuilderForTesting(t, transaction.State_Confirmed).Build()
 	c.transactionsByID[txn.GetID()] = txn
 
-	mockGrapher := transaction.NewMockGrapher(t)
-	mockGrapher.EXPECT().Forget(mock.Anything, txn.GetID()).Return(fmt.Errorf("forget failed"))
-	c.grapher = mockGrapher
+	// mockGrapher := transaction.NewMockGrapher(t)
+	// mockGrapher.EXPECT().Forget(mock.Anything, txn.GetID()).Return(fmt.Errorf("forget failed"))
+	// c.grapher = mockGrapher
```
annamcallister:
```suggestion
```

annamcallister (changes requested):
This is a huge step forward in abstracting dependency tracking and moving to a thread safe model.

core/go/internal/sequencer/coordinator/dependencytracker/dependency_tracker.go
```diff
@@ -194,10 +213,8 @@ func (d *dependencyChain) ensure(id uuid.UUID) *nodeLinks {
 // AddPrerequisites records that txID depends on each prerequisite (prerequisite IDs must
 // complete before txID). Updates both sides of each edge.
 func (d *dependencyChain) AddPrerequisites(txID uuid.UUID, prereq ...uuid.UUID) {
-	if d.singleChainOnly {
-		if len(prereq) > 1 {
-			panic("singleChainOnly dependency chain does not support multiple prerequisites")
-		}
+	if d.singleChainOnly && len(prereq) > 1 {
+		panic("singleChainOnly dependency chain does not support multiple prerequisites")
```
matthew1001: I left this in for initial testing to make it super obvious if we were creating any chains incorrectly, but I think maybe I'd swap this for an error now

core/go/internal/sequencer/coordinator/dependencytracker/dependency_tracker.go
```diff
@@ -207,8 +224,17 @@ func (d *dependencyChain) AddPrerequisites(txID uuid.UUID, prereq ...uuid.UUID)
...
 		}
 		prereqNode := d.ensure(preReq)
-		tx.dependsOn = appendUnique(tx.dependsOn, preReq)
-		prereqNode.prereqOf = appendUnique(prereqNode.prereqOf, txID)
+		if d.singleChainOnly {
+			// TODO: this mirrors the behaviour of preassembly dependencies before this new grapher
+			// where a new dependency would overwrite the old one, if it existed. This doesn't quite
+			// feel right as in principle why should there be an old one? I'm choosing to preserve
+			// the old tested behaviour until we have time to think about this more.
+			tx.dependsOn = []uuid.UUID{preReq}
```
matthew1001: Perhaps we go with a check for any existing one, and report an error if that's the case before we apply the new prereq?

core/go/internal/sequencer/coordinator/dependencytracker/dependency_tracker.go
annamcallister: I've added a new SingleDependencyChain type which only accepts a single prereq parameter- i.e. enforce at compile time, not at runtime

core/go/internal/sequencer/coordinator/dependencytracker/dependency_tracker.go
annamcallister: I've added warning logs in.

annamcallister (approved):

Issue #937, opened 2025-12-04
bug: Refactor `common.Transaction` in `snapshot.go` as per comments in the linked review


matthew1001:
### What happened?

See PR comment https://github.com/LFDT-Paladin/paladin/pull/704#discussion_r2586283415 raised during review of the new sequencer

### What did you expect to happen?

N/A

### How can we reproduce it (as minimally and precisely as possible)?

N/A

### Anything else we need to know?

N/A

### OS version

N/A

PR #1061, opened 2026-03-04
Transaction interface cleanup

annamcallister:
Address https://github.com/LFDT-Paladin/paladin/issues/937 by making names snapshot specific.

Clean up use of external transaction functions by:
* Moving creation of transaction snapshot entries into the transaction package
* Making more use of the transaction builder set functions in test to set a known value rather than querying the transaction
* Removing test error count - this is redundant now all the error conditions it was checking result in a transition back to pooled
* Not using external functions in package tests

This smaller surface area is now expressed as an interface, which the coordinator uses rather than the struct type. This lays the basis for coordinator tests being able to use a mock transaction, which will allow us to more easily get test coverage of some branches, or to simplify some tests which have to layout a number of mocks for transaction behaviour.

This PR also cleans up some unused functions/parameters which have been causing linter warning


matthew1001 (approved):
LGTM

PR #1101, opened 2026-03-25
groupmgr: add pente_getCode and pente_getCodeHash to Pente

awrichar:
Note: this has been UPDATED from the original PR, which proposed `domain_getCode` and `domain_getCodeHash`. The new proposal (detailed below) adds methods within the Pente domain alone.

---

This adds a new pair of RPCs to domain manager and group manager:
- `domain_invokeRPC` invokes a domain-specific RPC on a particular private smart contract
- `pgroup_invokeRPC` is a convenience helper to invoke a domain-specific RPC on a privacy group contract

On top of this new construct, two new RPCs are added to the Pente domain:
- `pente_getCode` behaves similarly to `eth_getCode`, and returns the runtime bytecode for a private smart contract deployed at a given address.
- `pente_getCodeHash` returns only a keccak hash of the bytecode, for a lighter-weight call.

Both are RPC wrappers that directly map to Besu EVM calls in Pente.

I'm open to opinions on the prefix/spelling for these "inner" RPC methods on the Pente domain. I considered the following:
- `getCode`
- `eth_getCode`
- `evm_getCode`
- `pente_getCode` <-- chose this as an initial proposal

These should allow users to more easily authenticate contracts deployed to privacy groups, such as by comparing them to known code snapshots or hashes.


peterbroadhurst (changes requested):
@awrichar - I'm really in favor of adding this function, but I am concerned per the above of taxing all domains with a spelling that is very unique to Pente-like EVM Private Smart Contract domains.

core/go/internal/domainmgr/rpcmodule.go
```diff
@@ -124,3 +125,30 @@ func (dm *domainManager) rpcGetSmartContractByAddress() rpcserver.RPCHandler {
...
+		ctx = log.WithComponent(ctx, "domainmanager")
+		var sc components.DomainSmartContract
+		var err error
+		var resultJSON pldtypes.RawJSON
+		err = dm.persistence.Transaction(ctx, func(ctx context.Context, dbTX persistence.DBTX) error {
+			sc, err = dm.GetSmartContractByAddress(ctx, dbTX, address)
+			if err != nil {
+				return err
+			}
+			dCtx := dm.stateStore.NewDomainContext(ctx, sc.Domain(), address)
```
peterbroadhurst, line 149: I understand we haven't fully baked the spelling for all the different types of domain context, such as "state of the private blockchain at block X", but I still think we need to ensure we have a parameter for it in the stack

toolkit/proto/protos/service.proto
```diff
@@ -99,6 +101,8 @@ message DomainMessage {
     WrapPrivacyGroupEVMTXResponse       wrap_privacy_group_evmtx_res =        1191;
     CheckStateCompletionResponse        check_state_completion_res =          1201;
     IsBaseLedgerRevertRetryableResponse is_base_ledger_revert_retryable_res = 1211;
+    GetCodeHashResponse                 get_code_hash_res =                   1221;
+    GetCodeResponse                     get_code_res =                        1231;
```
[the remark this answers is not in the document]
peterbroadhurst: State qualifier seemed to be missing still @awrichar

peterbroadhurst (commented):
Given I'm confident we need a state qualifier, I really do think we need to have a placeholder.

Also slightly prefer `pente_getCode` to the other options, in case that's helpful in the spelling

core/go/internal/domainmgr/rpcmodule.go
awrichar: Alright - added a `stateQualifier`, but only "available" is accepted for now. Is this what you had in mind?

peterbroadhurst (approved):
Great, thanks @awrichar 

Very pleased we've put in the placeholder for when in the coming weeks we have the model for referring to "a point in previous or future time" refined a little to spell the other options on the string.

PR #1129, opened 2026-04-09
fix: re-assemble immediately on endorsement rejection

### LINKED ISSUE #1068: bug: A rejected endorsement should result in re-assembly without relying on a timeout
### What happened?

In cases where an endorser rejects the endorsement request the coordinator should ideally re-pool and (potentially, depending on other transactions) re-assemble immediately.

On the `v1-develop` branch it appears that re-assembly depends on a timeout in the state machine and doesn't drive re-assembly immediately

Re-assemble on endorsement rejection without timeouts being the forcing function

TBC

_No response_

### OS version

_No response_

Suvrat1629:
Fixes #1068 by adding `action_ResetEndorsementRequests` to the `Event_EndorsedRejected` transition.

Changes:
- Fix: Added reset action to `Event_EndorsedRejected` in the transaction state machine.
- Test: Added `TestEndorsedRejected_ImmediateRepool` to verify immediate state transition and request cleanup.

### DIFF
--- a/core/go/internal/sequencer/coordinator/transaction/endorsed_rejected_test.go
+++ b/core/go/internal/sequencer/coordinator/transaction/endorsed_rejected_test.go
@@ -1,0 +1,34 @@
+package transaction
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// Verifies that when an endorsement is rejected, the transaction resets pending endorsement requests
+// and transitions back to the pooled state so it can be re-assembled immediately.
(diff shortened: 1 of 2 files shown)


annamcallister:
This does not fix the bug. It simply resets the endorsement requests. 

Endorsement requests are already reset on entry to `State_Pooled` via this function https://github.com/LFDT-Paladin/paladin/blob/main/core/go/internal/sequencer/coordinator/transaction/pooled.go#L59, so this change is a no-op.

Looking at the code change, there has already been a transition to `State_Pooled` added to this event handler so I will also close the original issue.

PR #1134, opened 2026-04-09
Decrease memory retention

annamcallister:
Memory leak fixes
* Nil pooled transactions in the underlying array once popped for assembly. This transaction still can't be GCed until other pointers to it have been cleared but `c.pooledTransactions = c.pooledTransactions[1:]` was changing the slice but not the backing array causing a build up of old transactions (inc all their post assembly data) https://github.com/LFDT-Paladin/paladin/pull/1134/commits/fe4fc93c320f82d63dd693bbb0cadd599710d711
* Close domain contexts used for handling event batches https://github.com/LFDT-Paladin/paladin/pull/1134/commits/cda322d93471de9fb0fd836228a821693d5040fe

Changes to make Paladin for garbage collector friendly, by reducing unnecessary memory allocations/freeing large objects as soon as possible.

* Clear post assembly data once dispatched rather than retaining it for the time to confirm plus the grace period before deleting https://github.com/LFDT-Paladin/paladin/pull/1134/commits/6052fc773b150e89e71d552d2f4f3658840742a4
* Copy data received back from DB scans in RawJSON https://github.com/LFDT-Paladin/paladin/pull/1134/commits/f6ba8a834092730c6570dc83144418e58bb608ed

### DIFF
--- a/core/go/internal/sequencer/coordinator/transaction/state_machine.go
+++ b/core/go/internal/sequencer/coordinator/transaction/state_machine.go
@@ -388,6 +388,7 @@ var stateDefinitionsMap = StateDefinitions{
 	State_Dispatched: {
 		OnTransitionTo: []ActionRule{
 			{Action: action_NotifyDispatched},
+			{Action: action_ReleaseAssemblyPayload},
 		},
 		Events: map[EventType]EventHandler{
 			Event_Collected: {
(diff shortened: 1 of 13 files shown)


core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -388,6 +388,7 @@ var stateDefinitionsMap = StateDefinitions{
 	State_Dispatched: {
 		OnTransitionTo: []ActionRule{
 			{Action: action_NotifyDispatched},
+			{Action: action_ReleaseAssemblyPayload},
```
matthew1001: Would "clean" be a more suitable verb, more in keeping with `action_CleanupUpTransaction`? The current wording feels a little implementation-specific

matthew1001 (approved):

core/go/internal/sequencer/coordinator/transaction/state_machine.go
annamcallister: changed to clean up

PR #1138, opened 2026-04-10
add native macOS ARM build test job to CI 

### LINKED ISSUE #857: Investigate retagging images from release candidate/building on arm runners to speed to release process

Time boxed investigation into whether we can
* retagging images from release candidate for a release
* build macos images on arm runners

### Why is this needed?

The release candidate and release github actions are currently quite slow. Most of this time is spent building the macos image on amd64 runners meaning that emulation is required. We also rebuild the images for the release rather than retagging the release candidate images, even though in theory they should be identical.

Neither of these issues are trivial to solve, hence why they haven't already been solved. The slowness is also inconvenient but not broken. Hence why we first want a time boxed investigation into the feasibility of any quick wins in this space.

divyam-jha123:
closes #857 

 Description of changes in cross-build-images.yaml:
   - Added a new job: core-macos-arm-native
   - This job uses the macos-14-arm64 GitHub Actions runner to build the Docker image natively
   on Apple Silicon (ARM).
   - The job checks out the code and runs a native ARM Docker build for the main Dockerfile.
   - This allows direct comparison of build time and compatibility between emulated and native
   ARM macOS builds.


.github/workflows/cross-build-images.yaml
```diff
@@ -56,6 +56,15 @@ jobs:
       username: ${{ secrets.username }}
       password: ${{ secrets.password }}
 
+  core-macos-arm-native:
```
annamcallister, line 59: We're aiming to build the macos images that already get built in the job above `core-macos` on the arm runners. This change is adding a new job which will build images that do not get used in the release candidate process.

annamcallister (changes requested):

divyam-jha123:
Hi, @annamcallister as you told me that this will not used in the release candidate process. so, it is for just testing purpose.
so I made the change accordingly.
- Refactors that job to reuse the shared .github/workflows/build-image.yaml workflow instead of running a raw docker build inline.
- Tags the experimental image as ${sha}-arm64-native, with a note that it is not used by the RC flow.
- Marks the macOS ARM job as continue-on-error: true, so failures there won’t fail the whole workflow.
- Updates the shared build workflow so Linux-only steps like disk cleanup and QEMU setup are skipped on macOS runners.

I'm encountering an "Invalid workflow file" error due to `uses` and `with` being marked as unexpected values. I initially thought it might be related to `continue-on-error: true`, but it seems more like a YAML structure issue. Could you clarify the correct placement for this reusable workflow call within the job?

divyam-jha123:
reusable workflow issue fix - 

The root cause is continue-on-error: true on a job that calls a reusable
  workflow in .github/workflows/cross-build-images.yaml:59. GitHub does not allow
  that key there, so it stops treating the job as a reusable-workflow call and
  starts complaining that runs-on is missing.

  I fixed it by moving that behavior into the reusable workflow itself:

  - Added a boolean continue-on-error input and wired it to the job in .github/
    workflows/build-image.yaml:46 and .github/workflows/build-image.yaml:60
  - Passed that input from .github/workflows/cross-build-images.yaml:61

annamcallister:
@divyam-jha123 I think you've misunderstood. I didn't say this won't be used in the release candidate process. I said that your first change was to build an image which, while built in the release candidate job, would not get picked up by the overall package. The current naming of packages in important 

For us to be able to merge this PR we would need
* The existing job that builds the macos images to be switched to use a macos runner
* To see a release candidate that includes a macos image built in this way
* This release candidate to have been installed and tested

There are some challenges to building a release candidate from a fork because some steps involving pushing images (e.g. to DockerHub) in a way that requires credentials. I don't know if this can be worked around by commenting steps out, or if this is completely impossible- it's not something we've tried.

There is definitely value to comparing image build times on the different architectures side by side, and this build step is something, (possibly with a temporary action) that you should be able to do in your own fork. One thing to note is that somethings a build is able to run from cached steps, so you would need to be careful to ensure that you're only comparing uncached builds.

I think the steps I would recommend are
1. create whatever GHAs you need in your own fork to compare build times and share those times/links to the builds on this issue
2. assuming you have a mac to do this on, use the native built image in a local kind install (you can override an image in the helm charts) and run all the Paladin examples against it, to verify the image
3. modify the step that builds the release candidate macos image to use the native arm runner - this should be the only change that goes in a PR back to main. The image should have exactly the same tags, be built at the same point in the process etc, it is only the runner architecture that should be changing.

At this point we can help in verification with a release candidate build to do final package verification on.

Step 2 is crucial though- I don't think this task can be worked on effectively without a mac system to verify built images on.

divyam-jha123:
Hi @annamcallister,

I have completed Step 2 (local verification on a macOS ARM system) and wanted to share the results before I move to modify the step that builds the release candidate macos image to use the native arm runner and submit the final PR 

Verification Results:

 - Native ARM Build: The image was successfully built on an my Mac machine and loaded into a local kind cluster using the Paladin Helm charts. The core runtime and cluster installation are fully functional.

 - Example Testing: I ran the Paladin examples. While bond and swap are currently failing, I have verified that these are pre-existing issues unrelated to the ARM migration.

 - Root Cause Analysis: The failures return PD200007: Parameter 'unlock' is required. A git diff origin/main...HEAD confirms my changes are strictly limited to .github/workflows and do not touch the SDK, contracts, or Noto logic where this API mismatch exists.

divyam-jha123:
Validation on native ARM (macOS) using Kind cluster:

All Paladin pods are in ```Running``` state with expected readiness (2/2 for nodes)
No restarts observed across components
Pods have valid cluster IPs, confirming networking is functional
All pods are scheduled correctly on the Kind worker node

image

<br>
image

annamcallister:
@divyam-jha123 - sorry for the slowness here- I will be pulling this in to test next time I do a release candidate, I just haven't had time or the need to do this recently.

My expectation is that these changes should work though and we will be good to merge after that :)

annamcallister:
I've tried this today. I merged your changes into https://github.com/LFDT-Paladin/paladin/tree/test-macos-build so I could run them in the release candidate action. I then removed the continue on error option since there is no use of this action where it is acceptable for the build to fail.

The first attempt (https://github.com/LFDT-Paladin/paladin/actions/runs/26565848800) did not work as there is no `macos-14-arm64` runner - I changed it to `macos-latest` which is an arm runner https://docs.github.com/en/actions/reference/runners/github-hosted-runners https://github.com/LFDT-Paladin/paladin/commit/c95b92cc00258897ad979156337f60d297c5161d

The next attempt ran but failed the build https://github.com/LFDT-Paladin/paladin/actions/runs/26566317598

PR #1139, opened 2026-04-13
separate component tests

annamcallister:
Move coordination and component tests out of the main core test task and run in a separate GHA


core/go/build.gradle
```diff
@@ -433,14 +433,18 @@ task coordinationTestPostgres(type: CoordinationTest, dependsOn: [
     ]) {
 }
 
-task buildCoverageTxt(type: Exec, dependsOn: [
-        protoc,
-        unitTestSQLite,
-        unitTestPostgres,
+task coordinationAndComponentTest(dependsOn: [
```
awrichar: Should this have a simpler name - possibly just `componentTests`? Even though it doesn't run in the default build, it should be easy for devs to run it locally if they want.

matthew1001 (approved):
LGTM

core/go/build.gradle
annamcallister: Yes I've renamed it to `componentTest` - but left the job name with both. We need to do some restructuring so coordination tests are actually just a suite within component tests anyway

PR #1144, opened 2026-04-13
Navigation enhancements

gabriel-indik:
With the new transactions UX that was recently introduced, when clicking on a Paladin transaction the user would be taken to a dedicated panel. The URL of that panel was `/ui/transactions/<hash>/<id>` where
 - `hash` corresponds to the blockchain transaction
 - `id` corresponds to the paladin transaction

This has now been improved so that the path is: `/ui/transactions/<hash or id>`. In other words, if the user knows either the blockchain transaction hash or the Paladin transaction ID, then it is possible to link/navigate directly to it without needing to have the two pieces of information. The UI automatically determines if the value in the URL path corresponds to a blockchain transaction hash or a Paladin transaction ID and makes the necessary API calls to resolve the values.

Also included in this PR: the submissions panel has been brought back, making the "Pending" section the default. This panel has been updated so that when clicking on a Paladin transaction the user will be taken to the same details panel as the one used in the Transactions panel. Internal logic has been added so that if the user chooses to go back to the list, the UI determines whether the user accessed the details from the Transactions or the Submissions panel, and will direct the user to the right spot.


gabriel-indik:
Updates to the UI: there is the transactions view:

image

And the submissions view has been updated to show either `pending` or `failed` transactions, so there is no overlap with the transactions view:

image

gabriel-indik:
Submissions panel has been refactored to use pagination instead of infinite scrolling. Also a convenience lookup option was added there.

image

image

core/go/internal/txmgr/transaction_query.go
```diff
@@ -45,6 +45,7 @@ var transactionFilters = filters.FieldMap{
 	"from":           filters.StringField(`"from"`),
 	"to":             filters.HexBytesField(`"to"`),
 	"type":           filters.StringField(`"type"`),
+	"success":        filters.BooleanField(`"TransactionReceipt"."success"`),
```
awrichar: This filter relies on a join with `TransactionReceipt` (which is present when using `getTransactionFull`). What happens if I pass this filter to another RPC that doesn't join by default (i.e. `getTransaction` or `queryTransactions`)?

awrichar: I guess the filter is only on "query" methods - so only applies to `ptx_queryTransactions`. I just want to be sure it doesn't break in a bad way.

gabriel-indik: @awrichar great question. I tested the behavior to double check.
For the case of `queryTransactions` the following error is returned:
```
ERROR: missing FROM-clause entry for table \"TransactionReceipt\" (SQLSTATE)
```
For the case of `getTransaction` there is no query argument in the JSON RPC method.

awrichar: Thanks for testing. I think a comment in the code here would be helpful, noting that the parameter is silently ignored when performing a query that doesn't include the join on `TransactionReceipt`.

awrichar: Ah I see the update - if it's not silently ignored, I think we should ensure that the error is a little more clear (whether that's handling this specifically, or using separate filter maps for the endpoints that support this)

gabriel-indik: @awrichar I updated the code. Now if the user attempts to use the "success" field in a query where it doesn't apply, they'll get a clear message: 
```
"PD010700: Unknown field 'success'"
```

matthew1001:
@gabriel-indik tested some with-and-without receipt examples which all return an empty response array if the `success` filter is set (to either `true` or `false`) and the transaction exists but a receipt doesn't.

matthew1001 (approved):

PR #1146, opened 2026-04-14
Chained transaction ordering

annamcallister:
https://github.com/LFDT-Paladin/paladin/issues/1105

Persist the dependency graph from the original transaction on a chained transaction to ensure that a dependency cannot be assembled or dispatched ahead of its prereqs, resulting in inevitable on chain failure.

Chained dependencies are a new dependency type as they
* persist across the entire transaction lifecycle (like application dependencies, unlike post assembly dependencies)
* can be optimistically dispatched ahead as there is base ledger contract protection for the ordering (like post assembly dependencies, unlike application dependencies)

Because chained dependencies exist across the entire lifecycle of a transaction, this change has resulted in an explosion of additions to the transaction state machine, and pushes harder on the question of whether we should be able to define event handlers which operate across multiple states.


core/go/internal/sequencer/coordinator/transaction/chained_child_store.go
```diff
@@ -0,0 +1,62 @@
...
+	"sync"
+
+	"github.com/google/uuid"
+)
+
+// ChainedChildStore is a thread-safe store that maps parent transaction IDs to
+// their chained child transaction IDs. It has its own mutex, separate from any
+// individual transaction lock, so the dispatch loop can read/write without
+// risking deadlocks with the event loop's notification cascades.
+type ChainedChildStore interface {
```
annamcallister, line 28: I have intentionally kept this outside of the grapher for now, but I wonder if it wants to be part of the information held in the new grapher?

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -171,22 +252,33 @@ var stateDefinitionsMap = StateDefinitions{
 					{
 						To: State_Confirming_Dispatchable,
 						If: statemachine.GuardAnd(guard_AttestationPlanFulfilled, statemachine.GuardNot(guard_HasDependenciesNotReady)),
-					}},
+					},
+					{
```
annamcallister, line 256: This is a missing transition that has nothing to do with the rest of the changes in this PR

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -288,22 +443,52 @@ var stateDefinitionsMap = StateDefinitions{
 				Actions: []ActionRule{
 					{
 						Action: action_UpdateSigningIdentity,
-						If:     statemachine.GuardAnd(guard_AttestationPlanFulfilled, statemachine.GuardNot(guard_HasSigner)),
+						If:     statemachine.GuardNot(guard_HasSigner),
 					}},
 				Transitions: []Transition{{
 					To: State_Confirming_Dispatchable,
-					If: statemachine.GuardAnd(guard_AttestationPlanFulfilled, statemachine.GuardNot(guard_HasDependenciesNotReady)),
+					If: statemachine.GuardNot(guard_HasDependenciesNotReady),
```
annamcallister, line 450: Another unrelated fix- a transaction can only transition into this state if `guard_AttestationPlanFulfilled` is `true`

core/go/internal/sequencer/coordinator/transaction/transaction.go
```diff
@@ -38,10 +38,36 @@ type CoordinatorTransaction interface {
...
+
+type PostAssembleDependencies struct {
+	DependsOn []uuid.UUID
+	PrereqOf  []uuid.UUID
+}
+
+type ChainedDependencies struct {
+	DependsOn   []uuid.UUID
+	PrereqOf    []uuid.UUID
+	Unassembled map[uuid.UUID]struct{}
```
annamcallister, line 56: This is tracked rather than querying the dependency's state to avoid a deadlock on the transaction's lock

core/go/internal/components/transaction.go
```diff
@@ -40,7 +40,9 @@ type TransactionPreAssembly struct {
 	RequiredVerifiers        []*prototk.ResolveVerifierRequest `json:"required_verifiers"`
 	Verifiers                []*prototk.ResolvedVerifier       `json:"verifiers"`
 	PublicTxOptions          pldapi.PublicTxOptions            `json:"public_tx_options"`
-	Dependencies             *pldapi.TransactionDependencies   `json:"dependencies"`
+	// Chained dependencies: ordering constraints from the parent coordinator's grapher,
```
matthew1001: It feels important to be really clear what the characteristics of these chained dependencies are in this comment. With this PR I think we're persisting the chained dependencies, which means that a new coordinator (and hence fresh grapher) of the "parent" TX won't actually result in a change to this list? The comment mentions in-memory ordering, which again maybe just needs clarifying against the persisted ordering.

core/go/internal/sequencer/coordinator/transaction/assembling.go
```diff
@@ -130,23 +129,33 @@ func (t *coordinatorTransaction) nudgeAssembleRequest(ctx context.Context) error
 	return t.pendingAssembleRequest.Nudge(ctx)
 }
 
-func action_NotifyPreAssembleDependentOfSelection(ctx context.Context, txn *coordinatorTransaction, _ common.Event) error {
-	return txn.notifyPreAssembleDependentOfSelection(ctx)
+func action_NotifyDependentsOfSelection(ctx context.Context, txn *coordinatorTransaction, _ common.Event) error {
```
matthew1001, line 139: With the growing complexity of dependencies/pre-reqs it would be good to have a comment on this function and/or `notifyDependentsOfSelection` as to what this action represents. Is it "notify a dependent transaction that it's "parent" has progressed to the point where this "child" TX can be selected?

core/go/internal/sequencer/coordinator/transaction/assembling.go
```diff
@@ -130,23 +129,33 @@ func (t *coordinatorTransaction) nudgeAssembleRequest(ctx context.Context) error
...
 	}
-	return dependent.HandleEvent(ctx, &DependencySelectedForAssemblyEvent{
-		BaseCoordinatorEvent: BaseCoordinatorEvent{
-			TransactionID: t.pt.ID,
-		},
-	})
+	return nil
 }
 
 func (t *coordinatorTransaction) calculatePostAssembleDependencies(ctx context.Context) error {
```
matthew1001, line 168: I've removed this in https://github.com/LFDT-Paladin/paladin/pull/1054/changes#r3085334456 so suggest we defer discussion on how to track these dependencies, and then put a comment here to summarise the outcome of that discussion

core/go/internal/sequencer/coordinator/transaction/chained_child_store.go
matthew1001: I think in reading the comment of what it's trying to achieve and what https://github.com/LFDT-Paladin/paladin/pull/1054 is doing, it feels like we definitely don't want 2 independent solutions to track dependencies. The grapher in `1054` is doing two things, of which chained dependencies only need one (the TX dependency chain tracking), but I think it still makes sense to use a single package for that job

matthew1001: Useful in-person chat came to the following conclusion:
 - The new standalone `grapher` package in https://github.com/LFDT-Paladin/paladin/pull/1054 is probably the right shape to provide the tracking that `ChainedChildStore` currently provides
 - Due to timing constraints on getting the fixes in this PR in, while https://github.com/LFDT-Paladin/paladin/pull/1054 is finalised and reviewed, we'll merge this PR without the new grapher, then raise a standalone PR to swap out the dependency tracking component

core/go/internal/sequencer/coordinator/transaction/confirming.go
```diff
@@ -32,24 +33,34 @@ func guard_CanRetryRevert(ctx context.Context, txn *coordinatorTransaction) bool
...
-		t.revertReason = nil
-		t.decodedRevertReason = ""
-		t.revertOnChain = nil
-		t.lastCanRetryRevert = false
 	case *ConfirmedRevertedEvent:
 		hash = e.Hash
 		t.revertReason = e.RevertReason
-		t.revertCount++
-		if len(e.RevertReason) == 0 {
+		if len(e.RevertReason) == 0 && strings.HasPrefix(e.FailureMessage, "PD012256") {
```
matthew1001: This makes me nervous :) I think it's the first place we've got analysis of the revert reason baked into the sequencing logic. I'm not sure it's wrong - but just wanted to flag to check we're comfortable with this approach

core/go/internal/sequencer/coordinator/transaction/confirming.go
```diff
@@ -32,24 +33,34 @@ func guard_CanRetryRevert(ctx context.Context, txn *coordinatorTransaction) bool
...
+			//
+			// It is not a failure of this transaction itself, so we don't count it against the retry threshold.
+			// Instead we reassemble on new states.
+			t.decodedRevertReason = e.FailureMessage
+			t.lastCanRetryRevert = true
+		} else if len(e.RevertReason) == 0 {
 			// Off-chain revert (for example a chained assembly failure propagated from another TX).
 			// This is not a base-ledger revert, so do not route through domain retryability logic.
-			t.revertOnChain = nil
+			t.revertCount++
```
matthew1001: We don't need to increment this when `PD012256` because it's trumped by `lastCanRetryRevert == true`?

core/go/internal/sequencer/coordinator/transaction/confirming.go
```diff
@@ -63,10 +74,7 @@ func action_RecordConfirmation(ctx context.Context, t *coordinatorTransaction, e
 				// - stop decoding the revert reason in the domain so we don't have two decode place- but this might result in us
 				//   decoding fewer reverts, since transaction manager doesn't necessarily have each domain's ABI
 				t.decodedRevertReason = i18n.NewError(ctx, msgs.MsgTxMgrRevertedDecodedData, decodedReason).Error()
-			} else {
-				t.decodedRevertReason = ""
-			}
-			if decodedReason == "" && e.FailureMessage != "" {
+			} else if e.FailureMessage != "" {
```
matthew1001, line 71: `len(e.FailureMessage) != 0` for consistency with the rest of the function?

core/go/internal/sequencer/coordinator/transaction/grapher.go
```diff
@@ -95,41 +95,34 @@ func (s *grapher) Forget(transactionID uuid.UUID) error {
...
 //
 // Note - this mutates transaction dependency metadata; it does not mutate grapher indexes directly.
-func (s *grapher) pruneDependencyLinks(txn CoordinatorTransaction) {
-	// Remove this TX from all dependent forward links (dependent.DependsOn).
-	dependentIDs := make(map[uuid.UUID]struct{})
-	if txn.GetDependencies() != nil {
-		for _, dependentID := range txn.GetDependencies().PrereqOf {
-			dependentIDs[dependentID] = struct{}{}
-		}
+func (s *grapher) pruneDependencyLinks(ctx context.Context, txn CoordinatorTransaction) {
```
matthew1001, line 98: Intentionally ignoring most of this by virtue of https://github.com/LFDT-Paladin/paladin/pull/1054 in the pipeline

core/go/internal/sequencer/sequencer.go
```diff
@@ -508,6 +508,9 @@ func (sMgr *sequencerManager) handleTx(ctx context.Context, dbTX persistence.DBT
 		return i18n.NewError(ctx, msgs.MsgSequencerInternalError, "PreAssembly is nil")
 	}
 
+	// Set chained dependencies after InitTransaction, which replaces PreAssembly
```
matthew1001: Don't quite follow the comment here - perhaps could be clarified? It looks like `PreAssembly` isn't being replaced here as such, but updated?

core/go/internal/txmgr/transaction_query.go
```diff
@@ -151,6 +151,9 @@ func (tm *txManager) mapPersistedTXResolved(pt *persistedTransaction) *component
 	for _, dep := range pt.TransactionDeps {
 		res.DependsOn = append(res.DependsOn, dep.DependsOn)
 	}
+	for _, dep := range pt.TransactionChainedDeps {
```
matthew1001, line 154: I think this potentially means Node 1 sees the TX ID for a TX that doesn't exist on Node 1, but only on e.g. Node 2. Just clarifying where this leaves us wrt sequencing activity and the idea of "subject ID". I don't have a strong opinion where it leaves us - perhaps we're now saying we're comfortable exposing TX IDs from other nodes on any node that is part of that TX chain, but I think we should call that out explicitly as a change this PR introduces.

matthew1001: Or actually on re-reading this, I'm not sure if that's the case - will there be a DB entry on Node 1 in the case where Node 2 originated the chained TX? Basically would you mind clarifying what the UX looks like from the perspective of an RPC caller, and whether that experience differs based on whether the chained TX is originated in the same place as the "parent" TX :)

core/go/internal/sequencer/coordinator/transaction/transaction.go
```diff
@@ -249,5 +315,11 @@ func (t *coordinatorTransaction) GetPrivateTransaction() *components.PrivateTran
 func (t *coordinatorTransaction) GetDependencies() *pldapi.TransactionDependencies {
 	t.RLock()
 	defer t.RUnlock()
-	return t.dependencies
+	// if t.dependencies == nil {
+	// 	return nil
+	// }
```
matthew1001:
```suggestion
```

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -92,8 +93,20 @@ type (
 var stateDefinitionsMap = StateDefinitions{
 	State_Initial: {
 		Events: map[EventType]EventHandler{
+			// State_Initial only needs to handle Event_Delegated. The transaction is created and
```
matthew1001, line 96: Very general comment this PR pushes on. I think what this PR is starting to highlight is how complex it is for a contributor to understand what "correct" looks like for the state machine. We've had discussions around how we might simplify it, but that's just to simplify the code, it's not to remove valid states/transitions/actions from the actual state machine behaviour. I don't know that there's a realistic way to make the actual state machine logic more simple, but I struggle with how a new Paladin contributor would have confidence in changing the state machine without causing regressions. Test coverage is the main way we can help with that, and I think we're in a fairly good place in that regard. Perhaps when we come to generate docs from the state machine it will help.

matthew1001: Case in point - I am struggling to look at the state machine changes in this PR and conclude that they are "correct". I don't think a reviewer necessarily needs to be able to do that, but we do need to be conscious of the growing complexity.

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -404,7 +653,7 @@ var stateDefinitionsMap = StateDefinitions{
 				Actions: []ActionRule{
 					{Action: action_RecordConfirmation},
 					{Action: action_NotifyOriginatorOfConfirmation},
-					{Action: action_NotifyDependantsOfSuccessfulConfirmation},
+					{Action: action_ResetLocksOnConfirmationIfNoRetentionGracePeriod},
```
matthew1001: I wonder if we should have a rule that actions are never worded in a conditional way. Should this not be `action_ResetLocksOnConfirmation` with an appropriate guard?

matthew1001: Coming to agreement on such rules might help in part with my lengthy comment about state machine complexity. Although it definitely won't outright solve it

core/go/internal/sequencer/coordinator/transaction/assembling.go
annamcallister: I think this was pure refactoring due to the relocation of the post assemble dependencies within the dependencies object in this PR - I think it becomes entirely superseded by #1054 - i.e. there is nothing to preserve from my changes to this function once merged

core/go/internal/sequencer/coordinator/transaction/confirming.go
annamcallister: I haven't incremented it as I didn't want failure due to the failure of an upstream dependency to count against the retry threshold as it didn't feel like a legitimate failure for this transaction, but this might be worthy of revisiting since it is maybe no different to a transaction reaching its retry threshold because it was repeatedly assembled on the outputs of other failing transactions.

Having seen this bypass of the limit in action I am starting to feel more unsure about leaving a potential unbounded retry path for a transaction.

core/go/internal/sequencer/coordinator/transaction/confirming.go
annamcallister: it's `len(e.RevertReason)` because revert reason is a byte array, where as failure message is a string. I don't think it makes more sense to use the more idiomatic empty check for each type than to keep them consistent with each other

core/go/internal/sequencer/sequencer.go
annamcallister: There is an earlier point in code where a DB read fully overwrites anything already set in PreAssembly.

I will try and make the comment clearer- possibly it is not needed at all, as we don't routinely comment every alternative code approach we could have taken but which wouldn't actually work 🙃

core/go/internal/txmgr/transaction_query.go
annamcallister: I think there's an assumption baked in here that all chained dependencies will be dispatched on the same node (that of the coordinator currently coordinating the original transactions). It would take a lot of different changes across the sequencer and domains for this to not be true. Even with sequencer handover I think we'd expect to see one node fully flush it's current graph of optimistic state dependencies before handing over to another node.

So any chained dependencies listed in here must be originated on the same node as this transaction is being originated on.

core/go/internal/sequencer/coordinator/transaction/transaction.go
annamcallister: I need to come back to this function, will remove this, but I also don't think we should be using the pldapi dependencies type inside the sequencer.

core/go/internal/sequencer/coordinator/transaction/state_machine.go
annamcallister: Agree with everything you're saying 👍

core/go/internal/sequencer/coordinator/transaction/state_machine.go
annamcallister: Agree- I've been trying to make actions as unconditional as possible (in their implementation, not just through the naming) - this one has slipped through the net. 

Use of separate guards/validators will make any generated docs for the state machine as useful as possible

core/go/internal/sequencer/coordinator/transaction/confirming.go
annamcallister: I _think_ it's ok because
* PD error codes should not change- we consider non additive changes to them to be breaking changes to our API
* A transaction receipt is the way that we report back the result of a chained transaction to the original transaction - I don't see that we have any other suitable vehicle to do this. We've added logic to the domains where they can interpret their own revert reasons. This feels roughly equivalent provided that failure messages interpreted by a sequencer are always created by a sequencer (and not, for example, by a domain contract)

matthew1001: Agreed, probably just useful to remember in any coding guidelines for contributors we put together for V1

core/go/internal/sequencer/coordinator/transaction/confirming.go
matthew1001: So just to check, you're potentially going to change this to increment the revert count?

core/go/internal/sequencer/coordinator/transaction/confirming.go
matthew1001: Fair enough

matthew1001 (approved):
Thanks for the lengthy in-person discussion. This nuanced case (that actually gets hit surprisingly often) could really do with docs update for V1 at some point.

Issue #1065, opened 2026-03-06
bug: Some JSON/RPC errors are returned without their error codes


matthew1001:
### What happened?

An example given when running on `v1-develop` was that this error:

```
"PD012616", "Domain reverted transaction on assemble"
```

returned only "Domain reverted transaction on assemble" to the caller. It should always include the error code

### What did you expect to happen?

Include the error code in JSON/RPC responses

### How can we reproduce it (as minimally and precisely as possible)?

TBC

shorsher:
Hey @matthew1001, I'm looking at picking up this issue. Could you assign it to me?

Roniscend:
Hey @matthew1001 I'd like to work on this issue if its available may i take it up?

PR #1147, opened 2026-04-14
Return HTTP 200 for proper jsonrpc responses

### LINKED ISSUE #1001: bug: idempotency conflicts return an HTTP 500
### What happened?

Idempotency conflicts return an HTTP 500 response code with the body giving Paladin error PD012220.

The HTTP response code should be 409 to indicate a conflict.

Make two requests with the same idempotency key.

_No response_

### OS version

_No response_

shorsher:
Previously, most jsonrpc errors would result in a HTTP 500 status code, e.g. an idempotency conflict. This commit changes that behavior to better align the with the JSON/RPC spec. Now, anytime the rpc response contains a proper JSON/RPC response body, the HTTP error code will be 200. Only authentication/authorization issues will result a 401/403 HTTP code.

resolves https://github.com/LFDT-Paladin/paladin/issues/1001

### DIFF
--- a/toolkit/go/pkg/rpcserver/rpchandler.go
+++ b/toolkit/go/pkg/rpcserver/rpchandler.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"io"
+	"sync"
 	"time"
 	"unicode"
 
@@ -31,10 +32,10 @@ import (
 )
 
 type handlerResult struct {
-	sendRes bool
-	isOK    bool
-	res     any
-	postSend func()
+	sendRes      bool
+	res          any
+	postSend     func()
+	unauthorized bool
 }
 
 func (s *rpcServer) rpcHandler(ctx context.Context, r io.Reader, wsc *webSocketConnection) handlerResult {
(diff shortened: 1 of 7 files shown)


annamcallister:
This feels more like a fix for #1001 than #1065?

There was some (probably offline and not recorded in GH) discussion that this switch could be fairly disruptive for consumers, since many HTTP clients handle 500 and 200 responses very differently. and we should maybe have a config option (in `rpcserver`?) which allows you to turn the old behaviour back on. The new behaviour should be the V1 default.

It doesn't have to be within this PR but I think #1001 also needs us to make some improvements in the RPC error codes we support. https://github.com/LFDT-Paladin/paladin/blob/main/sdk/go/pkg/rpcclient/client.go#L45 is a start at custom Paladin error codes, we can continue to add more in this space, where I think a new code for an idempotency error is the first candidate, but it would be worth a review of everywhere else in the code we are returning JSON RPC errors to see if there are any other obvious categories to formalize.

shorsher:
Ah yes you're right, I linked to the wrong issue, thank you! I'm happy to split out the legacy behavior changes into a separate issue/PR.

PR #1148, opened 2026-04-15
Cleanup of ILockableCapability and Noto V2

awrichar:
This is a proposal to clean up a few items in the `ILockableCapability` interface to make it easier to standardize. As a result of the changed method signatures, this also represents Noto V2.

Changes to `ILockableCapability`:
- Removed `content` and `options` from struct definitions.
  - Both of these are implementation-specific `bytes` and therefore shouldn't be required to be represented in storage.
  - ~~Added a `getLockContent` getter to retrieve the lock content without specifying how it's stored (although I'm on the fence as to whether this is useful/critical as part of the standard).~~
  - These are moved into implementation-specific inputs, as there's no reason for them to exist on the standard.
- Renamed `spendHash`/`cancelHash` to `spendCommitment`/`cancelCommitment`.
  - No change in functionality, but makes the standard less opinionated on whether they must be a hash.
- Added a separate `LockCreated` event.
  - This feels aligned with the rest of the events in the contract (rather than reusing `LockUpdated` for create+update).
  - This is also aligned with the custom `NotoLockCreated`/`NotoLockUpdated` methods in Noto.
- Tightened restrictions on `updateLock` so it may not be called at all after lock is delegated.
  - Mutability was very hazy before (partly due to the removed `options`), so this makes it simpler and clearer.
  - This does not prevent implementers from adding separate methods that can change implementation-specific options based on implementation-specific mutability constraints (as long as the base constraints hold).
- Standardized on "args" terminology for the implementation-specific inputs to each method (`createArgs`, `updateArgs`, etc).
  - Previously we were inconsistent with various terms including "inputs", "operations", and "params".
- Further clarified method docs and mandatory behaviors of the standard.

---

Although the above is the _reason_ for the change, the bulk of the code changes are in Noto to ensure we properly support V0/V1/V2. `createLock` and `updateLock` method signatures were changed, and require V1/V2 parsing to encode and invoke correctly. Numerous other things were renamed for consistency, but do not affect any encoding/decoding.

Although V0/V1/V2 will all continue to be supported, there is no seamless upgrade path for a Noto contract in place (due to slight differences in the storage layout). Careful migration from V1 to V2 would be possible by crafting a version of the contract with fallback logic for both storage layouts - but I'm not building this in upfront.


jimthematrix (approved):
LGTM! the new interfaces do feel more streamlined 👍

solidity/contracts/domains/noto/Noto.sol
```diff
@@ -362,48 +408,56 @@ contract Noto is EIP712Upgradeable, UUPSUpgradeable, INoto, INotoErrors {
...
         if (lock.owner != address(0)) {
             revert NotoDuplicateLock(lockId);
         }
 
-        // Set the initial ownership, and immutable contents of the lock
+        // Set the initial ownership, and immutable content of the lock
         lock.owner = msg.sender;
         lock.spender = msg.sender;
-        lock.content = abi.encode(lockOp.contents);
+        lock.content = abi.encode(args.content);
```
jimthematrix, line 439: the `content` captures the locked UTXOs, but its only usage seems to be at line 694:

```javascript
bytes32[] memory lockContent = abi.decode(lock.content, (bytes32[]));
```

while on spend or cancel, the inputs are supplied again as part of the args. it seems this storage cost can be saved by not persisting the lock content, without any loss of security because the entire spend/cancel operation is protected by the spendCommitment/cancelCommitment.

solidity/contracts/domains/interfaces/ILockableCapability.sol
```diff
@@ -78,26 +107,29 @@ interface ILockableCapability {
         bytes32 indexed lockId,
         address indexed from,
         address indexed to,
```
peterbroadhurst: As I read it fresh end to end, somehow `from` and `to` stick out as out of strange here.

Maybe @awrichar have a think about this?

Some ideas:
- `owner` `delegate`
- `previousSpender` `newSpender`

awrichar: `previousSpender` and `newSpender` were the other terms I considered (it's possible that neither is the owner in the case of re-delegation). I could switch over.

solidity/contracts/domains/noto/Noto.sol
awrichar: I included a `getLockContent` as part of the `ILockableCapability` interface, so the storage is to support that (line 266). I felt like it was important to be able to inspect/verify the states that were locked. But coming back to it now, it's probably much more important to be able to validate the spend/cancel states, which as you say are already tied to the spendCommitment/cancelCommitment.

I need to think a little more about whether `getLockContent` is needed...

awrichar: I've decided that `getLockContent` is no longer providing any value. I'll remove it from the interface.

PR #1150, opened 2026-04-15
Long running tests

annamcallister:
Renames project from `perf` to `test`- this might want to become more specific in future.

Adds two new test types
- high load of successful pente transactions while periodically killing nodes
- repeated noto transfers through a hooks contract with periodic forced failures


operator/charts/paladin-operator/templates/nodes/besugenesis.yaml
```diff
@@ -11,8 +11,8 @@ spec:
   chainID: 1337
   gasLimit: 700000000
   consensus: qbft
-  blockPeriod: 100ms
-  emptyBlockPeriod: 10s
+  blockPeriod: 1s
```
matthew1001, line 14: I'll probably merge this @annamcallister but adding a comment here to check the reason for this change. Was it incorrect before? Also wonder if we should just remove empty block period if we're not using it as it should default to `off` if it's unset

matthew1001 (approved):

operator/charts/paladin-operator/templates/nodes/besugenesis.yaml
annamcallister: no this is a mistake- I will restore this change. Thanks for spotting!

PR #1156, opened 2026-04-17
Ensure dialogs are not closed on tab switch

gabriel-indik:
Issue has been raised. Specifically in the domains panel, when switching to another browser tab and then coming back, any dialog box that had been left open is closed. This is due to the way the screen is re-rendered upon reloading the information from the API call.


matthew1001 (approved):
LGTM

PR #1160, opened 2026-04-21
Feat/noto transaction requester tracking

### LINKED ISSUE #825: Noto should track transaction requester in domain receipt

It's useful to know which party requested a specific transaction. Noto currently does not expose this in its domain receipt, but it should be easy to support. 

Noto tracks an info state called "TransactionData". During assembly, it can record the transaction "from" identity (maybe the lookup _and_ the resolved address) into this structure. Then it will be available when building the domain receipt, and new fields can be added to the domain receipt to expose it there.

### Why is this needed?

This allows any participant in a Noto transaction to know who initiated the transaction. It still preserves privacy since the info state is only distributed to those that need it, but it gives them more context on the transaction.

divyam-jha123:
closes #825 

🎉 Transaction Requester Tracking - Complete

Changes Made:

1. Transaction Data Structure (pkg/types/config.go)

 - Added From field to track requester identity lookup
 - Added FromAddress field to store resolved Ethereum address
 - Updated NotoTransactionDataABI_V1 to include the new fields

2. Receipt Types (pkg/types/states.go)

 - Created ReceiptRequester struct with:
  - Lookup: Transaction requester identity lookup
  - Address: Resolved Ethereum address
 - Added Requester field to NotoDomainReceipt

3. Encoding Logic (internal/noto/noto.go)

 - Updated encodeTransactionData() to accept verifiers parameter
 - Enhanced encodeTransactionDataV1() to:
  - Capture transaction From identity
  - Resolve the address using findEthAddressVerifier()
  - Encode both lookup and address in transaction data

4. Receipt Building (internal/noto/receipts.go)

 - Updated BuildReceipt() to:
  - Decode requester from TransactionData
  - Populate Requester field in domain receipt
  - Expose to all transaction participants

5. Handler Updates

 - Updated all handler calls to pass req.ResolvedVerifiers:
  - handler_mint.go
  - handler_transfer_common.go
  - handler_burn_common.go
  - handler_lock.go
  - handler_delegate_lock.go
  - handler_unlock_common.go
  - handler_prepare_unlock.go
  - handler_unlock.go
  - Lock creation handlers (mint, burn, transfer)

6. Test Support

 - Updated test helpers with nil verifiers parameter
 
 <b>Key Features:</b>

✅ Privacy Preserving - Info state distributed only to needed parties 
✅ Backward Compatible-Optional fields in both data and receipt structures 
✅ Complete - All handlers, types, and receipt generation updated 
✅ Signed - Commit signed with developer credentials


.github/workflows/compare-image-build-times.yaml
```diff
@@ -0,0 +1,141 @@
+name: Compare Image Build Times
```
awrichar: This new workflow seems unrelated to the PR - maybe this should be a separate proposal?

docker_build.log
awrichar, line 1: This log file should not be committed.

scripts/use-local-dev-env.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
awrichar: Not sure this script should be committed.

divyam-jha123:
hey @awrichar you are right [scripts/use-local-dev-env.sh](https://github.com/LFDT-Paladin/paladin/pull/1160/files/e67b85519a33ed3f87b3e4c925c036e93e6d96d6#diff-ac27a5bbc0ee3460e0bfed2b303b066569104261d8592fb3b1c4f5ab82df493c) should not be commited bcz this is a shell script snippet for setting up a local development environment for the project
I have made the cleanup based on your feedback

.github/workflows/release-candidate.yaml
```diff
@@ -171,7 +171,7 @@ jobs:
         git config --local user.email "action@github.com"
         git config --local user.name "GitHub Action"
         git add examples/
-        git commit -m "chore: Update SDK version in examples to ${{ needs.setup.outputs.tag }}"
+        git commit -s -m "chore: Update SDK version in examples to ${{ needs.setup.outputs.tag }}"
```
awrichar: I don't think the GitHub release workflows should be changed in this PR. If there's a concern or enhancement here, it should be a separate PR.

domains/noto/internal/noto/noto.go
```diff
@@ -1067,8 +1067,24 @@ func (n *Noto) encodeTransactionDataV1(ctx context.Context, infoStates []*protot
 		}
 	}
 
+	// Extract the requester identity and resolved address
+	var from *string
+	var fromAddress *pldtypes.EthAddress
+	if transaction != nil && transaction.From != "" {
```
awrichar: The domain can rely on these to be populated (it's part of the interface contract... Paladin won't provide an invalid or nil transaction)

domains/noto/internal/noto/noto.go
```diff
@@ -1067,8 +1067,24 @@ func (n *Noto) encodeTransactionDataV1(ctx context.Context, infoStates []*protot
...
+				fromAddress = fromAddr.address
+			}
+		}
+	}
+
 	dataValues := &types.NotoTransactionData_V1{
-		InfoStates: stateIDs,
+		InfoStates:  stateIDs,
+		From:        from,
+		FromAddress: fromAddress,
```
awrichar:
```suggestion
		From:        fromAddress,
```

I know that the original write-up suggested passing both the lookup and the resolved address, but after discussion with other maintainers, we landed on just passing the resolved address. In all other cases (including the "manifest" info state and the encoded details within each "coin" state) we always use the resolved address to identify the parties, and never directly include the local identity lookup string in what is sent to other nodes. Hence, we'd like to recommend only including "from" as the resolved sender address (very similar to Ethereum `msg.sender`), and omitting the lookup string.

awrichar (changes requested):
Thanks for this - I've requested a few more changes, and it looks like the unit tests are failing as well. We'd love to get this in an upcoming Paladin release if you're able to fix up these remaining items.

divyam-jha123:
@awrichar  Thanks for the review — I've updated the PR to:

1. Remove the release workflow changes from this PR (fixed in pr #1200)
2. Drop the redundant nil check on `transaction` (relying on the interface contract).
3. Use only the resolved address for `from` in `NotoTransactionData_V1` and `requester.from` on receipts (no lookup string on the wire).
4. Record the requester on `TransactionData` info states at assembly so receipts can expose it reliably.

domains/noto/pkg/types/states.go
```diff
@@ -25,6 +25,7 @@ type NotoDomainReceipt struct {
 	Transfers []*ReceiptTransfer `json:"transfers,omitempty"`
 	LockInfo  *ReceiptLockInfo   `json:"lockInfo,omitempty"`
 	Data      pldtypes.HexBytes  `json:"data,omitempty"`
+	Requester *ReceiptRequester  `json:"requester,omitempty"`
```
awrichar: Since we only have one field now, I think there's no reason to nest it. I would suggest `Sender` here (to align with Ethereum's `msg.sender`) and it can directly be an `*EthAddress`, rather than a nested struct.

domains/noto/pkg/types/config.go
```diff
@@ -78,7 +78,8 @@ type NotoTransactionData_V0 struct {
 }
 
 type NotoTransactionData_V1 struct {
-	InfoStates []pldtypes.Bytes32 `json:"infoStates"`
+	InfoStates []pldtypes.Bytes32   `json:"infoStates"`
+	From       *pldtypes.EthAddress `json:"from,omitempty"` // Resolved Ethereum address of the transaction requester
```
awrichar: This is the data structure that gets written to the base blockchain. We should _not_ expose the sender address here. It should only be inside one of the info states, and the ref to that info state will be in the InfoStates list. This way we only record the sender in the private data and don't expose it on the chain.

domains/noto/pkg/types/states.go
```diff
@@ -225,5 +231,6 @@ var TransactionDataABI_V1 = &abi.Parameter{
 		{Name: "salt", Type: "bytes32"},
 		{Name: "data", Type: "bytes"},
 		{Name: "variant", Type: "uint64"},
+		{Name: "from", Type: "address"},
```
awrichar, line 244: We will need a `TransactionData_V2` state. ABI encode/decode doesn't work with optional fields, so we must create a new version of the state data for each schema change.

We can still share the `TransactionData` struct above in the Go layer, but the ABI must be differentiated.

domains/noto/internal/noto/handler_burn_common.go
```diff
@@ -195,7 +195,7 @@ func (h *burnCommon) baseLedgerInvokeBurn(ctx context.Context, tx *types.ParsedT
 		return nil, i18n.NewError(ctx, msgs.MsgAttestationNotFound, "sender")
 	}
 
-	data, err := h.noto.encodeTransactionData(ctx, tx.DomainConfig, req.Transaction, req.InfoStates)
+	data, err := h.noto.encodeTransactionData(ctx, tx.DomainConfig, req.Transaction, req.InfoStates, req.ResolvedVerifiers)
```
awrichar: We do not need to encode the sender here. It's encoded in the private info states already up on line 95, which is the correct place. (currently you're encoding it in two different places)

awrichar (changes requested):
I noticed a few more important things - thanks for working this through. I think some tests are still failing as well, so please double-check.

divyam-jha123:
i have made the suggested changes and also ran test suite in `domains/noto` without caching to ensure all state transitions, receipt construction, and schemas are correctly validated: All unit tests build successfully and pass.


image

domains/noto/internal/noto/noto.go
```diff
@@ -475,7 +477,10 @@ func (n *Noto) LockInfoSchemaID() string {
 }
 
 func (n *Noto) DataSchemaID() string {
-	return n.dataSchemaV1.Id
+	if n.dataSchemaV2 == nil {
+		return n.dataSchemaV1.Id
+	}
+	return n.dataSchemaV2.Id
```
awrichar, line 480: We should be able to always assume `n.dataSchemaV2` is set. We pass a list of schemas during domain init, and are guaranteed to get back IDs for all of them (line 522).

domains/noto/internal/noto/states.go
```diff
@@ -190,8 +190,12 @@ func (n *Noto) makeNewInfoState(info *types.TransactionData, variant pldtypes.He
 			DistributionList: distributionList,
 		}, nil
 	}
+	schemaID := n.dataSchemaV1.Id
+	if n.dataSchemaV2 != nil {
```
awrichar: I think we can assume it's non nil (both schemas are defined in the same function call, so it's not like we can get into a situation where one is defined and the other isn't). We should just always encode V2 going forward.

domains/noto/internal/noto/states.go
```diff
@@ -451,12 +455,18 @@ func (n *Noto) prepareLockedOutputs(id pldtypes.Bytes32, owner *identityPair, am
...
 }
 
-func (n *Noto) prepareDataInfo(data pldtypes.HexBytes, variant pldtypes.HexUint64, distributionList []string) ([]*prototk.NewState, error) {
+func (n *Noto) prepareDataInfo(ctx context.Context, data pldtypes.HexBytes, variant pldtypes.HexUint64, distributionList []string, transaction *prototk.TransactionSpecification, verifiers []*prototk.ResolvedVerifier) ([]*prototk.NewState, error) {
 	newData := &types.TransactionData{
 		Salt:    pldtypes.RandBytes32(),
 		Data:    data,
 		Variant: variant,
 	}
+	if variant == types.NotoVariantDefault {
```
awrichar: Not sure why this is gated by variant?

domains/noto/internal/noto/receipts.go
```diff
@@ -31,7 +31,11 @@ import (
 func (n *Noto) BuildReceipt(ctx context.Context, req *prototk.BuildReceiptRequest) (res *prototk.BuildReceiptResponse, err error) {
 	receipt := &types.NotoDomainReceipt{}
 
-	infoStates := n.filterSchema(req.InfoStates, []string{n.dataSchemaV0.Id, n.dataSchemaV1.Id})
+	schemas := []string{n.dataSchemaV0.Id, n.dataSchemaV1.Id}
+	if n.dataSchemaV2 != nil {
```
awrichar: No nil check needed; all of these schemas will coexist

domains/noto/internal/noto/handler_unlock_common.go
```diff
@@ -356,7 +356,7 @@ func (h *unlockCommon) endorse(
 	}, nil
 }
 
-func (h *unlockCommon) buildPrepareUnlockParams(ctx context.Context, tx *types.ParsedTransaction, lt *lockTransition, proof pldtypes.HexBytes, lockedInputs, spendOutputs, cancelOutputs, infoStates []*prototk.EndorsableState) (_ *UpdateLockParams, err error) {
+func (h *unlockCommon) buildPrepareUnlockParams(ctx context.Context, tx *types.ParsedTransaction, lt *lockTransition, proof pldtypes.HexBytes, lockedInputs, spendOutputs, cancelOutputs, infoStates []*prototk.EndorsableState, verifiers []*prototk.ResolvedVerifier) (_ *UpdateLockParams, err error) {
```
awrichar: Is the new `verifiers` param used?

domains/noto/internal/noto/handler_unlock_common.go
```diff
@@ -400,7 +400,7 @@ func (h *unlockCommon) buildPrepareUnlockParams(ctx context.Context, tx *types.P
 
 }
 
-func (h *unlockCommon) buildCreateLockParams(ctx context.Context, tx *types.ParsedTransaction, lockTransition *lockTransition, proof pldtypes.HexBytes, inputs, lockedOutputs, additionalOutputs, spendOutputs, cancelOutputs, infoStates []*prototk.EndorsableState) (_ *CreateLockParams, err error) {
+func (h *unlockCommon) buildCreateLockParams(ctx context.Context, tx *types.ParsedTransaction, lockTransition *lockTransition, proof pldtypes.HexBytes, inputs, lockedOutputs, additionalOutputs, spendOutputs, cancelOutputs, infoStates []*prototk.EndorsableState, verifiers []*prototk.ResolvedVerifier) (_ *CreateLockParams, err error) {
```
awrichar: Is the new `verifiers` param used?

domains/noto/internal/noto/noto.go
```diff
@@ -1057,7 +1064,7 @@ func (n *Noto) encodeTransactionDataV0(ctx context.Context, transaction *prototk
 	return data, nil
 }
 
-func (n *Noto) encodeTransactionDataV1(ctx context.Context, infoStates []*prototk.EndorsableState) (pldtypes.HexBytes, error) {
+func (n *Noto) encodeTransactionDataV1(ctx context.Context, transaction *prototk.TransactionSpecification, infoStates []*prototk.EndorsableState) (pldtypes.HexBytes, error) {
```
awrichar: `transaction` param isn't used now

awrichar (changes requested):
I think the functionality is now correct - just a few cleanup items left due to iterating on the code.

domains/noto/internal/noto/states.go
divyam-jha123: The `NotoVariantDefault` check was leftover from when we were deciding when to populate `from`. Since we always encode V2 for non-legacy data states (which include from), the gate wasn’t needed - we now resolve `from` from verifiers whenever it’s available. Legacy still uses the V0 schema path in `makeNewInfoState`, which doesn’t include that field.

divyam-jha123:
hey @awrichar made the suggested cleanups,just wanted to ask why the build is failing on the main branch itself.
image
I thought that is because of the go version on my machine hence made an issue regarding that (#1201).
Do you have any suggestions regarding that

awrichar (approved):
Looks great. Thanks for the contribution!

awrichar:
As for the local build failure, it does seem potentially related to Go version. I think it can be triaged on the issue you opened.

Issue #1143, opened 2026-04-13
Coordinator selection based on block number


matthew1001:
### What would you like to be added?

The distributed sequencer algorithm uses a deterministic function for all nodes to arrive at a consistent view of who the current coordinator for a given domain instance should be.

This allows distributed nodes to efficiently agree on which node will perform the role of coordinator, allowing other nodes to rely on that node to coordinate new or pending transactions.

The current algorithm is deterministic but not based on current block height. It should factor in the current block height, and ensure that if the expected coordinator is offline or unavailable, another (deterministically selected) node takes over as coordinator.

### Why is this needed?

To ensure efficient and highly available coordinator of domain instance transactions, particularly for post-V1.0 domains that do not require all endorsers to be available for a transaction to be successful.

oindrilakha12-ui:
Hi @matthew1001, I’d like to work on this issue. Please let me know if I can proceed. Thanks!

sagarkhandagre998:
@matthew1001 Kindly take a look - https://github.com/LFDT-Paladin/paladin/pull/1161

PR #1161, opened 2026-04-22
feat(sequencer): add coordinator failover for endorser mode (#1143)

sagarkhandagre998:
feature# - https://github.com/LFDT-Paladin/paladin/issues/1143

Coordinator Failover for Endorser Mode (`issue-1143`)

### Problem

When the currently-observed coordinator node stopped heartbeating, the sequencer would immediately transition to `State_Idle` after `inactiveToIdleGracePeriod` intervals — even if other healthy nodes in the originator pool could take over. This caused in-flight transactions to stall unnecessarily until the original coordinator came back online.

### Solution

Introduce a `coordinatorFailoverIndex` field on the coordinator state machine. When the heartbeat threshold is exceeded in `State_Observing`, instead of going idle immediately, the system increments the failover index and deterministically selects the next candidate from the sorted originator pool using the same hash-modulus logic as primary selection. Since all nodes share the same sorted pool and block range, they independently arrive at the same failover candidate — no additional coordination is needed.

The failover index is reset to `0` in three cases:
- A valid heartbeat is received (`action_HeartbeatReceived`) — primary is back
- A block-range boundary is crossed (`action_NewBlock`) — new epoch, start fresh
- The coordinator enters idle (`action_Idle`) — pool fully exhausted

`OnNewBlockHeight` in the sequencer manager now broadcasts a `NewBlockEvent` to all loaded coordinators via `TryQueueEvent`, enabling `action_ReevaluateCoordinatorOnNewBlock` to re-select the primary when the range resets.

Only `COORDINATOR_ENDORSER` mode supports multi-node failover. `COORDINATOR_STATIC` and `COORDINATOR_SENDER` modes clear `activeCoordinatorNode` and go idle immediately as before.

Tests

**Unit tests** (`internal/sequencer/coordinator/`) — 13 packages, all passing:
- Failover index selection with index 0/1/2 and wrap-around
- `action_AttemptCoordinatorFailover` for single-node pool, multi-node pool, all-exhausted, static mode, sender mode
- `action_ReevaluateCoordinatorOnNewBlock` no-op cases and active re-selection
- Block-range boundary reset and heartbeat reset

**Integration tests** (`noderuntests/coordinationtest/coordinator_failover_test.go`) — 4 new tests, all passing:
- Primary coordinator stops → failover coordinator completes the transaction
- Primary stops and recovers → failover index resets → system re-converges
- All candidates unresponsive → system goes idle safely (no infinite loop)
- Block-range boundary reset → re-selects primary → stalled transaction completes

### DIFF
--- a/core/go/internal/sequencer/coordinator/coordinator.go
+++ b/core/go/internal/sequencer/coordinator/coordinator.go
@@ -91,6 +91,7 @@ type coordinator struct {
 	stateTimeout                      time.Duration
 	nodeName                          string
 	coordinatorSelectionBlockRange    uint64
+	coordinatorFailoverIndex          int
 	maxInflightTransactions           int
 	maxDispatchAhead                  int
 
--- a/core/go/internal/sequencer/coordinator/inactive.go
+++ b/core/go/internal/sequencer/coordinator/inactive.go
@@ -18,12 +18,20 @@ package coordinator
 import (
 	"context"
 
+	"github.com/LFDT-Paladin/paladin/common/go/pkg/log"
 	"github.com/LFDT-Paladin/paladin/core/internal/sequencer/common"
 )
 
 func action_NewBlock(ctx context.Context, c *coordinator, event common.Event) error {
 	e := event.(*NewBlockEvent)
+	prevEffective := c.currentBlockHeight - (c.currentBlockHeight % c.coordinatorSelectionBlockRange)
 	c.currentBlockHeight = e.BlockHeight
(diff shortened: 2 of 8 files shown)


sagarkhandagre998:
@matthew1001 Please take a look

sagarkhandagre998:
@annamcallister Can you please take a look

annamcallister:
@sagarkhandagre998 Thank you for looking at this. However, the issue has been resolved by maintainers as part of a bigger architecture redesign towards the V1 release, so I'm going to close this PR.

Issue #699, opened 2025-06-15
Domain UI should show domain-specific config


awrichar:
### What would you like to be added?

The Paladin UI has a simple section for domains. It would be useful to show more details that are specific to each deployed private smart contract.

For instance, Pente should show the privacy group members. Noto should show the notary config. Zeto should show the token type in use.

This information should all be stored already but may not yet be exposed on the RPCs for private smart contracts.

### Why is this needed?

This would be extremely useful for inspecting domains and demonstrating domain functionality.

awrichar:
Other thoughts: 
- As metadata is added (see https://github.com/LF-Decentralized-Trust-labs/paladin/issues/613), it should also be exposed.
- Private smart contracts need a better way to sort, including by creation time (and UI probably should show newest first by default).

awrichar:
Another small issue on this page - if you have a dialog open (such as the "deploy" or "transfer" dialog) and navigate to another browser tab, then return to the Paladin UI, the dialog closes. Something in a React hook must be resetting it. This is annoying if you are trying to copy a value into the dialog from another page.

This can be split into a separate issue if needed.

awrichar:
Here's an attempt at some more specific enhancements needed for Domains:
- Noto private contracts should show `name` and `symbol` prominently, and should show the entire `config` JSON in a detail view (consider if this is a popup, or if we need a new, linkable details page)
- Zeto private contracts should show the `tokenName` prominently, and should show the entire `config` JSON in a detail view
- "Deploy new" should be disabled for Pente
- "Deploy new" for Noto/Zeto should have a better progress indicator and should show you the address of the new contract you deployed (maybe dropping you to a details page?)
- Noto should support mint, transfer, burn, and [balance](https://github.com/LFDT-Paladin/paladin/pull/781) actions
- Zeto should support mint, transfer, burn, and balance actions
- Look up a deployed domain private contract by address (similar to the lookup for transactions)

In addition, I think we need a new "Privacy Groups" top-level section, separate from Domains:
- Render a table based on `pgroup_queryGroups` (visually similar to the Domains page)
- Important fields include: name (but must handle empty names), domain, contract address, members (maybe show count and a subset if there are lots of members)
- Details page (or popup?) should include full list of members and configuration
[...]

PR #1162, opened 2026-04-23
Update domains ux

gabriel-indik:
Included in this PR:
 - Changes to the `domain_querySmartContracts` RPC endpoint so that the deploy transaction information is included for each smart contract
 - Updates to the smart contract filter fields so it is possible to filter/sort by the `created` value of the deploy transaction
 - Updates to the UI to include updated interface with deployed transaction in smart contracts
 - Updates to the domains panel:
   - New column "Deployed" in the smart contract table
   - Make the new column sortable
   - Update the JSON RPC API call to include sorting value
   - Add pagination controls
   - Update the JSON RPC API call to include pagination

image

Defaults value set to: 
 - Default order: most recent contracts first
 - Show 10 contracts per page (can be changed to 25, 50 or 10)


core/go/internal/domainmgr/manager.go
```diff
@@ -344,27 +345,61 @@ func (dm *domainManager) populateContractConfig(result *pldapi.DomainSmartContra
...
 		Filters:     smartContractFilters,
 		Query:       jq,
-		MapResult: func(pt *PrivateSmartContract) (result *pldapi.DomainSmartContract, err error) {
-			_, dc, err := dm.enrichContractWithDomain(ctx, dbTX, pt)
-			if err == nil {
-				result = &pldapi.DomainSmartContract{
+		Finalize: func(q *gorm.DB) *gorm.DB {
+			return q.
+				Joins(`INNER JOIN "transactions" ON "transactions"."id" = "private_smart_contracts"."deploy_tx"`).
+				Select(`"private_smart_contracts".*`, `"transactions".*`)
```
awrichar, line 384: I don't think this will be reliable, except on the node that happened to deploy the contract (other nodes won't have the transaction).

Counter-proposal here (which I had sitting on a branch for a long time but never added to a PR) is to persist a `created` timestamp on private smart contracts: https://github.com/LFDT-Paladin/paladin/pull/1184

I added some changes to do a one-time best-effort backfill of creation time, following your JOIN logic from above - it will only be useful on the deployer node, and other existing contracts will just default to 0 (beginning of time).

awrichar (changes requested):
Pagination of smart contracts looks great, but I have concerns (noted inline) about using a join for the created timestamp.

There are also a number of other domain-related suggestions in https://github.com/LFDT-Paladin/paladin/issues/699#issuecomment-4400000622 - can decide if those are worth rolling in here or in a follow-up.

gabriel-indik:
Replaced by https://github.com/LFDT-Paladin/paladin/pull/1192

PR #1165, opened 2026-04-29
Tx receipts

gabriel-indik:
Ensure that the transactions panel shows paladin entries when there is a receipt for it, independently of whether the actual Paladin transaction is present or not. In addition, modify the transaction details panel so that it shows the receipt details in full for cases where the Paladin transaction is not present.


matthew1001 (approved):
LGTM

PR #1166, opened 2026-04-30
Show public Tx

gabriel-indik:
Handle the case where there is a receipt but no Paladin transaction, specifically display the associated blockchain transaction information in the details panel:

image


gabriel-indik:
Also in this PR: handle exceptional case where there are no blockchain transactions. In that case, instead of seeing an error message on the transactions panel, an empty state message "There are no transactions" will be shown.

awrichar:
General UX feedback - some related to this PR but some probably just add-ons.

---

It seems like there's nowhere to "click" on the overall transaction in the transaction list. I can click the Paladin transactions, but if there are no Paladin transactions, it seems like the only way to get to the detail page is by using the "lookup" feature. Feels like there should be a link at the overall transaction level.

---

The "contract address" field on the main transactions page is not very useful. This is generally only populated for contract deployment, and it looks odd to be unpopulated all the time (seems like we don't know what contract was invoked... but the invoked contract is actually `to` in this context).

I think it should still show on the transaction detail page, but only if populated (hide it completely if empty, to avoid confusion).

---

I'm confused about the Paladin transaction view now. I thought we had a tabbed interface to show all the Paladin transactions that are rolled up in a single blockchain transaction. But now, I seem to land on a page that only shows a single Paladin transaction:

Paladin
Paladin

It's always very useful to see that there were multiple Paladin transactions atomically confirmed by a single blockchain transaction, so I think we need to go back to some interface that shows this.

---

I don't get anything useful when I click on the event logs - just the Ethereum signature. Were we going to add some attempt at decoding them?

gabriel-indik:
This turned out to be quite an interesting and challenging update to the UI. Mainly because of the goal that was set to allow the user to provide either a blockchain transaction hash, or a receipt ID, or a Paladin transaction ID (for which there may be no receipt) and the UI should be smart enough to display the right information on screen. Not only that, but also in the cases where other Paladin receipts or transactions happen to be present within the same blockchain transaction, these should be automatically shown grouped. The logic has been optimized to make only the necessary API calls in order to retrieve information efficiently.

The scenarios that must be covered are:
 - Blockchain transaction hash for which there are no receipts
 - Blockchain transaction hash for which there are receipts
      - None of the receipts have an associated Paladin transaction
      - All of the receipts have an associated Paladin transaction
      - Some of the receipts have an associated Paladin transaction
  - Receipt ID for which there is no associated Paladin transaction
  - Receipt ID for which there is an associated Paladin transaction
  - Paladin transaction for which there is an associated receipt
  - Paladin transaction for which there is no associated receipt

Note that for all the bullets above (except for the last one) the UI would need to check for the presence of other receipts or Paladin transactions within the same Blockchain transaction in order to perform the grouping.

Also included in this PR:

Removal of `Contract address` and addition of a "navigate to / open" option for the blockchain transaction as requested in the feedback comments. Note that different approaches were tried out, like making the entire surface of the blockchain transaction clickable, but when testing it, it didn't feel right. Specifically, it did not feel intuitive (in part because of the nested clickable surfaces) but also because it would make it difficult to maintain accessibility.

image

awrichar (approved):
This looks great - thanks for the extra iteration here.

PR #1170, opened 2026-05-03
Paginate smart contracts table

pradhyum6144:
The smart contracts query in the Domains panel was using a hardcoded `limit: 100` with no way to fetch further results. Deployments with more than 100 contracts per domain would silently truncate the list.

Converts `SmartContractsTable` to use `useInfiniteQuery` with cursor-based pagination on `address`, consistent with how registry entries and indexed transactions are already paginated. Adds `SMART_CONTRACTS_QUERY_LIMIT` to the shared constants and an `ISmartContract` interface to replace the `any` type on contract list items.


awrichar:
@pradhyum6144 thanks very much for looking into this one. After some discussion, we've decided to steer away from infinite scroll for these pages, as it causes some friction when exploring long lists of items. Paginated tables for smart contracts have been included in https://github.com/LFDT-Paladin/paladin/pull/1162 among a few other changes to this page.

Sorry for the pivot here, but I hope you're able to contribute more in the future!

Issue #1055, opened 2026-03-02
Add substates to common state machine


annamcallister:
### What would you like to be added?

The ability to define substates within the existing common state machine definitions.

### Why is this needed?

We need to formalise the difference between states which result in meaningful difference in the events that are handled and the actions taken in response to them, and states which are just informational. Not doing this currently means we have some states which a state machine can skip, as it might not get the notification. This is resulting in 
* bugs- where we a state machine can get stuck unable to leave a state because it has missed a notification
* excessive over definition of event handling to avoid such bugs

sagarkhandagre998:
@annamcallister Can you please take a look - https://github.com/LFDT-Paladin/paladin/pull/1171

annamcallister:
Closing as we have released V1, and do not have ongoing concerns about not having these substates. A new issue can be opened if we see the need for them in the future.

PR #1171, opened 2026-05-04
feat(sequencer): add informational substates to common state machine

sagarkhandagre998:
Feature  - https://github.com/LFDT-Paladin/paladin/issues/1055

Added the concept of substates to the generic StateMachine to formalise the difference between states that affect event routing and states that are purely informational progress markers within a parent state.

Previously, fine-grained progress had to be modelled as full top-level states with their own event handlers. This caused two classes of bugs:
- State machines could get stuck when a notification arrived before the state was entered (missed-notification bug).
- Excessive defensive event-handler duplication across many states to work around the above.

Substates are informational labels only. They do not affect which events are handled or which transitions are taken, so there is no risk of a missed-notification bug. They are automatically cleared to nil on every outgoing transition so a stale substate can never leak into a new state.

Changes:
- Add  field to StateMachine
- Add GetCurrentSubstate(), SetSubstate(), and SubstateAction() methods
- Auto-clear substate at the start of every transition in evaluateTransitions
- Expose GetCurrentSubstate() on StateMachineEventLoop as a pass-through
- Add six tests covering: set on entry, cleared on transition, set in event-handler action, no effect on event routing, SubstateAction helper, and EventLoop accessor delegation

### DIFF
--- a/core/go/internal/sequencer/statemachine/statemachine.go
+++ b/core/go/internal/sequencer/statemachine/statemachine.go
@@ -159,6 +159,7 @@ type TransitionCallback[S State, E any] func(ctx context.Context, entity E, from
 type StateMachine[S State, E Lockable] struct {
 	stateMu            sync.RWMutex
 	currentState       S
+	substate           any // informational substate; nil when none. Cleared automatically on every transition.
 	lastStateChange    time.Time
 	latestEvent        string
 	definitions        StateDefinitions[S, E]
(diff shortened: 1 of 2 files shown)


sagarkhandagre998:
@annamcallister Please take a look

annamcallister:
Closing this PR as adding substates is only valuable if they are consumed and exercised by the sequencer state machines. Now we have released V1 without concerns about the lack of substates I am also closing the issue.

PR #1172, opened 2026-05-05
fix(sequencer): add expiry to assembly and endorsement requests

### LINKED ISSUE #1119: bug: Assembly and endorsement requests have no expiry
### What happened?

A sequencer for a contract built up a backlog of assembly and endorsement requests. These are handled on a single thread in the order they are received. If there is some delay in processing, the sequencer can get behind, and not start processing a request until after the coordinator who has sent the request has already stopped waiting for its response. At this point the coordinator sending the request has usually started to process a new transaction, sending more requests, which are even further down the queue not being handled. This results in cascading stage timeouts and all transactions for that contract becoming stuck.

Assembly and endorsement requests to include an expiry time to show when the requester will no longer accept the result back. If this expiry time is reached, the request should be abandoned and the most recent non expired request handled instead.

N/A

_No response_

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
# paste output here
```

</details>

Prantik-07:
## Fixes #1119

Assembly and endorsement requests had no expiry  if an originator node fell 
behind, stale requests would pile up and block newer ones, causing all 
transactions for that contract to time out.

`IdempotentRequest` already tracked `firstRequestTime`, so I added an 
`IsExpired()` check. When a request crosses the limit, it's abandoned and 
the transaction is moved back to the pool. Both expire at `stateTimeout * 2` 
by default.


matthew1001:
thanks for this PR @Prantik-07 We're currently (specifically @annamcallister ) doing some core refactoring of the state machines, so while these changes look reasonable at a glance, I think they probably need looking at once we have https://github.com/LFDT-Paladin/paladin/pull/1178 merged and the changes it makes. Particularly it probably needs Anna to review this PR which she probably won't have cycles for while working through that PR. Mainly wanted to say thank you for the contribution, let's leave it here for now if that's OK

Prantik-07:
Thanks for the heads up, Matthew! Completely understand. It makes total sense to wait for Anna's state machine refactoring in #1178 to land first. I'll leave this on hold and happy to rebase or adjust the logic once the new changes are merged.

annamcallister:
Apologies @Prantik-07 I did not see this PR and the issue got addressed in https://github.com/LFDT-Paladin/paladin/pull/1205. 

The changes here though wouldn't have fixed the observed issue though. This adds an expiry time on the side of the sender, which is already covered by state/request timeouts. What was needed here was to allow the receiver of the message to understand the point in time at which they should no longer work on actioning the request, since the sender is no longer listening for a response.

PR #1176, opened 2026-05-06
Block height and private state aware assembly

annamcallister:
formatted markdown for viewing diagrams https://github.com/LFDT-Paladin/paladin/blob/private-state-completeness/core/go/internal/sequencer/private-state-completeness-index.md

### DIFF
--- a/core/go/internal/sequencer/private-state-completeness-index.md
+++ b/core/go/internal/sequencer/private-state-completeness-index.md
@@ -1,0 +1,263 @@
+# Private State Completeness Index
+
+## Problem
+
+Private state distribution and on-chain transaction confirmation are independent processes. A transaction can be confirmed at block N while the private state payloads (coin states, lock states, info states) are still in transit from the coordinating node. There is currently no way for a node to answer:
+
+> *"Have I received all the private state data I am entitled to for every transaction confirmed up to block N?"*
+
+This question is a prerequisite for knowing whether a node is in a position to correctly assemble new transactions. A node that is missing private states it should have may produce inconsistent results when assembling against an incomplete view of the ledger.
+
+---
+
+## Design Goals
+
+- Provide a queryable index: `GetPrivateStateCompleteness(domain, contract, blockNumber)` returning whether the node is complete and, if not, which transactions are still waiting and for which state.
+- Core infrastructure must not decode domain state payloads or reason about domain-specific entitlement rules (which participants are entitled to which states). Those decisions are delegated to the domain plugin. Core observes only generic state categories (`Info`, `Coin`, `Data`, `Unavailable`) that are present for all domains, without interpreting their content.
+- The answer must be **incremental**: computed once per transaction at confirmation time and updated as private states arrive, not recomputed on demand across all history.
+- The index only covers successful confirmed transactions. Failed transactions do not produce private states that participants need to receive.
+
+---
+
+## Domain-Agnostic Design via `CheckStateCompletion`
+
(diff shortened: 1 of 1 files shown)


core/go/internal/sequencer/private-state-completeness-index.md
```diff
@@ -0,0 +1,263 @@
...
+
+### The `INDETERMINATE` Problem
+
+Not all domains can determine which missing states a node is entitled to. For domains without entitlement metadata (Noto V0 has no manifest; Zeto has no equivalent), the existing `CheckStateCompletion` behaviour returns `FirstUnavailableId` regardless of whether the node has any entitlement to that state. This is incorrect for the completeness index: on a Noto V0 transfer from Alice to Bob, Alice's change coin appears in Bob's `state_confirm_records` (written by the event indexer from the on-chain `outputs[]` array) but was never distributed to Bob. Bob's node would be permanently marked as waiting for a state that will never arrive.
+
+A new `completeness_indeterminate` boolean field is added to `CheckStateCompletionResponse`. When set, the completeness worker does not insert or update any row for that transaction. Because the table contains only waiting rows, the transaction's absence means the block-height query does not count it — the same outcome as if it were explicitly excluded. The receipt listener call site receives the new return value but ignores it, preserving existing behaviour.
```
matthew1001, line 36: Just wanted to check which table `does not insert or update any row for that transaction` is referring to

matthew1001:
Thanks for this document @annamcallister, it's really clear and generally looks to be a sensible way to balance assembling ahead of the base ledger without assembling with out-of-date state.

I have a couple of high-level questions.

1. What's the reason we need to persist the missing states/transactions we are waiting for? I think the answer is that in the current design, Paladin asks every successful transaction to persist any missing state records, and hence the coordinator can determine how far back in the chain there are missing states for without re-asking the transaction(s). I wonder if you considered any approaches that didn't persist per-transaction missing state records? 
2. I think the design implies that all TX assembly for a given domain instance stops once the distance between the oldest known missing state exceeds the configured tolerance. Do you think there are any hypothetical cases where a coordinator crashes for an extended period of time before it had the opportunity to send states, is taken over by another instance (for new transaction coordination), but that the missing state prevents any future assembly until the original coordinator is resumed? I'm primarily asking the question to make sure we know all cases where assembly stops for a period of time so we can document them. But also in case there are ways in which a non-essential node ultimately prevents that domain instance being able to process new transactions.

annamcallister:
> What's the reason we need to persist the missing states/transactions we are waiting for? I think the answer is that in the current design, Paladin asks every successful transaction to persist any missing state records, and hence the coordinator can determine how far back in the chain there are missing states for without re-asking the transaction(s). I wonder if you considered any approaches that didn't persist per-transaction missing state records?

I'm not sure I fully understand what the question is here. The transaction is the unit at which we learn what private states we should have for a given block height. Building an index of these missing ones is the only efficient way we can query this data at assembly time. I don' really see another way to answer this question. Did you have an alternative proposal?

> I think the design implies that all TX assembly for a given domain instance stops once the distance between the oldest known missing state exceeds the configured tolerance. Do you think there are any hypothetical cases where a coordinator crashes for an extended period of time before it had the opportunity to send states, is taken over by another instance (for new transaction coordination), but that the missing state prevents any future assembly until the original coordinator is resumed? I'm primarily asking the question to make sure we know all cases where assembly stops for a period of time so we can document them. But also in case there are ways in which a non-essential node ultimately prevents that domain instance being able to process new transactions.

It doesn't stop on the whole domain instance, only for transactions which require the node which has dropped sufficiently behind on block indexing/receiving private states. For Pente this effectively stops the whole domain. But for Noto it really only affects transactions that the behind node has originated. (I think it is impossible for the notary to get behind since it submits all transactions and distributes all private states). In all cases though, a node who is behind by this definition is incapable of successfully assembling or endorsing transactions, we just move to a model where we pause on this condition rather than letting these transactions fail.

core/go/internal/sequencer/private-state-completeness-index.md
annamcallister: It's referring to the new table that will be created to support this feature.

Issue #1142, opened 2026-04-13
docs: Doc the distributed sequencer state machines


matthew1001:
### Current State

Users need to have a reasonable understanding of the way multiple Paladin nodes coordinate transactions between themselves. For problem diagnosis, it is important to be able to understand the reasons why a particular transaction has got stuck in a particular state, or not entered an expected state.

The state machine states are internal to Paladin, and hence are subject to change over future versions of Paladin. As such, the state machine docs would ideally be machine generated and automatically updated when the state machine changes.

It's also really useful for future contributors to be able to learn about the internal code structure and behaviours, so these docs should aim to serve both use-cases.

### Desired State

A new state machine section of the `Distributed sequencer` architecture docs.

oindrilakha12-ui:
Hi @matthew1001, I’d like to work on this issue. Please let me know if there are any guidelines to follow. Thanks!

oindrilakha12-ui:
Hi @matthew1001 , I’ve created a PR for this issue: #1158. Please review when you get time.

annamcallister:
https://github.com/LFDT-Paladin/paladin/pull/1269

PR #1178, opened 2026-05-13
Block height aware coordinator selection

annamcallister:
This PR delivers support for block height based deterministic coordinator selection in the distributed sequencer. 

I expect user facing documentation for these changes to be delivered under #1142; however, this PR description has been written in a way to support that task.

Included in these changes is substantial test rework to improve code hygiene.

---

# Paladin Sequencer

The sequencer is responsible for coordinating the assembly, endorsement, and base-ledger submission of private transactions across the nodes of a privacy group. For any given privacy-group contract, one node at a time acts as the **active coordinator**: it receives delegated transactions from originators, drives them through assembly and endorsement, and submits the resulting base-ledger transactions. All other nodes observe the process and delegate their locally-created transactions to the active coordinator.

The design prioritises throughput and stability over fairness of distribution. An active coordinator is never displaced by epoch rotation or priority recalculation alone — it remains active until it receives an explicit challenge from another node that has itself begun accepting delegations. That situation only arises when two nodes have simultaneously received work for the same contract, creating a conflict that needs to be resolved. Because displacement only ever happens through explicit challenge, the correct policy for observers and originators is to follow whoever is actively heartbeating rather than continuously re-evaluating the priority list. There is no mechanism to rotate coordination work across nodes for load-balancing purposes, and a single node may coordinate for an extended period if no conflict arises.

This document covers the architecture of coordinator selection, the coordinator and originator state machines, the handover protocol between coordinators, signing key rotation, and the ahead-of-chain state tracking maintained by the grapher.

---

### `COORDINATOR_ENDORSER` mode

For privacy groups configured with `COORDINATOR_ENDORSER`, coordinator selection is fully deterministic from contract configuration and block height — no messages are exchanged to agree on it. Both the coordinator and originator call the same selection function independently and arrive at the same answer.

The pool of candidate coordinators is fixed at contract initialisation time: it is the sorted list of node names extracted from the endorser identity locators declared in the contract configuration. This pool never changes at runtime.

Selection proceeds as follows:

1. The current block height is **rounded down to the nearest block-range boundary** (`effectiveBlockNumber = height - (height % blockRange)`). All blocks within a range of `blockRange` consecutive blocks map to the same epoch, so the coordinator does not rotate on every block.
2. An FNV-32a hash of the effective block number (as a decimal string) is taken, and the preferred coordinator is `pool[hash % len(pool)]`.
3. The result is returned as a **walk-around ordered list**: the preferred coordinator appears at index 0, followed by the remaining pool members in their original sorted order, wrapping around. For a pool of `[A, B, C, D]` with `C` preferred, the result is `[C, D, A, B]`. Index 0 is the highest-priority coordinator; a lower index means higher priority.

The walk-around ordering encodes a tie-breaking rule for conflicts, not a schedule for distributing work. A node at index 0 has the strongest claim if a conflict arises in the current epoch, but it will not receive that work unless a conflict actually surfaces — for example, because it received a delegation before hearing a heartbeat from the existing active coordinator. Priority determines which node yields when two live coordinators discover each other; it has no effect during normal single-coordinator operation.

The block range is configurable (default: 100 blocks, minimum: 10). A larger range means less frequent key rotations and less rotational overhead; a smaller range gives more frequent key hygiene cycles. The block range has no effect on which node accumulates coordinator work — an active coordinator is never rotated out for load-balancing purposes.

### `COORDINATOR_STATIC` and `COORDINATOR_SENDER` modes

For `COORDINATOR_STATIC` the coordinator node is read once from the contract configuration at startup and never changes. For `COORDINATOR_SENDER` the coordinator is always the local node itself. Block-height-based selection, priority lists, and epoch rotation do not apply in either mode.

### Active coordinator tracking

Both sides track `currentActiveCoordinator`, the node they believe is currently responsible. This is seeded from the priority list or static config and updated reactively on heartbeats and delegation rejections. Neither side initiates any protocol step without consulting this field.

Because a coordinator is only ever displaced through an explicit challenge, the correct update policy is simply to follow whoever is actively heartbeating. Continuously re-evaluating priority against the list would cause spurious handovers; instead, priority is consulted only at the moment of conflict, when a node has received a delegation and must decide whether to challenge the current coordinator. As a consequence, two live coordinators can briefly coexist and an observer may flip between them; this is a transient state that the handover protocol resolves.

---

## Coordinator State Machine

The coordinator runs a declarative event-driven state machine. States, transitions, and side-effects are expressed as guards, validators, and actions attached to event handlers — there is no ad-hoc conditional logic in the core processing path.

### `Idle`

No active coordinator is known and this node is not coordinating. Any delegation received is accepted immediately — the node becomes `Active` without going through the handover protocol. A live heartbeat from another node moves the coordinator to `Observing`.

### `Observing`

Another node is known to be the active coordinator. Delegations are rejected with the current active coordinator named in the rejection payload unless this node is higher priority — in which case accepting the delegation would create a conflict that warrants a handover, so the node transitions to `Elect` and sends a `HandoverRequest` to the current active coordinator. If no heartbeat is received for `inactiveGracePeriod` intervals the coordinator returns to `Idle`.

### `Elect`

This node has received a delegation while higher priority than the current active coordinator. It has sent a `HandoverRequest` to the current active coordinator and is waiting for that coordinator to begin flushing. While in `Elect` the node heartbeats and accepts further delegations (but does not yet dispatch them). A state timeout prevents indefinite waiting: if the active coordinator does not acknowledge within the timeout window, the node transitions directly to `Active` and the base ledger resolves any submission conflicts.

### `Prepared`

The active coordinator has acknowledged the handover by entering `Closing_Flush` or `Closing`. This node is now waiting for the coordinator to finish draining its in-flight transactions. There is no state timeout here — once the coordinator has acknowledged the handover, this node waits as long as necessary for it to finish. If the inactive grace period elapses without a heartbeat from any coordinator, the node proceeds to `Active` directly.

### `Active`

The node is the active coordinator and remains so indefinitely — epoch boundaries and priority recalculations do not cause it to step down. It assembles transactions, drives endorsement, queues transactions for dispatch, and heartbeats on the normal interval. Step-down is triggered only when a higher-priority node makes an explicit challenge: either a `HandoverRequest` or a heartbeat arriving from a node already in a live coordinating state (`Elect`, `Prepared`, `Active`, or `Active_Flush`). Both signals indicate that the other node has itself received delegations and a conflict exists; at that point this node initiates the step-down path.

### `Active_Flush`

A key-rotation flush. The node remains the active coordinator but the dispatch loop has been stopped and no new transactions are dispatched. The node waits for all already-dispatched transactions to confirm before re-entering `Active` with a fresh signing key. Delegations are accepted normally; newly assembled transactions are pooled and will be dispatched under the new key. If a higher-priority node arrives during a flush, the node transitions to the step-down path.

### `Closing_Flush`

A step-down flush. The node is no longer the active coordinator (it has been superseded by a higher-priority node) but still has dispatched transactions in flight that must be confirmed before it can report a clean exit. Pre-dispatch transactions are evicted at the point of entering this state so originators see them drop from the heartbeat snapshot and know to redelegate. An immediate heartbeat is sent on entry so any waiting node in `Elect` receives the step-down acknowledgement without delay.

### `Closing`

All in-flight transactions have confirmed. The node continues to heartbeat for a grace period so that any node in `Prepared` sees the clean-exit signal and can take over safely. An immediate heartbeat is sent on entry. After the grace period the node transitions to `Idle` or `Observing` depending on whether it has recently seen a heartbeat from another live coordinator.

---

## Handover Protocol

The handover protocol is the resolution mechanism for a specific conflict: two nodes have simultaneously received delegations for the same contract, and the higher-priority one is demanding the step-down. It is never triggered by epoch rotation or priority recalculation alone. The two-phase structure (`Elect → Prepared → Active`) gives the outgoing coordinator time to drain its in-flight transactions and hand over confirmed lock and state data cleanly, while keeping the incoming coordinator informed of progress at each step.

```
1. Node B (Observing) receives a delegation and is higher priority than the current active
   coordinator A. → State_Elect. Sends a HandoverRequest to A.

2. Node A (Active or Active_Flush) receives the HandoverRequest. Evicts all pre-dispatch
   transactions. → State_Closing_Flush (if dispatched txns remain) or State_Closing.
   Sends an immediate heartbeat so B sees the acknowledgement without waiting.

3. Node B sees a Closing_Flush or Closing heartbeat from A. → State_Prepared.
   Clears its timeout schedule; now waits for A to finish draining.

4. Node A drains all dispatched transactions. → State_Closing.
   Sends an immediate heartbeat so B sees the flush-complete signal.

5. Node B sees a Closing heartbeat from A. Imports A's confirmed lock and state data.
   → State_Active.
```

**Timeout path**: if no `Closing_Flush` or `Closing` heartbeat arrives from A within the state timeout, B transitions directly to `Active`. The base ledger provides idempotency protection against any simultaneous submissions.

**Higher-priority interruption**: if a third node C with higher priority than B announces itself before the handover from A completes, B stands down. Depending on whether B has its own inflight transactions, it moves to `Observing`, `Closing`, or `Closing_Flush`, and C becomes the new incoming coordinator.

---

## Signing Key Rotation

Each active coordinator uses a per-tenure signing identity for base-ledger submissions, of the form `domains.<contractAddress>.submit.<uuid>`. A fresh identity is generated on every transition into `Active` — including both the initial entry and every re-entry after a key-rotation self-transition.

**Why rotate at every epoch boundary**: key rotation applies even when the same node remains active across many epochs without any reselection event. The purpose is purely key hygiene — preventing any single address from accumulating an unbounded nonce history — not load balancing or ensuring other nodes get a turn. The rotation happens in all coordinator modes.

**Point of commitment**: the signing key is not applied to a transaction when it is assembled or placed in the dispatch queue. It is captured at the exact moment the dispatch loop processes the transaction — on the transition to `State_Dispatched`. Because this happens inside the dispatch goroutine synchronously, and the dispatch loop is stopped before any key rotation step, there is no window in which a transaction could pick up the wrong key.

**Key rotation flush**: when an epoch boundary arrives while dispatched transactions are still in flight, the coordinator transitions to `Active_Flush`. The dispatch loop is stopped — no new transactions are dispatched — and the node waits for all in-flight transactions to confirm. Only then does it self-transition back to `Active`, where `OnTransitionTo` generates the new identity and restarts the dispatch loop. Any transactions that were queued but not yet dispatched remain in the queue and are dispatched under the new key when the loop restarts. This is correct — they belong to the new epoch.

---

## Dispatch Loop

The dispatch loop is a dedicated goroutine separate from the coordinator's single-threaded event loop. The coordinator enqueues `CoordinatorTransaction` references onto a buffered channel; the dispatch loop pulls them off, calls the transaction's `DispatchedEvent` handler synchronously, and posts the resulting state-transition event back to the coordinator's event queue.

**Synchronous dispatch**: `tx.HandleEvent(DispatchedEvent)` runs to completion inside the dispatch goroutine. The signing key is allocated, the base-ledger submission is made, and the transaction advances to `State_Dispatched` — all before the loop moves on to the next item. The resulting `TransactionStateTransitionEvent` is queued back to the coordinator's event loop and processed there in the normal way.

**Synchronous stop**: stopping the dispatch loop is a synchronous operation. The loop's context is cancelled and the coordinator blocks until the goroutine exits. This gives the coordinator a precise, stable snapshot: once `stopDispatchLoop` returns, there are no further dispatch calls in progress and no further `TransactionStateTransitionEvent` messages will be produced by the loop (though those already queued will still be delivered).

**Step-down drain**: when handing over to a higher-priority node the coordinator stops the dispatch loop and then evicts all pre-dispatch transactions. As part of this eviction the dispatch channel is also drained — any references still buffered there are discarded. This prevents a future dispatch loop (should this node later be re-elected) from processing stale transaction objects that no longer exist in the coordinator's tracking structures. Without the drain, stale objects would be dispatched under the new epoch's signing key and could cause duplicate base-ledger submissions.

**Thread safety**: the coordinator event loop is single-threaded and is the only writer to `transactionsByID`. The dispatch loop is the only reader of the dispatch channel. `c.signingIdentity` is protected by the stop/start discipline: the loop is stopped before a new identity is written, and the loop is not restarted until after the write is complete. The in-flight throttle (`inFlightTxns`) is protected by a condition variable shared between the two goroutines.

---

## Endorsement Gathering

When a coordinator transaction enters the endorsement phase, the coordinator iterates its attestation plan. For each `ENDORSE`-type request it sends endorsement requests to all listed parties and nudges any outstanding requests on each heartbeat interval until the threshold is met.

**Threshold**: each `AttestationRequest` carries an optional `threshold` field specifying how many endorsements are required (M-of-N). When the threshold is unset or zero, all listed parties must endorse. Once the received endorsement count meets the effective threshold the request is considered fulfilled and the transaction proceeds; parties that have not yet responded are still nudged so that any response already in flight is not silently discarded.

**Idempotent nudging**: each outstanding request is tracked by an `IdempotentRequest` that suppresses duplicate sends within the request-timeout window. The coordinator can call `sendEndorsementRequests` freely on every heartbeat interval without generating spurious network traffic.

---

## Grapher: Ahead-of-Chain Locks and Private State

The grapher maintains the coordinator's ahead-of-chain view of lock state and private state data. Its primary purpose is to allow new transactions to be assembled correctly even when earlier transactions have not yet been confirmed on the base ledger — a prerequisite for the pipeline throughput that makes the sequencer useful.

### Lock lifecycle

A lock is first recorded as transaction-owned when a transaction is assembled. On base-ledger confirmation the lock's transaction reference is cleared and a `ConfirmedAtBlock` timestamp is stamped. The lock is retained until `currentBlockHeight >= confirmedAtBlock + blockHeightTolerance`, at which point the persisted state records are expected to have caught up and the lock is released. Locks imported from an outgoing coordinator on handover follow the same lifecycle from the point of import.

### Private state and `AllowedNodes`

Private state data is associated with the states a transaction mints. Each state carries an `AllowedNodes` list derived from the assembly response's distribution list, recording which nodes are permitted to hold that state's private data. When the coordinator exports its grapher state for inclusion in a heartbeat or handover snapshot, it produces a per-recipient filtered view: each node only ever receives the private state data it is authorised to hold. Lock metadata (state IDs, lock types, block numbers) is on-chain information and is exported unfiltered.

`AllowedNodes` must always be explicitly set. A nil or empty list causes the state to be silently excluded from all exports. This is an intentional security decision: accidental omission errs on the side of withholding data rather than inadvertently broadcasting it.

### Coordinator handover import

When the incoming coordinator transitions from `Prepared` to `Active` on receipt of a `Closing` heartbeat, it imports the outgoing coordinator's lock and state snapshot. This gives it the full ahead-of-chain view needed to assemble new transactions correctly from the moment it takes over. Existing entries in the importer's grapher are never overwritten — local knowledge always takes precedence.

---

## Originator State Machine

The originator is the client-side counterpart of the coordinator. It creates transactions and delegates them to whichever node it believes is the active coordinator. It does not drive endorsement or dispatch; it only submits delegation requests and monitors heartbeats to track the coordinator's progress.

### `Idle`

No active coordinator is known and this originator has no transactions to send. A live heartbeat moves it to `Observing`; a new transaction moves it directly to `Sending`. On entry, and on each epoch boundary while `Idle`, the originator resets `currentActiveCoordinator` to the highest-priority candidate so that the next delegation starts fresh rather than targeting a potentially stale coordinator from a previous sending cycle.

### `Observing`

An active coordinator is known but there are no outstanding transactions. The originator tracks the coordinator's liveness and rotates `currentActiveCoordinator` reactively. If no heartbeat arrives for `inactiveGracePeriod` intervals it returns to `Idle`.

### `Sending`

The originator has one or more transactions awaiting confirmation. On entry it sends a delegation request to `currentActiveCoordinator` immediately. It continues to process heartbeats and update its view of the active coordinator. Once all transactions are confirmed or finalised it returns to `Observing`.

---

## Originator Heartbeat Processing

In `State_Sending`, every incoming heartbeat is processed through a fixed five-step ordered action chain. Because each step can update `currentActiveCoordinator`, later steps see the updated value — in particular, if step 2 or 3 switches the active coordinator to the heartbeat sender, step 4's check naturally matches that same sender without any additional logic.

| Step | Condition | Action |
|---|---|---|
| 1 | Unconditional | Process confirmed transactions from the snapshot |
| 2 | Sender is live AND sender is higher priority than `currentActiveCoordinator` | Switch active coordinator to sender; reset liveness timer |
| 3 | Sender is live AND sender is NOT `currentActiveCoordinator` AND inactive grace period exceeded | Switch active coordinator to sender; reset liveness timer |
| 4 | Sender is live AND sender IS `currentActiveCoordinator` (possibly just updated in step 2 or 3) | Reset liveness timer; process dispatched-transaction updates from the snapshot |
| 5 | Sender IS `currentActiveCoordinator` AND snapshot is missing one of our in-flight transactions | Redelegate all outstanding transactions |

**Liveness**: a heartbeat sender is considered live when its reported coordinator state is `Elect`, `Prepared`, `Active`, or `Active_Flush`. `Elect` is included because a node in `Elect` is already heartbeating and accepting delegations; treating it as non-live would cause originators to ignore it unnecessarily.

**Step 2 — immediate switch on higher priority**: when a live heartbeat arrives from a node with higher priority than the current coordinator, the originator switches immediately rather than waiting for the old coordinator to step down. A live heartbeat from a higher-priority node signals that a conflict is underway — that node has received delegations and the handover protocol has likely been triggered. Switching immediately avoids the originator cycling into failover while the conflict resolves itself. The policy of following any live heartbeat sender is deliberate: since coordinator changes only happen through conflict resolution, the safest approach is to track whoever is actively coordinating, even if two live coordinators are briefly visible simultaneously.

**No redelegate flag**: the decision to redelegate is made entirely from observable state at step 5 — whether the current coordinator's snapshot is missing a transaction this originator believes is still in flight. There is no stored flag to manage; the check is idempotent and can safely fire on every heartbeat where the condition holds.

**Inactive coordinator**: if the liveness timer exceeds `inactiveGracePeriod` intervals without a qualifying heartbeat, the originator advances to the next candidate in the coordinator priority list and redelegates to it (see Coordinator Failover). For single-node pools this degenerates to a re-send to the same coordinator. If a different live node becomes available via steps 2 or 3, the originator switches to it immediately without waiting for the timer to expire.

---

## Coordinator Failover

When the active coordinator is unresponsive and the inactive grace period elapses in `State_Sending`, the originator walks the coordinator priority list rather than repeatedly delegating to the same unresponsive node.

The originator maintains a `failoverIndex` — an index into `coordinatorPriorityList` pointing to the next candidate to try. Each time the timer fires without a qualifying heartbeat, the walk advances one step:

1. `currentActiveCoordinator` is set to `coordinatorPriorityList[failoverIndex]`.
2. `failoverIndex` is incremented modulo the list length, wrapping from the last entry back to index 0.
3. The liveness counter is reset and a delegation request is sent to the new coordinator.

On any externally-driven change to `currentActiveCoordinator` — a heartbeat switch (steps 2 or 3 of the heartbeat chain), a delegation-rejection redirect, or a priority-list recalculation — `failoverIndex` is reset so the next self-driven walk step immediately follows the new active coordinator in priority order. If the new active coordinator is the top-priority node (index 0), `failoverIndex` is set to 1; otherwise it is set to 0.

The failover mechanism is a no-op for single-node pools and for `COORDINATOR_STATIC` / `COORDINATOR_SENDER` modes: in those cases there is no alternative coordinator to walk to, and the behaviour reduces to a redelegate to the same node.

---

## Coordinator ↔ Originator Communication

**Heartbeat**: broadcast on the normal heartbeat interval by any coordinator in `Elect`, `Prepared`, `Active`, `Active_Flush`, `Closing_Flush`, or `Closing`. The payload includes the sender's coordinator state, a snapshot of in-flight and confirmed transactions, and (for `Closing_Flush` and `Closing`) the grapher's filtered lock and state data for each recipient.

**Immediate heartbeats**: on entry to `Closing_Flush` an immediate heartbeat is sent so any node in `Elect` receives the step-down acknowledgement without waiting for the next interval. On entry to `Closing` an immediate heartbeat is sent so any node in `Prepared` receives the flush-complete signal without waiting.

**`HandoverRequest`**: the explicit signal that a conflict exists — both nodes have received delegations for the same contract. It is sent by a node entering `Elect` to the current active coordinator and is the only mechanism by which an active coordinator is asked to step down. Carries the sender's node identity and contract address. The request is implemented as an idempotent send and is re-sent on each request timeout interval until the node transitions out of `Elect`. The receiver handles it identically to receiving a higher-priority live heartbeat.

**`DelegationResponse` rejection payload**: when a coordinator rejects a delegation, the rejection includes the coordinator's `currentActiveCoordinator`. This allows the originator to fast-redirect to the right node without waiting for a heartbeat cycle.

**`CoordinatorPriorityListUpdated`**: an internal event (not transported over the network) queued by the coordinator to its co-located originator after recomputing the priority list. The originator uses this to keep its own priority list in sync for coordinator-selection decisions.

---

## Configuration Reference

The following configuration parameters (under `sequencer`) are most relevant to coordinator selection and handover behaviour.

| Parameter | Default | Minimum | Description |
|---|---|---|---|
| `blockRange` | 100 | 10 | Number of consecutive blocks forming a single coordinator epoch. |
| `closingGracePeriod` | 2 | 1 | Number of heartbeat intervals the outgoing coordinator stays in `Closing` after its flush is complete. |
| `inactiveGracePeriod` | 2 | 1 | Number of heartbeat intervals without a heartbeat before a coordinator or originator treats the remote node as gone. |
| `heartbeatInterval` | 10s | 1s | Interval between heartbeat transmissions from an active coordinator. |
| `blockHeightTolerance` | 5 | 1 | Number of blocks after confirmation at which the grapher releases confirmed locks and their associated private state data. Should be large enough that persisted state records have caught up with the chain within this window. |

---

# Sequencer test improvements

## 1. Mocks moved out of packages into a central `core/go/mocks/` directory

Every package under `sequencer` previously had a hand-managed (or generator-output) `*mocks.go` file sitting directly inside that package:

| Deleted (package-local) | Moved to (central) |
|---|---|
| `common/commonmocks.go` | `core/go/mocks/sequencercommonmocks/` |
| `coordinator/coordinatormocks.go` | `core/go/mocks/coordinatormocks/` |
| `coordinator/transaction/transactionmocks.go` | `core/go/mocks/coordinatortransactionmocks/` |
| `originator/originatormocks.go` | `core/go/mocks/originatormocks/` |
| `originator/transaction/transactionmocks.go` | `core/go/mocks/originatortransactionmocks/` |
| `transport/transportmocks.go` | `core/go/mocks/sequencertransportmocks/` |
| `syncpoints/syncpointsmocks.go` | `core/go/mocks/syncpointsmocks/` |
| `metrics/metricsmocks.go` | `core/go/mocks/sequencermetricsmocks/` |
| `sequencermocks.go` | `core/go/mocks/...` |

The old files were a mix of hand-written and generated mocks living alongside production code, making the package's own import graph prone to circular dependencies. They now follow the project-wide standard pattern (as used by `componentsmocks`, etc.), with each mock sub-package named descriptively.

The builder test files carry a `//go:build !generate_mocks` tag to avoid circular dependencies when the mock generation tool runs.

---

## 2. `test_utils.go` / `coordinator_test_utils.go` → `*_builder_test.go` (builder pattern)

The old helper pattern used monolithic factory functions that constructed test subjects with a long list of hard-coded defaults and `.Maybe()` overrides all in one go — e.g., `NewCoordinatorForUnitTest(t, ctx, originatorIdentityPool)` or `NewOriginatorForUnitTest(...)`. These were in non-test files (no `_test.go` suffix), meaning they were compiled into the production package for all build targets.

They've been replaced with proper builder types in `*_builder_test.go` files — `CoordinatorBuilderForTesting`, `OriginatorBuilderForTesting`, `TransactionBuilderForTesting` — with a fluent `With...()` setter API and a final `.Build(ctx)` call. This means:

- Each test is self-documenting about what it's actually varying
- Default config comes from `SequencerDefaults`, with `copySequencerDefaultsForTest()` doing a deep copy to prevent test pollution when a test mutates a pointer field
- The builder types are `_test.go`-only, so they do not count towards package coverage

---

## 3. The `SentMessageRecorder` consolidated into a shared `testutil` package

Previously, `SentMessageRecorder` was duplicated across packages:

- `coordinator/coordinator_test_utils.go` had one version that embedded `transaction.SentMessageRecorder`
- `originator/test_utils.go` had another version that also embedded `transaction.SentMessageRecorder`

Both versions added their own package-specific fields (e.g., `hasSentHandoverRequest`, `hasSentDelegationRequest`) on top. This is now a single unified `testutil.SentMessageRecorder` at `core/go/internal/sequencer/testutil/sent_message_recorder.go` that covers all message types across both coordinator and originator sides in one place, and is used via composition by both the `CoordinatorDependencyMocks` and `OriginatorDependencyMocks` structs.

---

## 4. The `spec/` package has been deleted entirely

The `spec/` package (5 files: ~2,500 lines) was a cross-package test layer testing coordinator/originator behavior from the outside — a pattern that was causing coverage overlap and making it hard to tell which package's tests owned which scenarios. All those scenarios have been absorbed into `state_machine_test.go` in each respective package, which grew substantially as a result.


matthew1001:
I'm starting to work through the review of this draft PR @annamcallister but I'm dropping a couple of questions/clarifications on the description here for discussion.

The `Elect` state is described as "This node has been selected as coordinator by the algorithm but is waiting for the previous coordinator's flush to complete before accepting transactions.". I think I'm right in saying that this algorithm could happen on the node that is the current active coordinator already given this follow on description:

"The Elect path is used by the current active coordinator at an epoch boundary."

Coupling that with this statement "In Elect the node is not yet accepting delegated transactions." I'm trying to understand if a) this is all still true as of our latest in-person discussions, and if so, b) is there a side effect to nodes trying to originate transactions during an epoch changeover. Will they get delegation rejections from the coordinator even though it might still carry on as the active coordinator?

core/go/internal/sequencer/common/events.go
```diff
@@ -29,6 +30,9 @@ type EmitEvent func(event Event)
 const (
 	Event_HeartbeatInterval          EventType = iota // emitted on a regular basis, interval defined by the sequencer config
 	Event_TransactionStateTransition                  // transaction state machine transition; originator/coordinator handle cleanup and side effects
+	Event_NewBlock                                    // a new block has been confirmed on the base ledger
+	Event_HeartbeatReceived                           // a heartbeat notification was received from the active coordinator
+	Event_EndorserNodesDiscovered                     // endorser node names discovered by the coordinator; used to grow the originator node pool
```
matthew1001: Quite like this 👍

core/go/internal/sequencer/common/states.go
```diff
@@ -0,0 +1,177 @@
...
+
+package common
+
+import "fmt"
+
+// OriginatorState is the state of the originator top-level state machine.
+type OriginatorState int
+
+const (
+	OriginatorState_Initial   OriginatorState = iota
```
matthew1001: Was there a specific reason for splitting out the common state types into originator/coordinator specific ones? I thought there was something nice about the fact that each one used the relevant subset of the full list of common states. Was it just clarity of the code you wanted to improve?

core/go/internal/sequencer/common/selection.go
```diff
@@ -0,0 +1,76 @@
...
+// deduped endorser pool. The selection is deterministic: all nodes that call this function with
+// the same pool and block height will independently reach the same result.
+//
+// For COORDINATOR_STATIC and COORDINATOR_SENDER modes, the coordinator field is set once at
+// construction/Start time and this function is never invoked.
+func SelectCoordinatorNode(
+	ctx context.Context,
+	nodePool []string,
+	currentBlockHeight uint64,
+	blockRange uint64,
```
matthew1001, line 112: I think this is one area of configuration that users need to ensure is consistent across all nodes? I think there are probably others as well, such as heartbeat intervals. Mainly just a note to make sure we have something in the docs for V1 which is clear about the config options that are likely to cause problems if not rolled out consistently across the nodes.

core/go/internal/sequencer/coordinator/grapher/grapher.go
```diff
@@ -33,60 +33,110 @@ import (
...
+	// distribution is unknown and it is excluded from all exports to avoid leaking private data.
+	// All locks are returned unfiltered — lock data (state IDs, types, block numbers) is on-chain metadata and
+	// does not need privacy protection.
+	ExportStatesAndLocks(ctx context.Context, node string) (ExportableStates, error)
+	// Forget fully removes a transaction and all its locks. Used for failure/reset paths (revert, repool, eviction).
+	// No-op if the transaction is not known (e.g. already confirmed).
 	Forget(ctx context.Context, transactionID uuid.UUID)
+	// ConfirmTransaction removes the transaction from the grapher's dependency tracking and minter index,
+	// and stamps confirmedAtBlock on its locks, clearing the transaction reference.
+	ConfirmTransaction(ctx context.Context, transactionID uuid.UUID, confirmedAtBlock uint64)
```
matthew1001: It's a minor point, but I quite like the principal of keeping the grapher cleaner from the perspective of states/outcomes/general sequencer activity. The other functions are quite explicit about the side effect they cause. Since the comment says what this function does, I'd favour calling it something clearer (we've used `Forget*` elsewhere - not that is has to be that, but something about the action the function is going to perform would feel more in keeping with the grapher being a standalone package

core/go/internal/sequencer/coordinator/grapher/grapher.go
```diff
@@ -33,60 +33,110 @@ import (
...
 	GetDependents(ctx context.Context, transactionID uuid.UUID) []uuid.UUID
 	LockMintsOnCreate(ctx context.Context, upserts []*components.StateUpsert, states []*components.FullState, transactionID uuid.UUID)
 	LockMintsOnReadAndSpend(ctx context.Context, readStates []*components.FullState, spendStates []*components.FullState, transactionID uuid.UUID)
 }
 
+// OutputState wraps a StateUpsert with AllowedNodes for coordinator handover and assembly requests.
+// AllowedNodes is the set of nodes expected to hold this state's private data, derived from
+// the assembly response DistributionList. AllowedNodes must always be explicitly provided —
+// a nil or empty list means the distribution is unknown and the state is excluded from all exports.
+type OutputState struct {
```
matthew1001, line 91: You could almost make a case for the grapher being privacy-agnostic, and being passed a function that knows how to treat the states it is given by the grapher (ditch them because the recipient should have them etc). Not a strong opinion that this PR should change. But I like the principle of keeping some of the new re-usable packages as clean as possible so maybe something to consider for future. Or for you to provide a strong rebuttal against :)

matthew1001: Thinking this through more, the code that ensures private data doesn't go to the wrong party is one of the most important components in Paladin. So the more I think about it, the more strongly I think I'm landing on splitting this out. A future PR on the grapher _**should**_ be fairly safe from accidentally breaking how private states are distributed. Incorporating privacy-maintaining logic into it feels more prone to really dangerous regressions than most other areas of the codebase. I'd still accept "future PR to handle" (or a rebuttal) as a response, in which case I think the grapher comments should be super clear that it has direct responsibility for maintaining some of Paladin's privacy characteristics.

core/go/noderuntests/pkg/domains/simple_domain.go
```diff
@@ -380,12 +375,9 @@ type simpleTokenParser struct {
 	Amount *ethtypes.HexInteger  `json:"amount"`
 }
 
-type SimpleDomainConfig struct {
-	SubmitMode string `json:"submitMode"`
```
matthew1001, line 384: Were these actually being used before?

core/go/internal/sequencer/coordinator/transaction/endorsing.go
```diff
@@ -115,6 +116,9 @@ func (t *coordinatorTransaction) sendEndorsementRequests(ctx context.Context) er
 		//this is done by emitting events rather so that this behavior is obvious from the state machine definition
 		t.scheduleRequestTimeout(ctx)
 		t.pendingEndorsementRequests = make(map[string]map[string]*common.IdempotentRequest)
+		// Notify the coordinator about endorser nodes discovered from the attestation plan so it can
+		// grow the originator node pool even when no candidates were pre-configured.
+		t.notifyEndorserNodes(t.extractEndorserNodes(ctx)...)
```
matthew1001: I think we discussed the issue of originators who send one transaction then have no further involvement but continue to receive heartbeats for the lifetime of the coordinator instance. Is the originator pool shrunk at any point to avoid that?

matthew1001: Sorry, paging back in other discussions. I think we landed on privacy groups (who I think the originator pool really only applies to?) being queryable wrt its members, and hence who the candidate coordinators are? So just wondering if the idea of the originator node pool eventually goes away, and if so, where that fits with these changes?

core/go/internal/sequencer/coordinator/transaction/events.go
```diff
@@ -400,3 +401,18 @@ func (*PreAssembleDependencyTerminatedEvent) Type() EventType {
 func (*PreAssembleDependencyTerminatedEvent) TypeString() string {
 	return "Event_PreAssembleDependencyTerminated"
 }
+
+// NotActiveCoordinatorEvent is received by the coordinator when the originator informs it
```
matthew1001, line 405: Just trying to parse this comment to understand the logic being added.

 - Originator (a different node potentially) sends this to someone who thinks they are the active coordinator (but they are not anymore)
 - Coordinator puts the TX back into pooled (it doesn't drop it at this point but keeps it in its own pool? Wondering why we keep it in the pool rather than remove it for the originator to redelegate in the future?)
 - `so it can be re-delegated to the correct coordinator` - isn't this something the originator should do, not the coordinator?

core/go/internal/sequencer/common/events.go
```diff
@@ -73,3 +79,80 @@ func (*TransactionStateTransitionEvent[S]) Type() EventType {
...
+
+func (*CoordinatorPriorityListUpdatedEvent) Type() EventType {
+	return Event_CoordinatorPriorityListUpdated
+}
+
+func (*CoordinatorPriorityListUpdatedEvent) TypeString() string {
+	return "Event_CoordinatorPriorityListUpdated"
+}
+
+// DelegationRejectedEvent is queued by the transport client to the originator when a
```
matthew1001: Again just wanting to parse this comment. This says that if `DelegationRequestAcknowledgement` with accepted = false arrives then a `DelegationRejectedEvent` is queued to the originator - but doesn't the `DelegationRequestAcknowledgement` itself arrived at the originator, not the coordinator?

core/go/internal/sequencer/common/events.go
```diff
@@ -73,3 +79,80 @@ func (*TransactionStateTransitionEvent[S]) Type() EventType {
...
+
+func (*DelegationRejectedEvent) Type() EventType {
+	return Event_DelegationRejected
+}
+
+func (*DelegationRejectedEvent) TypeString() string {
+	return "Event_DelegationRejected"
+}
+
+// HandoverRequestEvent is queued by the transport client to the coordinator when a
```
matthew1001: Because these events in in `common` it's not completely clear who is "receiving" and "sending" in the comments here. It might be helpful in the comments to clarify who is sending and who is receiving (as per my other comment above). I know the source of truth is the state machine, but if we're adding comments to describe behaviour it would help to make them a little clearer.

matthew1001 (commented):
Submitting initial comments, have ~20 files left to review which I'll submit a separate review for (browser felt like it was struggling with 120+ files + comments :) )

core/go/internal/sequencer/coordinator/transaction/confirming.go
```diff
@@ -205,6 +202,20 @@ func action_FinalizeOnChainedDependencyFailure(ctx context.Context, t *coordinat
 	return nil
 }
 
+// action_ConfirmTransactionInGrapher removes the transaction from the grapher's dependency
+// tracking and converts its locks to detached locks stamped with confirmedAtBlock.
+func action_ConfirmTransactionInGrapher(ctx context.Context, t *coordinatorTransaction, event common.Event) error {
```
matthew1001: I'll make an argument for `action_ConfirmTransaction` to minimise the need for state-machine related actions to spell out exactly what they're doing and which internal packages they're going to use.

core/go/internal/sequencer/coordinator/inactive.go
```diff
@@ -21,43 +21,45 @@ import (
...
-func action_Idle(_ context.Context, c *coordinator, _ common.Event) error {
-	c.coordinatorIdle(c.contractAddress)
+// action_UpdateActiveCoordinator updates the current active coordinator from either a received
+// heartbeat or a handover request. Both events carry the sender identity in different fields.
+func action_UpdateActiveCoordinator(_ context.Context, c *coordinator, event common.Event) error {
+	switch e := event.(type) {
+	case *common.HeartbeatReceivedEvent:
+		c.currentActiveCoordinator = e.From
+	case *common.HandoverRequestEvent:
+		c.currentActiveCoordinator = e.FromNode
```
matthew1001: Can we make these field names consistent?

core/go/internal/sequencer/coordinator/selection.go
```diff
@@ -17,91 +17,42 @@ package coordinator
...
-func (c *coordinator) updateOriginatorNodePool(originatorNode string) {
-	if !slices.Contains(c.originatorNodePool, originatorNode) {
-		c.originatorNodePool = append(c.originatorNodePool, originatorNode)
-	}
-	if !slices.Contains(c.originatorNodePool, c.nodeName) {
-		// As coordinator we should always be in the pool as it's used to select the next coordinator when necessary
-		c.originatorNodePool = append(c.originatorNodePool, c.nodeName)
-	}
-	slices.Sort(c.originatorNodePool)
+func guard_IsNewBlockRangeEpoch(_ context.Context, c *coordinator) bool {
```
matthew1001:
```suggestion
func guard_IsOnEpochBoundary(_ context.Context, c *coordinator) bool {
```

This feels a little clearer, because in my initial read I struggled with the concept of `IsNewBlockRange` - since we'll be in a new block range every N blocks.

core/go/internal/sequencer/coordinator/selection.go
```diff
@@ -17,91 +17,42 @@ package coordinator
...
+	// This is set when we update the block height, and remains valid until the next block height update
+	return c.newBlockRangeEpoch
```
matthew1001: Related, I wonder if `c.onEpochBoundary` or similar makes it clearer that this is `true` only infrequently.

core/go/internal/sequencer/originator/transaction/delegated.go
```diff
@@ -43,17 +66,25 @@ func action_SendPreDispatchResponse(ctx context.Context, txn *originatorTransact
...
 	if !ok {
 		log.L(ctx).Errorf("expected event type *AssembleRequestReceivedEvent, got %T", event)
 		return false, nil
 	}
 
-	log.L(ctx).Debugf("originator transaction validating assemble request - event coordinator %s, TX current delegate = %s", assembleRequestEvent.Coordinator, txn.currentDelegate)
-	return assembleRequestEvent.Coordinator == txn.currentDelegate, nil
-
+	if assembleRequestEvent.Coordinator != txn.currentDelegate {
+		// TODO: Should a validator be able to have a side effect like this?
```
matthew1001: Personally I don't like the idea that it changes things. Similarly for `guard_*`.

core/go/internal/sequencer/originator/originating.go
```diff
@@ -199,16 +181,51 @@ func (o *originator) removeTransaction(ctx context.Context, txnID uuid.UUID) {
...
+	e := event.(*common.CoordinatorPriorityListUpdatedEvent)
+	if len(e.Nodes) > 0 {
+		o.coordinatorPriorityList = e.Nodes
+	}
+	return nil
+}
+
+// action_UpdateActiveCoordinatorFromHeartbeat records the heartbeat sender as the current active
+// coordinator. Called only when the sender is in Active or Elect state.
+func action_UpdateActiveCoordinatorFromHeartbeat(_ context.Context, o *originator, event common.Event) error {
```
matthew1001, line 280: Just want to check the protection against this causing flip-flopping between nodes, where A -> HB -> B and B -> HB -> A, and the timings are such that they both just accept each others' heartbeats and stand down. I'm pretty sure the protection is there from your design doc and discussions, just seemed like this `action_*` function was taking the HB as gospel and doing what it says.

core/go/internal/sequencer/originator/state_machine.go
```diff
@@ -27,26 +27,26 @@ import (
...
 
-type State int
+type State = common.OriginatorState
 type EventType = common.EventType
 
 const (
-	State_Idle      State = iota //Not acting as a originator and not aware of any active coordinators
-	State_Observing              //Not acting as a originator but aware of a node (which may be the same node) acting as a coordinator
-	State_Sending                //Has some transactions that have been sent to a coordinator but not yet confirmed TODO should this be named State_Monitoring or State_Delegated or even State_Sent.  Sending sounds like it is in the process of sending the request message.
+	State_Initial   = common.OriginatorState_Initial   // Waiting for OriginatorCreatedEvent to fire initial coordinator selection
```
matthew1001, line 34: Something about `common.Originator...` feels a little odd :) `originator.State_Initial` or `common.State_Initial` feel easy to parse, but having originator/coordinator specific states in `common` does look a little odd. Related to my other comment about moving to originator/coordinator specific states where they are quite overlapping sets

core/go/internal/sequencer/originator/state_machine.go
```diff
@@ -55,13 +55,21 @@ type (
 )
 
 var stateDefinitionsMap = StateDefinitions{
-	State_Idle: {
+	State_Initial: {
```
matthew1001, line 59: I'm reading this change as `We no longer go back into this state from another one, so `State_Initial` is clearer/more accurate. I think if we go back into it then `State_Initial` feels a little too "first time only" for me. If we don't, then `State_Initial` seems fine

core/go/internal/sequencer/transport/transport_writer.go
```diff
@@ -38,13 +38,14 @@ type TransportWriter interface {
 	StartLoopbackWriter()
 	WaitForDone(ctx context.Context)
 	SendDelegationRequest(ctx context.Context, coordinatorNode string, transactions []*components.PrivateTransaction, blockHeight uint64) error
-	SendDelegationRequestAcknowledgment(ctx context.Context, delegatingNodeName string, delegationId string, transactionIDs []string, errors []int64) error
+	SendDelegationRequestAcknowledgment(ctx context.Context, delegatingNodeName string, delegationId string, transactionIDs []string, errors []int64, blockHeight uint64) error
+	SendDelegationRequestRejection(ctx context.Context, delegatingNodeName string, delegationId string, blockHeight uint64, activeCoordinator string) error
```
matthew1001, line 42: Just wonder if you can confirm/clarify the difference between `SendDelegationRequestRejection` and `SendDelegationRequestAcknowledgement`, since the comment on calls to the latter (see `coordinating.go`) says "Optionally errors can be returned". Is it that the former is for "You can't delegate anything at all to me" and the latter is "Some of these TXNs couldn't be accepted"?

matthew1001: I think now I've read `coordinator/state_machine.go` it looks like that's the case, so I wonder if `action_RejectDelegatedTransactions` should be `action_RejectDelegationRequest`. Currently it sounds like it might be a subset of them

core/go/internal/sequencer/statemachine/statemachine.go
```diff
@@ -571,8 +598,46 @@ func (sel *StateMachineEventLoop[S, E]) TryQueuePriorityEvent(ctx context.Contex
...
 func (sel *StateMachineEventLoop[S, E]) ProcessEvent(ctx context.Context, event common.Event) error {
-	return sel.stateMachine.ProcessEvent(ctx, sel.entity, event)
+	return sel.processEvent(ctx, event)
+}
+
+// DrainPendingEvents processes all events currently buffered in the priority and regular queues,
+// using the same pipeline as HandleEvent. It stops when both queues are empty.
+// Used in tests to flush internally-queued follow-up events (e.g. TransactionStateTransitionEvents
+// queued by transaction state machines during synchronous processing) before making assertions.
+func (sel *StateMachineEventLoop[S, E]) DrainPendingEvents(ctx context.Context) error {
```
matthew1001, line 608: It always feels a little icky to have external interfaces just for testing, but I realise it probably makes tests much easier to write reliably

core/go/internal/sequencer/coordinator/coordinator.go
```diff
@@ -64,38 +66,46 @@ type coordinator struct {
...
+	pooledTransactions                 []transaction.CoordinatorTransaction
+	currentBlockHeight                 uint64
+	dependencyTracker                  dependencytracker.DependencyTracker
+	grapher                            grapher.Grapher
+	nodePool                           []string // pool of nodes for coordinator priority list and heartbeat fan-out
+	coordinatorPriorityList            []string // priority-ordered list; index 0 is current active coordinator
+	newBlockRangeEpoch                 bool
+
+	signingIdentity     string
+	signingIdentityUsed bool // set when a transaction first retrieves the signing identity; cleared on key rotation
```
matthew1001: I'm wondering what the reason for the boolean is in general wrt to signing key rotation. I can see that `guard_SigningIdentityUsed` is using it, I think to avoid a flush or stop of the dispatch loop if the key hasn't been used. But if it hasn't, presumably the dispatch loop would be idle and there would be nothing to flush? I guess I'm just trying to work out what the logic would look like if there was no `signingIdentityUsed` at all and at rotation time the string just changes. Is it primarily here for a slight optimisation, and if so, is that optimisation necessary if the key hasn't been used? If we need to enforce doing something before key rotation, I'm not sure I'm completely convinced this extra bool is the only/best way we could do that, but feel free to convince me on all of the above :)

matthew1001 (commented):
Second lot of comments. Thanks for the great PR. Sorry it's a lot of comments and discussion, many of them are "please explain XYZ to me"

core/go/internal/sequencer/common/states.go
annamcallister: I'm not quite sure what you mean here since there was never a common set of state types. Prior to this change they were defined independently in each of the packages. I moved them all to the common package since originators needed to be able interpret the state in a coordinator heartbeat, but I just lifted and shifted the types as is.

core/go/internal/sequencer/common/selection.go
annamcallister: 👍 block range absolutely needs to be consistent across nodes. That should be easy enough to add into the description for field in the configuration reference, but is important enough to call out in the main sequencer docs too. Block height tolerance is the other one that I think is critical.

I don't think it is as fundamental for any of the others, including heartbeat intervals. It is definitely optimal to have the same configuration, but the protocol should be able to tolerate some differences.

core/go/noderuntests/pkg/domains/simple_domain.go
annamcallister: No - this was just a clean up rather than a change.

core/go/internal/sequencer/coordinator/transaction/endorsing.go
annamcallister: The latest code has moved to having a set of endorser candidates maintained separately from the a record of which originators we have interacted with recently and therefore should receive heartbeats.

core/go/internal/sequencer/coordinator/transaction/events.go
annamcallister: Good spot- this comment does not match the code at all now. When a coordinator receives this response it evicts the transaction. In order for an originator to send this response, it must have already redelegated the transaction to a different coordinator.

core/go/internal/sequencer/common/events.go
annamcallister: These events shouldn't really be in common anyway, as they're not comments- I'll move them to where they should be

core/go/internal/sequencer/originator/state_machine.go
annamcallister: I left a response to the other comment explaining the import rational behind the move to common. But I wonder if the answer here is to eliminate the `common` package and instead have a `types` package - for anything that is just a definition like this, and then functionality like idempotent request, engine integration has its own self describing package.

core/go/internal/sequencer/originator/state_machine.go
annamcallister: `State_Initial` is the pattern we've used across other state machines to capture the idea that there is a one time first event that the state machine receives and takes an action in response to, and only when its done that can it move into states where it can process real work. So it's definitely a one time only state. On last review I had actually removed that first time action so this transition didn't make sense, but I've since added one back in.

core/go/internal/sequencer/transport/transport_writer.go
annamcallister: > Is it that the former is for "You can't delegate anything at all to me" and the latter is "Some of these TXNs couldn't be accepted"?
 
 Yes - and agree with the proposed name change

core/go/internal/sequencer/statemachine/statemachine.go
annamcallister: I sort of justified this one to myself with the fact that an external consumer _could_ chose to use this as a non test function. It just happens to be that the way we use it today, this capability is only needed for tests.

I'm in two minds as to whether the comments on a component like this which talk about how it is used are actually useful. I think they likely aren't and should be cleaned up. Particularly as I'm wondering whether in the longer run this state machine is a component that could move to the `toolkit` and be used far more widely.

core/go/internal/sequencer/coordinator/coordinator.go
annamcallister: After our in person discussion I've switched to
* keeping the key and its used value in a struct so they are clearly tied together
* always rotating the key, but avoiding the flush if the key has never been used

core/go/internal/sequencer/originator/originating.go
annamcallister: > the timings are such that they both just accept each others' heartbeats and stand down

I don't think this is possible. There's a few parts to this:

* A coordinator doesn't just stand down because it receives a HB from another active node, it only stands down if the heartbeat it sees is from a node that is higher priority than itself. So unless we're in a very strange uneven indexing situation (where nothing is likely to work anyway), only one node should stand down.
* This action is in the originator, not the coordinator. Additionally it is only called in `Idle` and `Observing` states. Even if there is some sort of flip flopping while the coordinators attempt to work it out between them, the impact of setting the active coordinator from this action is that it is the first node that the originator attempts delegation to if it moves to `Sending`. Most of the time this will simply be the agreed active coordinator. If there is contention still going on when the originator is in `Sending`, the originator starts using its own view on priority to decide whether it should redelegate to another node or not.

core/go/internal/sequencer/common/states.go
annamcallister: Discussed in person that for now they will stay like this.

core/go/internal/sequencer/coordinator/grapher/grapher.go
annamcallister: I've renamed so we have
* `ForgetTransaction` (formerly `ConfirmTransaction`)
* `ForgetTransactionAndLocks` (formerly `ForgetTransaction`)

core/go/internal/sequencer/coordinator/grapher/grapher.go
annamcallister: I think this does need to go into a separate PR for the sake of trying to get closure here, but it should be addressed ASAP.

core/go/internal/sequencer/common/events.go
annamcallister: The intention was to say that it arrives at the originator via the transport client. I.e. there is a message type called `DelegationRequestAcknowledgement` which the transport client translates into a `DelegationRejectedEvent` if the request was not accepted. 

I've updated the comment- I've also renamed the message type from acknowledgement to response in keeping  with the request/response pattern that other message types use as this has been causing me confusion

core/go/internal/sequencer/coordinator/transaction/confirming.go
annamcallister: Agree, I've done some refactoring around the other actions that the state machine does for confirmed (success and reverted) transactions so that we can have clearer action names that read more like an external spec

core/go/internal/sequencer/originator/transaction/delegated.go
annamcallister: I've changed this now.

core/go/internal/sequencer/statemachine/statemachine.go
annamcallister: Comments updated to not indirectly doc other packages.

core/go/internal/sequencer/coordinator/transaction/endorsing.go
matthew1001: Thanks for confirmation & in-person discussion. Summarising for posterity:
 - originator node pool is now totally detached from candidate coordinators
 - domain config is queryable for list of endorsers (not the same as privacy group members, but coincidentally the same for Pente)
 - originator node pool is now solely for heartbeats for originator status of their TXNs

core/go/internal/sequencer/coordinator/grapher/grapher.go
matthew1001: Do you mind raising a V1 item for this so we don't lose track of it? We can decide if it's must-do, but at least putting it on the V1 board forces the conversation

core/go/internal/sequencer/originator/originating.go
matthew1001: Thank you for the detailed confirmation 👍

core/go/internal/sequencer/originator/state_machine.go
matthew1001: I'd be happy with that approach, and for it to land in another PR if that's your preference

core/go/noderuntests/coordinationtest/coordination_test.go
```diff
@@ -1845,3 +1847,121 @@ func TestTransactionFailsIfExplicitPrereqTransactionFails(t *testing.T) {
 	)
 }
 
+func TestCoordinatorFailover(t *testing.T) {
+	// Test that when the preferred (highest-priority) coordinator goes offline, the non-preferred
+	// node takes over coordination, and that when the preferred node restarts it does not disrupt
+	// the non-preferred coordinator that is already active.
+
```
matthew1001, line 1855: 👍

core/go/internal/sequencer/coordinator/snapshot.go
```diff
@@ -21,23 +21,37 @@ import (
...
+	"github.com/LFDT-Paladin/paladin/toolkit/pkg/prototk"
 )
 
 func action_SendHeartbeat(ctx context.Context, c *coordinator, _ common.Event) error {
 	return c.sendHeartbeat(ctx, c.contractAddress)
 }
 
-// sendHeartbeat builds the base snapshot once, then sends a per-node copy to each node in the
-// originator pool. In Active_Flush/Closing_Flush/Closing states, the grapher is queried per-node:
+// sendHeartbeat builds the base snapshot once, then sends a per-node copy to each heartbeat
```
matthew1001, line 31: With the extra complexity we have here now, I think the introductory comment needs to be a little clearer about the different categories of heartbeat. I think they are:
 - Heartbeats with base snapshot data (into which nothing private goes)
 - Heartbeats with the base data, plus locks and private states

In fact I think my preference for this function is to split it into 2 for complete clarity. Maybe we introduce terms like "basic heartbeat" and "full heartbeat", or something else that appropriately distinguishes which is being set and when (and to whom). This really relates to the comments I made in the grapher about wanting to make it as difficult as possible to accidentally regress behaviour that results in private data going where it shouldn't, but also it's about clarity of what the sequencer uses heartbeating for

core/go/internal/sequencer/originator/transaction/state_machine.go
```diff
@@ -209,16 +212,18 @@ var stateDefinitionsMap = StateDefinitions{
...
-						If:     statemachine.GuardNot(guard_AssembleRequestMatchesPreviousResponse),
-						Action: action_AssembleAndSign,
+						Validator: validator_AssembleRequestMatches,
+						If:        statemachine.GuardNot(guard_AssembleRequestMatchesPreviousResponse),
+						Action:    action_AssembleAndSign,
 					},
 					{
-						If:     guard_AssembleRequestMatchesPreviousResponse,
-						Action: action_ResendAssembleSuccessResponse,
+						Validator: validator_AssembleRequestMatches,
```
matthew1001, line 224: I don't know if we have it already, but clarifying somewhere when a `guard` should be used vs a `validator` would be useful, for future contributions to the state machine.

matthew1001 (approved):
Thanks for the revisions Anna. I've added a few remaining comments but they're really about adding clarity in comments and/or function structure

core/go/internal/sequencer/coordinator/snapshot.go
annamcallister: Will put into a new PR so I can get the state machine refactor linked below in first.

core/go/internal/sequencer/originator/transaction/state_machine.go
annamcallister: https://github.com/LFDT-Paladin/paladin/pull/1191 is addressing this and other state machine structural issues

PR #1179, opened 2026-05-13
Core test coverage

annamcallister:
Increases UT coverage of all packages that contribute towards core coverage but are not part of the sequencer to 100%.


matthew1001 (approved):
This generally looks good. There are a few product changes to help with coverage, which  I'd hope we could keep to a minimum across the codebase to achieve 100% coverage.

I don't know if further product refactoring would make 100% coverage possible without test-specific fields, so I'm happy this goes in like this for the time being but think we should try to minimise it where possible

Issue #1006, opened 2026-02-11
logging: Remove non-NLS error/warning/info messages from across the codebase


matthew1001:
### What would you like to be added?

We should check for all cases of `msg := fmt.Sprintf` and go through them, adding custom translated error messages for most/all of them

### Why is this needed?

To ensure we don't have untranslated error messages.

biswas2200:
Hey @matthew1001 ! I'm Biswas, student applying for the LFX Mentorship Program (Java SDK for Paladin). I'd like to contribute to this issue — just wanted to check if it's still active before diving in.

If so, is there a preferred approach or style guide for the NLS messages I should follow? Happy to get started!

Thanks 🙏

annamcallister:
Moving to V1.1 board as this is a relatively small task which will help with code hygiene and logging visibility

PR #1182, opened 2026-05-18
Wrap sequencer errors with MsgSequencer* errors

matthew1001:
A number of `MsgSequencer*` error codes weren't being used to wrap errors.

  - I've also removed `MsgSequencerStateHashContention` as there is a new `MsgSequencerGrapherAddMinterAlreadyExistsError`
     - As a result I've bumped the error codes down by 1 (actually by 2 since there was a gap in the code sequence)
     - Since these are new to V1 it's the last opportunity we have to bump them without breaking users

Relates to https://github.com/LFDT-Paladin/paladin/pull/1181


core/go/internal/sequencer/transport_client.go
```diff
@@ -77,12 +79,12 @@ func (sMgr *sequencerManager) HandlePaladinMsg(ctx context.Context, message *com
 	case transport.MessageType_TransactionUnknown:
 		go sMgr.handleTransactionUnknown(sMgr.ctx, message)
 	default:
-		log.L(ctx).Errorf("Unknown message type: %s", message.MessageType)
+		log.L(ctx).Errorf("%s: %s", i18n.NewError(ctx, msgs.MsgSequencerInvalidMessage), message.MessageType)
```
annamcallister: should `message.MessageType` be a parameter in the error message?

core/go/internal/sequencer/transport_client.go
```diff
@@ -77,12 +79,12 @@ func (sMgr *sequencerManager) HandlePaladinMsg(ctx context.Context, message *com
...
 	}
 }
 
 func (sMgr *sequencerManager) logPaladinMessageUnmarshalError(ctx context.Context, message *components.ReceivedMessage, err error) {
-	log.L(ctx).Errorf("<< ERROR unmarshalling proto message%s from %s: %s", message.MessageType, message.FromNode, err)
+	log.L(ctx).Errorf("<< ERROR unmarshalling proto message%s from %s: %s", message.MessageType, message.FromNode, i18n.WrapError(ctx, err, msgs.MsgSequencerParseFailed))
```
annamcallister: Why is only part the error log put into an i18n error?

annamcallister (changes requested):
I've added a couple of comments on the errors only used for logs.

core/go/internal/sequencer/transport_client.go
matthew1001: I might actually drop this change from the PR. For now it's probably useful to keep the `<< ...` format so we probably want to decide whether to introduce coded error messages for all of those cases in this file

core/go/internal/sequencer/transport_client.go
matthew1001: Yeah, I've added in the latest commit

matthew1001:
I decided to put back into draft because it makes sense to do more thorough review of the non-NLS'd error logs in the sequencer and close https://github.com/LFDT-Paladin/paladin/issues/1006 once the PR merges. It also means we can put the merge effort on this PR once the block-aware work is in `main`.

PR #1183, opened 2026-05-18
JSON/RPC calls should return HTTP 200 in all cases except auth error and internal error

### LINKED ISSUE #1001: bug: idempotency conflicts return an HTTP 500

matthew1001:
Fixes https://github.com/LFDT-Paladin/paladin/issues/1001

This PR changes the way JSON/RPC responses are handled wrt HTTP status codes. It makes Paladin compliant with https://www.jsonrpc.org/specification#error_object

 - Non error cases are returned as HTTP `200`
 - Auth error cases are returned as HTTP `403`
    - Currently the only auth error paths are cases where an authenticated user is forbidden from accessing a resource, so there are no cases where `401` is returned currently.
 - Internal server errors are returned as HTTP `500`
 - All other error cases are returned as HTTP `200` with a suitable Paladin error in the response body

With regard to existing applications using the SDKs:

 - The GO SDK already correctly handled non-`2xx` HTTP responses, and 2xx HTTP responses containing an error so no changes are required to avoid breaking existing applications
 - The Typescript SDK didn't handle HTTP `2xx` responses that contained error bodies, so it has been updated to throw an error for those cases, keeping the same behaviour as `500` responses.

Since this has the potential to affect existing user applications, a new RPC server config option `legacyReturnCodes` has been added so that there is a way to revert to the previous behaviour for the time being. We would ideally remove this in a future release once there has been sufficient implicit testing of the new behaviour.

I've added a test that specifically exercises the issue highlighted in https://github.com/LFDT-Paladin/paladin/issues/1001

Below is the result of running the `notarized-tokens` sample for a second time, against a node with this PR, where I modified the sample to set a fixed string on the `mint(...)` call:

```
Initializing Paladin clients from the environment configuration...
Step 1: Deploying a Noto cash token...
Noto cash token deployed successfully!
Step 2: Minting 2000 units of cash to Node1...
JSON-RPC error from ptx_sendTransaction (200 OK) PD012220: idempotencyKey already used by submitted transaction mint-cash-tokens=4c13d6f8-5bd4-41e7-9610-195925c47485
Exiting due to an uncaught error: AxiosError: PD012220: idempotencyKey already used by submitted transaction mint-cash-tokens=4c13d6f8-5bd4-41e7-9610-195925c47485
    at PaladinClient.post (/Users/mwhitehead/github-kaleido/paladin/sdk/typescript/build/paladin.js:543:28)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Object.sendTransaction (/Users/mwhitehead/github-kaleido/paladin/sdk/typescript/build/paladin.js:59:29)
    at async TransactionFuture.waitForReceipt (/Users/mwhitehead/github-kaleido/paladin/sdk/typescript/build/transaction.js:14:44)
    at async main (/Users/mwhitehead/github-kaleido/paladin/examples/notarized-tokens/src/index.ts:78:23) {
  code: '-32001',
```

 - Response code is `200 OK`
 - Code is correctly set to `-32001` (RPCCodeConflict)

Here is the same modified sample run (twice) against Paladin with `legacyReturnCodes: true` configured:

```
Initializing Paladin clients from the environment configuration...
Step 1: Deploying a Noto cash token...
Noto cash token deployed successfully!
Step 2: Minting 2000 units of cash to Node1...
JSON-RPC error from ptx_sendTransaction (500 Internal Server Error) PD012220: idempotencyKey already used by submitted transaction mint-cash-tokens=120c6de7-6807-465a-b39c-2666871b8826
Exiting due to an uncaught error: AxiosError: Request failed with status code 500
```


awrichar:
Does this supersede https://github.com/LFDT-Paladin/paladin/pull/1147?

matthew1001:
> Does this supersede #1147?

Yeah I discussed this with @shorsher having not realised he had raised a PR for it. I've included a config option in this PR that addresses Anna's comment https://github.com/LFDT-Paladin/paladin/pull/1147#issuecomment-4253824383 and so for now I might propose we move ahead with this one. It wasn't really intentional to supersede #1147 but that might be the best approach this time.

toolkit/go/pkg/rpcserver/rpcmethod.go
annamcallister, line 69: It's a bigger refactor- but I think that this impl (and the different param number variants) needs to be able to specify the the code to return. It doesn't feel right to have a server in a toolkit interpreting specific PD code

toolkit/go/pkg/rpcserver/rpcmethod.go
```diff
@@ -169,7 +170,11 @@ func mapResponse(ctx context.Context, req *rpcclient.RPCRequest, result interfac
 		}
 	}
 	if code == 0 {
-		code = rpcclient.RPCCodeInternalError
+		if strings.Contains(err.Error(), "PD012220") {
+			code = rpcclient.RPCCodeConflict
```
annamcallister: Comment above about not putting code mapping in this package

sdk/go/pkg/rpcclient/client.go
```diff
@@ -43,6 +43,7 @@ const (
 	// JSON-RPC 2.0 specification reserves -32000 to -32099 for "implementation-defined server-errors"
 	// Paladin uses this range for custom application errors like authentication failures
 	RPCCodeUnauthorized RPCCode = -32000 // Unauthorized request - authentication failed
+	RPCCodeConflict     RPCCode = -32001 // Idempotency key clash - request already submitted with this key
```
annamcallister: Give comments below about where the mapping for these codes needs to happen- I wonder if these actually belong in the paladin client package and not the generic rpc client package?

toolkit/go/pkg/rpcserver/rpcserver.go
```diff
@@ -162,17 +167,57 @@ func (s *rpcServer) httpHandler(res http.ResponseWriter, req *http.Request) {
...
+	func() {
+		defer func() {
+			if rc := recover(); rc != nil {
+				log.L(ctx).Errorf("Panic in RPC handler: %v", rc)
+				r = handlerResult{httpStatus: http.StatusInternalServerError, sendRes: true,
+					res: rpcclient.NewRPCErrorResponse(fmt.Errorf("%v", rc), nil, rpcclient.RPCCodeInternalError)}
+			}
+		}()
+		r = s.rpcHandler(ctx, req.Body, nil /* not websockets */)
+	}()
```
annamcallister, line 180: nice 👍

annamcallister (changes requested):
Looks good but some refinement needed in generic rpc vs paladin client/server behaviour

toolkit/go/pkg/rpcserver/rpcmethod.go
matthew1001: I've started working towards doing that refactor, but before I carry on with updating tests etc. could you take a look at https://github.com/LFDT-Paladin/paladin/pull/1183/changes/4232e1d6db214c470f41f3f678f3e422b430727a to see if that looks like the direction you were expecting? It adds a new `rpcclient` package to the toolkit, alongside the existing `rpcserver` package, and changes the rpcmethod variants to return an RPC error code.

sdk/go/pkg/rpcclient/client.go
```diff
@@ -40,9 +40,6 @@ const (
 	RPCCodeParseError     RPCCode = -32700
 	RPCCodeInvalidRequest RPCCode = -32600
 	RPCCodeInternalError  RPCCode = -32603
```
annamcallister, line 42: These could be redeclared through pldclient package as well for completeness

toolkit/go/pkg/rpcserver/rpcmethod.go
```diff
@@ -84,6 +84,18 @@ func RPCMethod1[R any, P0 any](impl func(ctx context.Context, param0 P0) (R, err
 	})
 }
 
+func RPCMethod1WithRPCCode[R any, P0 any](impl func(ctx context.Context, param0 P0) (R, rpcclient.RPCCode, error)) RPCHandler {
```
annamcallister, line 82: For completeness we should have the full 0-N versions of these functions.

I think we could also avoid duplication by having wrapping the coded version to get the non coded version, something along the lines of:

```
func RPCMethod1[R any, P0 any](impl func(ctx context.Context, param0 P0) (R, error)) RPCHandler {
    implWithCode := func(ctx context.Context, param0 P0) (R, rpcclient.RPCCode, error) {
      r, err := impl(ctx, *param0)
      return r, 0, err
    }
    return RPCMethod1WithRPCCode(implWithCode)
}
```

annamcallister (commented):
I've made a couple more suggestions for completeness.

Have you considered whether there's any behaviour impact to https://github.com/LFDT-Paladin/paladin/blob/main/sdk/go/pkg/pldclient/txbuilder.go we should account for?

toolkit/go/pkg/rpcserver/rpcmethod.go
matthew1001: Thanks, I've addressed the 2nd part and collapsed the two calls so one calls the other. Re

> For completeness we should have the full 0-N versions of these functions.

I intentionally left them out while we have almost no uses for methods that return specific codes. If you're really keen to have them all in ready to use in the future I can add them in but it was intentional not to overbloat the code until they are needed

annamcallister: There's a tension here between toolkit as a library - independent from the core package - and the fact that its usage is exclusively within the core package.

I would have leant towards adding them all but I don't feel that strongly about it. They can be added when needed.

sdk/go/pkg/rpcclient/client.go
matthew1001: Sorry I missed this comment. Address in the latest commit.

toolkit/go/pkg/rpcserver/rpcmethod.go
matthew1001: I think I'd vote for leave minimal for now until the next time a custom code is needed - and next time we add another *WithRPCCode version we add them all?

matthew1001:
> I've made a couple more suggestions for completeness.
> 
> Have you considered whether there's any behaviour impact to https://github.com/LFDT-Paladin/paladin/blob/main/sdk/go/pkg/pldclient/txbuilder.go we should account for?

Sorry I missed this comment earlier.

Yes good spot, we can indeed update txbuilder to check for the custom error code. I've done that in the latest commit. However I don't think we can remove the old message string check for now, since new SDK connected to old Paladin will need to be able to support checking the old way for a while.

annamcallister (approved):

PR #1185, opened 2026-05-20
Create a GHA to run longer running tests on branches

matthew1001:
As part of making pre-release testing more standardized, and to make that testing easier to run and re-run, this github action runs the chosen longer-running test, for a configurable period of time.

Other configuration changes can be made by including a change to the test script in the branch being tested.

This is only a first version of a GHA-based test run, which I suspect we will want to iterate on over time.


annamcallister (approved):
Approving as a first delivery needs to go to main in order to be able to start iterating on this action in branches

PR #1189, opened 2026-05-20
Zeto V1 support with multi-version support

### LINKED ISSUE #1193: Enhance Paladin Zeto domain to support multiple Zeto versions with v0.5.0 (V1 contract interface)
Zeto release v0.5.0 introduced the updated locking capability as ILockableCapability.sol, which is shared with Noto. Paladin's Zeto domain should be enhanced to support multiple Zeto versions and specifically the new locking capabilities, the updated locked UTXO transfer mechanism (which consumes the locked UTXOs with commitments rather than with nullifiers).

Note that migration from one version to a newer version in-place is out of scope for this PR. This is not yet supported by Zeto between releases.

jimthematrix:
resolves #1193

## Summary

This PR adds **multi-version Zeto support** so Paladin can run **V0** (upstream zeto-contracts ~v0.2.2) and **V1** (upstream ~v0.5.1) side by side in one deployment, and completes **fungible V1 lock lifecycle** support aligned with upstream `ZetoLockable / ILockableCapability`.

127 files changed (+14,597 / −2,835). Major themes:

- Three independent version axes (fungible domain ABI, target token core ABI, Paladin factory)
- V1 lock handlers: `createLock`, `spendLock`, `cancelLock` with persisted commitment preimages
- Version-aware event dispatch, artifact download, and integration tests
- SMT storage moved to shared `toolkit/go/pkg/smt` with `LFDT-Paladin/smt` dependency
- `ZetoFactoryV1` deployed behind an `ERC1967Proxy`; build/toolchain updates for Go 1.24 and TS 6.x

| Axis | V0 | V1 |
| --- | --- | --- |
| (1) Fungible domain ABI | `IZetoFungible.json` — legacy `lock`, `transferLocked` | `IZetoFungible_V1.json` — `createLock`, `spendLock`, `cancelLock` |
| (2) Target token core ABI | `IZeto.json` (~v0.2.x events/core) | `IZeto_V1.json` (~v0.5.x events/core) |
| (3) Paladin factory | `ZetoFactory.sol` | `ZetoFactoryV1.sol` (upgradeable proxy) |

### Domain config & registration

- `domain_config_codec.go` — V1 configs prefixed with `ZetoDomainConfigID_V1` and ABI-encoded tuple (`zetoVariant`, `factoryVersion`, `circuitBundleId`, …); legacy configs unchanged.
- `versions.go` — documents and implements the three version axes; helpers for V1-specific behaviour (`UseZetoOnchainPackedProofCalldata`, `LockTransitionVerifierOutputOrderLockedFirst`).

### Event ABI dispatch

- `events_abi.go` — selects `IZeto.json` vs `IZeto_V1.json` and merges lockable/KYC events per `ZetoTargetContractABIVersion`.
- Handler registration and prepare/assemble paths branch on `DomainInstanceConfig` rather than assuming a single Zeto generation.

### Build & upstream artifacts

- `domains/zeto/build.gradle` — `zetoVersions` table for **v0.2.2** and **v0.5.1** (wasm, proving keys, contracts); Gradle tasks download/extract per version and copy renamed `*_V1` ABIs for V1.
- Removed checked-in `internal/zeto/factoryabis`; ABIs now come from extracted release artifacts.
- Root `build.gradle`, examples, operator, and `release-solidity-contracts` workflow updated to resolve ZKP paths by version.
- `helpers/zeto_zkp_versions.go` — `PALADIN_ZETO_ZKP_VERSION` / `PALADIN_ZETO_ZKP_MATRIX` for integration-test artifact selection.

### V1 on-chain proof calldata

- `common/onchain_proof_bytes.go` — packs Groth16 proof + public inputs into a single ABI `bytes proof` blob for V1 `transfer` / `deposit` / `withdraw` / lock unlock paths (upstream ~v0.5.x shape).
- V0 keeps discrete proof tuple + public fields in Prepare.

## V1 lock lifecycle

Upstream V1 pools use `ILockableCapability` on-chain while Paladin exposes a thinner domain API via `IZetoFungible_V1`.

### Lock info schema (`ZetoLockInfoState`)

- Consolidated `LockedOutputs` (shared collateral commitments for both unlock paths).
- Added `SpendOutputs`, `SpendData`, `CancelOutputs`, `CancelData`, `SpendCommitment`, `CancelCommitment` so off-chain state mirrors on-chain `computeSpendHash` / `computeCancelHash` and later `spendLock` / `cancelLock` assembly.
- `zeto_lock_commitment.go` — domain-separated commitment hashing matching `zeto_lockable._buildUnlockHash`.
- `lock_id.go` — `keccak256(abi.encode(pool, publicCreateLockMsgSender, txId))` matching on-chain lock IDs.

### Handlers

| Handler | Notes |
| --- | --- |
| `createLock` | Pre-pins spend/cancel outputs; computes real `SpendCommitment` / `CancelCommitment` at create time |
| `spendLock` | Uses `SpendData` in inner `ZetoSpendLockArgs.data`; shared unlock plumbing via `lock_unlock_common.go` |
| `cancelLock` | Mirrors `spendLock` with cancel preimage fields; registered for V1 only (`METHOD_CANCEL_LOCK`) |

Legacy V0 `lock` / `transferLocked` / `delegateLock` handlers remain for existing pools.

## SMT toolkit refactor

- Moved Paladin SMT state storage from `domains/zeto/internal/zeto/smt` to **`toolkit/go/pkg/smt`** (node/root/spec/storage + tests).
- Zeto domain now depends on **`github.com/LFDT-Paladin/smt v0.2.0`** for tree operations; Paladin-specific persistence stays in toolkit.
- Added SMT error messages to `common/go/pkg/pldmsgs/en_errors.go`.

## Solidity & deployment

- **`IZetoFungible_V1.sol`** — Paladin domain methods: `createLock`, `spendLock`, `cancelLock`.
- **`ZetoFactoryV1.sol`** — wraps `ZetoTokenFactoryUpgradeable` from zeto-contracts v0.5.1; forwards registration `data` verbatim (V1 metadata encoded off-chain).
- Integration/operator deploy uses **`ERC1967Proxy`** + `initialize` (see `helpers/zeto_factory_v1_deploy.go`).
- **`Atom.sol`** — pragma bump for TS 6.x / OpenZeppelin compatibility.

### Integration

- **`zeto_fungible_v1_test.go`** — full V1 lock scenario: three `createLock`s, `spendLock`, `delegateLock` + `spendLock`, `cancelLock`; validates on-chain `lockId` against keccak formula.
- **`zeto_dual_version_test.go`** — V0 (`v0.2.2`) + V1 (`v0.5.1`) `Zeto_AnonNullifier` in one testbed (`config-for-deploy-anon-nullifier-dual.yaml`).
- Existing fungible/non-fungible suites updated for version labels, deploy YAMLs, and artifact path resolution.
- **`config-for-deploy-fungible-v1.yaml`** — V1 fungible deploy config.

### Unit & component

- Extensive handler/unit coverage for lock handlers, events, deploy factory, domain config codec, SMT storage.
- **`operator/contractpkg/contract_map.json`** — added missing `SmtLib` linked contract for `Zeto_AnonNullifier` operator tests.
- Component test dependency refresh (`core/go/go.mod`).

## Toolchain & dependencies

- Go **1.24.4** / toolchain **1.24.13** across modules; dependency bumps (gRPC, testify, firefly-common, etc.).
- **`solidity/tsconfig.json`** — explicit `rootDir` for TypeScript 6.x.
- **`Dockerfile`** — Go 1.24.4.

## Out of scope

V1 level support for the non-fungible token implementations, primarily around the ILockableCapability interface enablement, to be added in a follow-up PR.


annamcallister:
Go and toolchain updates to 1.24/1.24.13 and dependency bumps superseded by #1270 which moved whole project to 1.25 and is merged into this branch.

solidity/contracts/shared/Atom.sol
```diff
@@ -2,10 +2,10 @@
 pragma solidity ^0.8.20;
 
 import {Address} from "@openzeppelin/contracts/utils/Address.sol";
+import {ReentrancyGuard} from "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
 import {Initializable} from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
-import {ReentrancyGuardUpgradeable} from "@openzeppelin/contracts-upgradeable/utils/ReentrancyGuardUpgradeable.sol";
 
-contract Atom is Initializable, ReentrancyGuardUpgradeable {
+contract Atom is Initializable, ReentrancyGuard {
```
annamcallister, line 10: It's not totally clear to me why we've moved away from the upgradeable version here. Possibly because we deploy Atoms behind a proxy because of gas efficiency rather than because we have any intention of upgrading them (and don't implement an upgrade function)?

It would be good to capture the reasoning in a comment on this contract so its clearer in future.

solidity/contracts/domains/zeto/ZetoFactoryV1.sol
annamcallister, line 1: I think the file naming/versioning might need some work here.

The existing contracts are
* `ZetoFactory_V0.sol` - the very first zeto factory
* `ZetoFactory.sol` - the `v1` factory (following conventions set by noto where the latest version does not have a version in the contract name) - where the change forcing the new version is that the factory has become upgradable. 

So if we need a new factory contract we should have
* `ZetoFactory_V0.sol` - stays the same
* `ZetoFactory_V1.sol` - the current `ZetoFactory.sol` renamed
* `ZetoFactory.sol` - the current  `ZetoFactoryV1.sol` which is actually the V2 factory

However, I also notice that the interface on the new factory hasn't changed from V1; the difference is that it deploys `0.5.1` tokens. If the zeto domain has multi version support then the factory could also have multi version support. The latest noto factory handles deploying `v2` noto as the default implementation, and the ability to register other implementations to deploy earlier versions. I don't think this is as straight forward to achieve in zeto though?

As a side note: the actual interface for deploying tokens did not change between v0 and v1, and maybe shouldn't have taken a major factory version, instead being a v0.1, or even going as far as replacing v0. Because there was no interface change, nothing is depending on this being v1, so we could choose a different approach as part of this PR.

solidity/contracts/domains/interfaces/IZetoFungible_V1.sol
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.20;
+
+/**
+ * @title IZetoFungibleV1
+ * @dev ABI for the Zeto domain transaction interface, specifically implemented in Go.
```
annamcallister, line 6: I think this is right that the new version is `V1` - it's the V1 of the Paladin zeto domain transaction interface (which today can be backed by v0.5.1 zeto), although it doesn't immediately sit right and needs a while to think through.

I think it would be helpful to have a docs update to explain the relationship between zeto versions and Paladin zeto domain versions (and possible factory versions).

Do we have a guaranteed future relationship between `v1` and `v0.5.1`? I think the answer is no. We might support future zeto versions against the `v1` interface, and may even have a `v2` interface which could be supported by `v0.5.1`. This adds an extra dimension to versioning which hasn't been necessary for noto and I wonder if this also needs to feed into the factory discussion in my other comment- it probably depends on where we land with multi version support in the factory?

annamcallister: There's also an inconsistency with file naming to noto, which would have used  `IZetoFungible.sol` for the new file and renamed the old one to `IZetoFungible_V0.sol`.

I wonder whether we do a bigger rename to put version numbers into all current and previous versions of every file, so no version (oldest or latest) takes the unversioned name.

annamcallister (commented):
An initial review done on the solidity changes.

solidity/contracts/domains/zeto/ZetoFactoryV1.sol
annamcallister: I think the minimum here would be the suggested rename. 

> So if we need a new factory contract we should have
> 
> ZetoFactory_V0.sol - stays the same
> ZetoFactory_V1.sol - the current ZetoFactory.sol renamed
> ZetoFactory.sol - the current ZetoFactoryV1.sol which is actually the V2 factory

perhaps differing from Noto and putting the `V2` in the latest factory name? - `ZetoFactory_V2.sol`

Multi version factory is a nice to have but probably down to the domain whether it is supported or not rather than a requirement Paladin puts on domains.

annamcallister:
Observation that https://github.com/LFDT-Paladin/smt specifies a minimum go version of 1.26 but Paladin is currently running 1.25

.github/workflows/release-solidity-contracts.yaml
```diff
@@ -90,8 +90,8 @@ jobs:
       run: |
         mkdir -p tools/contracts
 
-        # Download zeto contracts (downloads to domains/zeto/tools/zeto-contracts-v0.2.0.tar.gz)
-        ../../gradlew downloadZetoCompiledContracts
+        # Download primary zeto contracts archive (see domains/zeto/build.gradle zetoVersions)
```
annamcallister, line 93: I'm curious as to why v0.2.2 is still the default. Is there a plan at a later date to move to v0.5.1 being the default?

For anyone with an existing deployment they will have the v0.2.2 zeto factory already deployed on chain; the move to v0.5.1 will only come when they upgrade their factory implementation. The contracts version we package is most relevant to new deployments, and I would have expected we'd want to be encouraging these onto the latest version?

operator/build.gradle
```diff
@@ -64,6 +55,22 @@ ext {
     paladinImageTag = project.hasProperty('paladinImageTag') ? project.paladinImageTag : 'test'
 
     artifactDir = project.hasProperty('artifactDir') ? project.artifactDir : 'artifacts'
+
+    // Zeto contract ABI export from :domains:zeto (see zetoArtifactConfigurationName). Default: zetoPrimaryArtifacts.
+    // Override release: ./gradlew :operator:copyZetoSolidity -PzetoZkpRoot=v0.5.1  →  zetoArtifacts_v0_5_1
```
annamcallister, line 60: Does this command result in the right zeto factory being installed or just the other contracts?

examples/zeto/build.gradle
```diff
@@ -30,9 +30,11 @@ dependencies {
     buildCommon project(path: ':examples:common', configuration: 'buildCommon')
 }
 
-task copyContracts(type: Copy, dependsOn: [":domains:zeto:extractZetoArtifacts"]) {
-    from fileTree("../../domains/zeto/zkp/artifacts/contracts") {
-        include '**/SampleERC20.json'
+task copyContracts(type: Copy, dependsOn: [":domains:zeto:extractZetoPrimaryZkp"]) {
+    from {
+        fileTree("../../domains/zeto/zkp/${project(':domains:zeto').ext.zetoVersion}/artifacts/contracts") {
```
annamcallister, line 35: I don't think this is actually configurable by `zetoVersion` since the task depends on `:domains:zeto:extractZetoPrimaryZkp`, plus the example uses the V0 API?

examples/private-stablecoin/build.gradle
```diff
@@ -30,10 +30,12 @@ dependencies {
     buildCommon project(path: ':examples:common', configuration: 'buildCommon')
 }
 
-task copyContracts(type: Copy, dependsOn: [":domains:zeto:extractZetoArtifacts"]) {
-    from fileTree("../../domains/zeto/zkp/artifacts/contracts") {
-        include '**/SampleERC20.json'
-        include '**/IZetoKyc.json'
+task copyContracts(type: Copy, dependsOn: [":domains:zeto:extractZetoPrimaryZkp"]) {
+    from {
+        fileTree("../../domains/zeto/zkp/${project(':domains:zeto').ext.zetoVersion}/artifacts/contracts") {
```
annamcallister, line 35: same comment as the other example regarding version configurability

build.gradle
```diff
@@ -241,7 +241,9 @@ task copyZetoZKPFiles(type: Copy) {
     // These can be very large binaries, so adding to this list in the base image should be done with care.
     // Likely if the range of circuits grows, a separate docker tag will be required for builds
     // of Paladin that have a wider range of pre-built circuits included.
-    from fileTree('domains/zeto/zkp') {
+    // Resolve version lazily: root script configures before :domains:zeto ext exists.
+    from {
+        fileTree("domains/zeto/zkp/${project(':domains:zeto').ext.zetoVersion}") {
```
annamcallister, line 246: I think I need to stop here and get a clearer understanding of what multi version support means.

I had expected that zeto would go in a direction like noto where a single domain plugin can concurrently support multiple token versions (even if the factory implementation can only deploy one version). But it looks here like we only copy in one version of the circuits, and then only build one prover; so even though the domain implementation has toggles on version, only one version can actually be used at a time.

As the comment says, these are very large binaries, so I understand we can't just forever add to this list, but should it at least be a configurable option to include support for multiple versions at a time?

annamcallister (commented):
I have questions on what is meant by multi version support and what exactly we're aiming for here which I think I need help with before I can finish the review.

build.gradle
annamcallister: A bit of back tracking here as I've gone and done more digging into what changed between zeto v0.2.2/v0.5.1

Apologies @jimthematrix while I get up to speed here. I'm certainly learning a lot as I go though!

My understanding now is that most of the circuits are backwards/forwards compatible with the verifiers deployed on chain for each version. The exceptions are:
* you can't use the v0.2.2 style locks with v0.5.1 circuits (but you can do ILockable at v0.5.1 with v0.2.2 circuits)
* the KYC circuits are incompatible in each direction

So I think I see now why v0.2.2 would be the default choice for the circuits when doing multiversion. I think I'd still expect to see the operator deploy the newest factory by default though, and use v0.5.5/V1 for e2e tests and examples.

It looks like there isn't multiversion support for KYC tokens still?

solidity/contracts/shared/Atom.sol
jimthematrix: This is actually NOT making the Atom non-upgradeable, because it has never been upgradeable, by design (the contract can only be trusted if there's no way to change its behavior b/w the time the parties committing to the traded values and the time it's executed).

It's simply reacting to the fact that the re-entrancy guard capabilities has been moved to the non-upgradeble base class by OpenZeppelin.

Nonetheless, good suggestion to add inline docs for clarity.

solidity/contracts/domains/zeto/ZetoFactoryV1.sol
jimthematrix: I actually deliberately deviated from the Noto scheme, and adopted the following instead:
- `ZetoFactory.sol`: left alone as "non-versioned" legacy
- `ZetoFactory_V0.sol`: adopted the upgradeable factory from the upstream dependency
- `ZetoFactory_V1.sol`: updated to the latest version of the upstream dependency

But after investigating this, I realized all 3 versions can be consolidated to a single solidity, so we now only have `ZetoFactory_V0.sol`. added the `_V0` suffix for potential future versions.

solidity/contracts/domains/interfaces/IZetoFungible_V1.sol
jimthematrix: since these interfaces are for the private transactions API, we do have some control over the evolution when the underlying Zeto solidity interfaces are updated.

I landed with these two versions, which is overlapping with the Noto versioning scheme but not completely identical:
- IZetoFungible_V0.sol
- IZetoFungible_V1.sol

I think this is the most streamlined versioning scheme. not seeing a need to have a `IZetoFungible.sol` interface for the latest version.

.github/workflows/release-solidity-contracts.yaml
jimthematrix: agreed. this was an oversight. updating to v0.5.1 for the release package.

build.gradle
jimthematrix: This is pointing to a genuine gap in the current code, where the artifacts involved in the proof generation are not properly delineated among the versions. I will need to fix this by making the artifacts (upstream zeto contracts, circuit assemblies aka the .wasm files, and the proving keys) completely isolated by version.

PR #1190, opened 2026-05-21
Add state visibility component

annamcallister:
Addresses comment on #1178 about the grapher being responsible for state visibility by adding a new component who's sole responsibility is ensuring privacy, and maintains an interface and testsuite dedicated to this.


core/go/internal/sequencer/coordinator/statevisibility/statevisibility.go
```diff
@@ -0,0 +1,164 @@
...
+//
+// It tracks which nodes are permitted to hold each private state's data, derived from the assembly
+// response DistributionList. The default posture is deny: a state whose AllowedNodes is nil or empty
+// is invisible to every node. This is the correct fail-safe for unknown distributions — no state data
+// is leaked by default.
+//
+// The store is internally thread-safe; callers do not need external synchronisation.
+//
+// Two callers interact with this package:
+//   - The assembly path (assembling.go) calls RecordAssemblyOutput — the only path that writes new
```
matthew1001: This is great but almost certainly going to be out-of-date at some point. Maybe the examples of who calls which functions needs to be termed as exactly that - examples. E.g. "The grapher uses calls such as ABC to ensure the data it exports only contains data appropriate to the receiving node"

core/go/internal/sequencer/coordinator/statevisibility/statevisibility.go
```diff
@@ -0,0 +1,164 @@
...
+//     visibility entries. It must be called after the grapher's AddMinter has succeeded, so that a
+//     failed AddMinter (duplicate state) leaves no stale visibility entry.
+//   - The grapher calls GetForNode, ImportIfAbsent, Has, and Delete for export, handover import,
+//     and lock-expiry cleanup respectively.
+package statevisibility
```
matthew1001: I really like how clear this makes it to a developer that this code closely relates to who gets what 👍

core/go/internal/sequencer/coordinator/coordinator.go
```diff
@@ -162,7 +165,8 @@ func NewCoordinator(
 		transportWriter:                    transportWriter,
 		contractAddress:                    contractAddress,
 		dependencyTracker:                  dependencyTracker,
-		grapher:                            grapher.NewGrapher(dependencyTracker, confutil.Uint64Min(configuration.BlockHeightTolerance, pldconf.SequencerMinimum.BlockHeightTolerance, *pldconf.SequencerDefaults.BlockHeightTolerance)),
+		stateVisibility:                    stateVisibility,
```
matthew1001: Purely a naming thing, but I wonder if `stateVisibilityTracker` is too long? It feels like it correlates with `dependencyTracker` and `stateVisibility<NoNounHere>` feels a little less clear what it's for

matthew1001 (approved):
LGTM - a few comments

PR #1191, opened 2026-05-21
Multiple event handler support in sequencer state machine

annamcallister:
Add support for multiple event handlers per event in the sequencer state machine. We previously had some event handlers which in order to support the different permutations of values within an event type had many actions and transitions defined in a single handler, all carrying a different event validators.

By making this change it no longer becomes necessary to be able to define validators on individual actions and transitions. Validators only exist at the top level of an event handler and determine whether the handler should be executed, based on the current properties of the state machine. Actions and transitions may only use guards, which are a pure boolean function on the state machine.

Event handlers must now be provided in a slice. There are two possible matching semantics:
* match first (default): evaluate the validators in order. Execute the first event handler that returns `true` (where a nil validator means `true`)
* match all: evaluate the validators in order, executing each event handler that returns `true`. Stop when we perform a transition, or when there are no more handlers.

This PR contains a small number of semantic changes to the state machine, where the refactor demonstrated gaps or bugs.


core/go/internal/sequencer/coordinator/snapshot.go
```diff
@@ -20,24 +20,25 @@ import (
...
-	"github.com/LFDT-Paladin/paladin/sdk/go/pkg/pldtypes"
 	"github.com/LFDT-Paladin/paladin/toolkit/pkg/prototk"
 )
 
 func action_SendHeartbeat(ctx context.Context, c *coordinator, _ common.Event) error {
-	return c.sendHeartbeat(ctx, c.contractAddress)
+	return c.sendHeartbeat(ctx, false)
+}
+
+func action_SendHeartbeatWithLocks(ctx context.Context, c *coordinator, _ common.Event) error {
```
matthew1001, line 30: Was this from a different PR? Looks correct, just think I reviewed this elsewhere so just double checking

core/go/internal/sequencer/statemachine/statemachine.go
```diff
@@ -103,28 +103,36 @@ type Action[E any] func(ctx context.Context, entity E, event common.Event) error
 // or if an action should be executed.
 type Guard[E any] func(ctx context.Context, entity E) bool
 
+// MatchMode controls how handlers in an EventHandlers set are evaluated.
+type MatchMode int
+
+const (
+	// MatchFirst (zero value, default): the first handler whose Validator passes is
```
matthew1001, line 110: The question of the default behaviour is interesting. I almost wonder if there shouldn't be a default. Looking at the new places we use the `EventHandlers` where currently there is only a single handler, adding a second one later might assume a given behaviour based on the contributor's assumptions. It does make the state machine more verbose to require it on every code location, so I'm not 100% sure what's correct. I suppose anyone changing a state machine is going to think carefully about what they're adding anyway, so perhaps we just leave it as you've got it for. But thought it worth raising to ponder

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -127,56 +128,55 @@ var stateDefinitionsMap = StateDefinitions{
...
 		},
 	},
 	State_Pooled: {
 		OnTransitionTo: []ActionRule{
 			{Action: action_InitializeForNewAssembly},
 		},
-		Events: map[EventType]EventHandler{
-			Event_Selected: {
+		Events: map[EventType]EventHandlers{
+			Event_Selected: {Handlers: []EventHandler{{
```
matthew1001, line 179: It's really picky but I find the `Handlers:...` always being defined on the `Event_*` line makes the indentation a little harder to follow.

matthew1001 (approved):
Approved. I think it's a great way to make the state machine more capable. Couple of comments on minor things.

core/go/internal/sequencer/coordinator/snapshot.go
annamcallister: Yes - a later PR was merged into this branch

core/go/internal/sequencer/statemachine/statemachine.go
annamcallister: Will address in a follow on PR to reduce merge conflicts with later work

core/go/internal/sequencer/coordinator/transaction/state_machine.go
annamcallister: Will address in a follow on PR to reduce merge conflicts with later work

PR #1194, opened 2026-05-21
E2e test reliability

annamcallister:
https://github.com/LFDT-Paladin/paladin/blob/fc36ed29c527eebbf91b51f0055773a86b31a71e/operator/test/e2e/e2e_pente_parallel_test.go#L294

This test has been failing intermittently for some time. This happens when the three nodes cannot agree on who is active coordinator in the test time frame. The distributed sequencer protocol tolerates nodes disagreeing for a period of time, with retries and base ledger double spend protection giving time for nodes to become eventually consistent.

It is problematic for this test though because most resolution is done across heartbeat intervals, which are by default 10s, but this test runs against a chain with a 100ms block period, which means that when nodes do not agree at the start of the test, they can quickly get through all the retries for their transactions before they've ever heard a heartbeat from another node. This is not a realistic real world scenario.

This PR aims to speed up the time to resolution (which would be beneficial in more real world set ups too):
*  the coordinator always sends a heartbeat when transitioning to State_Active (this must have been a regression when we moved to always having the heartbeat interval loop running rather than only starting it when active)
* a sequencer is always loaded to handle a heartbeat.
  * In `COORDINATOR_ENDORSER` mode all the sequencers should be loaded anyway when the domain instance is active. If one isn't loaded (e.g. after node restart), it will reach agreement quicker on who is the existing coordinator if it is loaded for a heartbeat and not for an endorsement request.
  * In `COORDINATOR_STATIC` mode  originators are no longer sent heartbeats indefinitely so should only be receiving heartbeats if they contain details about their transactions, in which case the sequencer should already be loaded.
  
  In switching heartbeat handling to always load a sequencer, I've addressed the TODO about many other handlers unnecessarily loading sequencers.


core/go/internal/sequencer/transport_client.go
```diff
@@ -128,9 +128,10 @@ func (sMgr *sequencerManager) handleAssembleRequest(ctx context.Context, message
 		return
 	}
 
-	seq, err := sMgr.LoadSequencer(ctx, sMgr.components.Persistence().NOTX(), *contractAddress, nil, nil)
-	if seq == nil || err != nil {
-		log.L(ctx).Errorf("failed to obtain sequencer to pass assemble request event to %v:", err)
+	// Get rather than load the sequencer- it must already have the transaction in memory to process the assemble request
+	seq := sMgr.GetSequencer(ctx, *contractAddress)
+	if seq == nil {
+		log.L(ctx).Debugf("ignoring assemble request for contract %s as sequencer is not loaded", contractAddress)
```
matthew1001: It would be helpful for PD to explain why that's fine. If it's already loaded we won't his this statement, so if we do should we mention in the debug log when sequencer loading will take place if it's not here - or what the expected sequence of events is that will cause us to come back to this point for successful assembly at a future time?

I think this applies to any of these failed-to-get-seqeuencer debug logs because we know the state machine & coordination logic is now pretty complex, so aiding PD as much as possible seems important.

matthew1001 (approved):
Approved but with a strong request about debug logs re. failure to get a sequencer we expect to be able to get

core/go/internal/sequencer/transport_client.go
annamcallister: I've switched to a warning log since it isn't really fine. There is no point in proceeding with a newly loaded sequencer as the event will not be handled, but there should, in normal functioning always be a loaded sequencer for these messages since they are generally part of a back and forth that started with delegation (which does load a sequencer). If a sequencer is not loaded it is either because it has been paged out after reaching maximum limits, or because a node has restarted.

PR #1196, opened 2026-05-21
Different coordinator heartbeat actions

annamcallister:
Addresses review comment on #1178 about having two different heartbeat actions depending on whether locks are being included or not.


matthew1001 (approved):
LGTM. I was pondering an alternative to `action_SendHeartbeat` so it's clear in the state machine it's the one without locks, but couldn't come up with something I liked. `action_SendBaseHeartbeat`, `action_SendHeartbeatNoLocks` - nothing quite sounded right.

PR #1198, opened 2026-05-21
Fix transport stop deadlock

annamcallister:
Fixes two deadlocks in stopping the transport plugin which seem now to be the most common cause of coordination test failures.

- **Transport manager (`manager.go`):** `Stop()` previously called `tm.cancelCtx()` at the end, so any concurrent `connectPeer` call blocked inside `startSender` — waiting on `GetNodeTransports` or `ActivatePeer` while holding the `peersLock` write-lock — would prevent `Stop()` from ever acquiring the read-lock in `listActivePeers`, deadlocking permanently. Moving `tm.cancelCtx()` to the top of `Stop()` cancels `p.ctx` (which derives from `bgCtx`) immediately, unblocking those calls and releasing the lock before it is needed.

- **gRPC transport (`grpc_transport.go`):** `shutdownTransport()` previously called `GracefulStop()` while `bgCtx` was still live, leaving any in-progress TLS handshake goroutine stuck in `VerifyConnection → getTransportDetails → RequestFromPlugin → inflight.Wait()` indefinitely — the inflight request's context derived from the never-cancelled `context.Background()`, and `pr.inflight.Close()` (the call that would have unblocked it) was unreachable because it is deferred until after `closePlugin()` returns. Adding `bgCtxCancel` to the struct and calling it at the top of `shutdownTransport()` propagates cancellation into the inflight request's context, causing `inflight.Wait()` to return immediately, the handshake to fail, and `GracefulStop()` to drain cleanly.

### DIFF
--- a/core/go/internal/transportmgr/manager.go
+++ b/core/go/internal/transportmgr/manager.go
@@ -154,6 +154,10 @@ func (tm *transportManager) Start() error {
 }
 
 func (tm *transportManager) Stop() {
+	// Cancel bgCtx first so any goroutines blocked on p.ctx (derived from bgCtx),
+	// such as startSender holding peersLock while waiting on GetNodeTransports or
+	// ActivatePeer, are unblocked before we attempt to acquire peersLock below.
+	tm.cancelCtx()
 
 	peers := tm.listActivePeers()
 	for _, p := range peers {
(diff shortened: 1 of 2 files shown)


matthew1001 (approved):
LGTM

PR #1202, opened 2026-05-27
Dedupe at least once delivery events in component test

annamcallister:
The test was assuming exactly once delivery but block chain event listeners have at least once delivery. The test needs to be able to recognise it's received the same event twice.


core/go/noderuntests/componenttest/component_test.go
```diff
@@ -308,7 +309,22 @@ func subscribeAndSendDataToChannel(ctx context.Context, t *testing.T, wsClient p
...
+							(e.BlockNumber == lastBlock && e.TransactionIndex == lastTxIndex && e.LogIndex > lastLogIndex):
+							// New event - forward it
+							lastBlock, lastTxIndex, lastLogIndex = e.BlockNumber, e.TransactionIndex, e.LogIndex
+							eventData = append(eventData, e.Data.String())
+						case e.BlockNumber == lastBlock && e.TransactionIndex == lastTxIndex && e.LogIndex == lastLogIndex:
+							// Exact duplicate - at-least-once re-delivery, discard silently
+							t.Logf("Discarding duplicate event on %s from %d/%d/%d", listenerName, e.BlockNumber, e.TransactionIndex, e.LogIndex)
+						default:
+							// Event position went backwards - this is a serious bug
+							t.Errorf("Event on %s went backwards: received %d/%d/%d after last seen %d/%d/%d",
```
matthew1001, line 326: Do we want to make this something like `require.FailNowf` instead of just logging the error?

annamcallister: We can't fail the test cleanly from a separate goroutine. I have added a return so the listener immediately stops  and doesn't process any more events if this happens. This should cause the main test to fail, with the explanatory error log the last thing printed out.

matthew1001: Sounds good, thanks

matthew1001 (approved):

PR #1204, opened 2026-05-27
[noto] Add manifests to noto unlocks

shorsher:
Building on top of @annamcallister's work allowing states to be associated with multiple transactions, noto now generates a manifest for unlocks. 

This PR only adds the unlock manifest, but does not address the issue that spend/cancel currently encode the same data. That will be addressed in a separate, follow up PR. 

example state receipt for a `createMintLock`:
```json
{
  "confirmed": [
    {
      "id": "0x1e48c16971d75dd8bed58aadf3742de615e72c4513c1dff449d421a9a5ec5db4",
      "created": "2026-05-28T19:01:26.56719768Z",
      "domain": "noto",
      "schema": "0x6e50b1cc42e93eeef8fe46ed41955e4dcab45278d94174a93a822ce7694d5f0e",
      "contractAddress": "0x52bdc3b892e74269656920c99bd3d07410d0f64c",
      "data": {
        "cancelData": "0x00020000000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000021c52b6d972ae22846bc1ee27a13d22f01329286b1b7a1ae7fbbef2b5c731d6dc8913210d4c3ccfe7db0a6f731fec06810e593c51c0ebea717219a49fc0f69f51",
        "cancelOutputs": [],
        "lockId": "0xed1b45881387132b36b96f91198c40bf94a10058d02cd69d6b10d5a262eaec9f",
        "owner": "0x4b4db1960fe373873072dc8030af15cc5be13f15",
        "replaces": "0x0000000000000000000000000000000000000000000000000000000000000000",
        "salt": "0x49b5775a00f70050d05c20f4e6d23d63a16c1930a54d9f248f81b7164dc32dc5",
        "spendData": "0x0002000000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000002792271448fa3e470bac1b39294eaa53ca6224cf3c1eac3a704986be2a1050fd68913210d4c3ccfe7db0a6f731fec06810e593c51c0ebea717219a49fc0f69f51",
        "spendOutputs": [
          "0x144f2cd213c2fa132fe8b14c8b80b3ba091f9fedff34dc738d3d188b4be0ebb5"
        ],
        "spendTxId": "0x0e886e9e8cd64f02b99c3f40b2632c8700000000000000000000000000000000",
        "spender": "0x4b4db1960fe373873072dc8030af15cc5be13f15"
      }
    }
  ],
  "info": [
    {
      "id": "0xedbf3bc11d279ec255e3e3ed37e671215409afcd7696921a203b11fd4e5bd6f2",
      "created": "2026-05-28T19:01:26.567354721Z",
      "domain": "noto",
      "schema": "0x6f0923c2fe39e0d9e0486cbf5483df4c8b9f724a4969e9a8c23d76cb5782729e",
      "contractAddress": "0x52bdc3b892e74269656920c99bd3d07410d0f64c",
      "data": {
        "salt": "0x11309395b92144dbcd291c6282820a16a05e6904bf4e700687c28ee03ac6867d",
        "states": [
          {
            "participants": [
              "0x4b4db1960fe373873072dc8030af15cc5be13f15",
              "0xfe2b177716513d9ec5582b2bdc93ea09b8d75286"
            ],
            "state": "0x792271448fa3e470bac1b39294eaa53ca6224cf3c1eac3a704986be2a1050fd6"
          },
          {
            "participants": [
              "0x4b4db1960fe373873072dc8030af15cc5be13f15",
              "0xfe2b177716513d9ec5582b2bdc93ea09b8d75286"
            ],
            "state": "0x1c52b6d972ae22846bc1ee27a13d22f01329286b1b7a1ae7fbbef2b5c731d6dc"
          },
          {
            "participants": [
              "0x4b4db1960fe373873072dc8030af15cc5be13f15",
              "0xfe2b177716513d9ec5582b2bdc93ea09b8d75286"
            ],
            "state": "0x8913210d4c3ccfe7db0a6f731fec06810e593c51c0ebea717219a49fc0f69f51"
          },
          {
            "participants": [
              "0x4b4db1960fe373873072dc8030af15cc5be13f15",
              "0xfe2b177716513d9ec5582b2bdc93ea09b8d75286"
            ],
            "state": "0x5950aac2a0db9a1af5168cb302e34f995dd663b50810f6579e26875a3ebf689e"
          },
          {
            "participants": [
              "0x4b4db1960fe373873072dc8030af15cc5be13f15",
              "0xfe2b177716513d9ec5582b2bdc93ea09b8d75286"
            ],
            "state": "0x144f2cd213c2fa132fe8b14c8b80b3ba091f9fedff34dc738d3d188b4be0ebb5"
          },
          {
            "participants": [
              "0x4b4db1960fe373873072dc8030af15cc5be13f15"
            ],
            "state": "0x1e48c16971d75dd8bed58aadf3742de615e72c4513c1dff449d421a9a5ec5db4"
          }
        ]
      }
    },
    {
      "id": "0x792271448fa3e470bac1b39294eaa53ca6224cf3c1eac3a704986be2a1050fd6",
      "created": "2026-05-28T19:01:26.567380763Z",
      "domain": "noto",
      "schema": "0x6f0923c2fe39e0d9e0486cbf5483df4c8b9f724a4969e9a8c23d76cb5782729e",
      "contractAddress": "0x52bdc3b892e74269656920c99bd3d07410d0f64c",
      "data": {
        "salt": "0x8d0116112bb39522022f3c71e505d50d06595db9a0272dbadd8b63555e8b45c2",
        "states": [
          {
            "participants": [
              "0x4b4db1960fe373873072dc8030af15cc5be13f15",
              "0xfe2b177716513d9ec5582b2bdc93ea09b8d75286"
            ],
            "state": "0x8913210d4c3ccfe7db0a6f731fec06810e593c51c0ebea717219a49fc0f69f51"
          },
          {
            "participants": [
              "0x4b4db1960fe373873072dc8030af15cc5be13f15",
              "0xfe2b177716513d9ec5582b2bdc93ea09b8d75286"
            ],
            "state": "0x144f2cd213c2fa132fe8b14c8b80b3ba091f9fedff34dc738d3d188b4be0ebb5"
          }
        ]
      }
    },
    {
      "id": "0x1c52b6d972ae22846bc1ee27a13d22f01329286b1b7a1ae7fbbef2b5c731d6dc",
      "created": "2026-05-28T19:01:26.567398971Z",
      "domain": "noto",
      "schema": "0x6f0923c2fe39e0d9e0486cbf5483df4c8b9f724a4969e9a8c23d76cb5782729e",
      "contractAddress": "0x52bdc3b892e74269656920c99bd3d07410d0f64c",
      "data": {
        "salt": "0x6362a154e20a8e096831ade02a6efc5de3409295faac1145ffc69d1b2b68c55d",
        "states": [
          {
            "participants": [
              "0x4b4db1960fe373873072dc8030af15cc5be13f15",
              "0xfe2b177716513d9ec5582b2bdc93ea09b8d75286"
            ],
            "state": "0x8913210d4c3ccfe7db0a6f731fec06810e593c51c0ebea717219a49fc0f69f51"
          }
        ]
      }
    },
    {
      "id": "0x8913210d4c3ccfe7db0a6f731fec06810e593c51c0ebea717219a49fc0f69f51",
      "created": "2026-05-28T19:01:26.567420055Z",
      "domain": "noto",
      "schema": "0x6aff8162370b8f8834bfcc9ace55e31130783fcebe647299b785ebd7cc19a258",
      "contractAddress": "0x52bdc3b892e74269656920c99bd3d07410d0f64c",
      "data": {
        "data": "0x",
        "from": "0x4b4db1960fe373873072dc8030af15cc5be13f15",
        "salt": "0xbe018ccc897838d43365e648d6dd784d921c3fa191c0638b5b0eabfe6d74355a",
        "variant": "2"
      }
    },
    {
      "id": "0x5950aac2a0db9a1af5168cb302e34f995dd663b50810f6579e26875a3ebf689e",
      "created": "2026-05-28T19:01:26.567434263Z",
      "domain": "noto",
      "schema": "0x6aff8162370b8f8834bfcc9ace55e31130783fcebe647299b785ebd7cc19a258",
      "contractAddress": "0x52bdc3b892e74269656920c99bd3d07410d0f64c",
      "data": {
        "data": "0x",
        "from": "0x4b4db1960fe373873072dc8030af15cc5be13f15",
        "salt": "0xfa71e9c34c2dde78bb5754bf64001c88ae7366a0e62e08bebd67dee615d8ac54",
        "variant": "2"
      }
    },
    {
      "id": "0x144f2cd213c2fa132fe8b14c8b80b3ba091f9fedff34dc738d3d188b4be0ebb5",
      "created": "2026-05-28T19:01:26.567448555Z",
      "domain": "noto",
      "schema": "0x39aa9c8bb0d781b43c3da1beff4ab34043e45e007316ce77412c243868815a02",
      "contractAddress": "0x52bdc3b892e74269656920c99bd3d07410d0f64c",
      "data": {
        "amount": "500",
        "owner": "0xfe2b177716513d9ec5582b2bdc93ea09b8d75286",
        "salt": "0xe22758f6bf251f2bc19d825bc419e0bd805b67e49dca14b6be84665dae997f79"
      }
    }
  ]
}
```


sdk/typescript/src/domains/noto.ts
```diff
@@ -190,6 +190,13 @@ export interface NotoUnlockPublicParams {
 
 export interface SpendLockPublicParams {
   lockId: string;
+  spendInputs: string;
```
awrichar: I think the latest ABI from #1148 uses "spendArgs" and "cancelArgs", if we want to go ahead and align the SDK for consistency.

domains/noto/internal/noto/handler_unlock_common.go
```diff
@@ -99,6 +106,17 @@ func (h *unlockCommon) buildUnlockData(ctx context.Context, notaryID, senderID,
 		// We need to know the IDs of the states at this point
 		err = h.noto.allocateStateIDs(ctx, stateQueryContext, infoStates)
 	}
+	var manifestState *prototk.NewState
+	if tx.DomainConfig.IsV1() {
```
awrichar: After #1148 this may need to be "V1 or V2"

domains/noto/internal/noto/handler_unlock_common.go
```diff
@@ -108,6 +126,18 @@ func (h *unlockCommon) buildUnlockData(ctx context.Context, notaryID, senderID,
...
+
+		if tx.DomainConfig.IsV1() {
+			// add the manifest state first
+			endorsableInfoStates = append([]*prototk.EndorsableState{{
+				Id:            *manifestState.Id,
+				SchemaId:      manifestState.SchemaId,
+				StateDataJson: manifestState.StateDataJson,
+			}}, endorsableInfoStates...)
+
+			// also append to info states being returned
```
awrichar: append -> prepend?

awrichar: We could probably do this _before_ building `endorsableInfoStates` (line 121) and it would be slightly more efficient (because we wouldn't have to separate edit that array; it would get built correctly once based on the final form of `infoStates`)

domains/noto/internal/noto/handler_unlock_common.go
```diff
@@ -30,6 +30,13 @@ type unlockCommon struct {
 	noto *Noto
 }
 
+type unlockDataResult struct {
```
awrichar: Is this used?

domains/noto/internal/noto/handler_unlock_common.go
```diff
@@ -99,6 +106,17 @@ func (h *unlockCommon) buildUnlockData(ctx context.Context, notaryID, senderID,
 		// We need to know the IDs of the states at this point
 		err = h.noto.allocateStateIDs(ctx, stateQueryContext, infoStates)
 	}
+	var manifestState *prototk.NewState
+	if tx.DomainConfig.IsV1() {
+		manifestState, err = h.noto.newManifestBuilder().addInfoStates(infoDistribution, infoStates...).buildManifest(ctx, stateQueryContext)
```
awrichar: Don't we need separate manifests for "spend" and "cancel", and don't they need to include the states that will be spent/confirmed (not just info states)?

awrichar: We're getting deep here, but I think what we need to do is:

1. build common info states (i.e. lock info and transaction info)
2. prepare spend transaction (input and output coins, and lock transition)
3. prepare spend manifest (includes spend/confirm states from 2, info states from 1)
4. prepare cancel transaction (input and output coins, and lock transition)
5. prepare cancel manifest (includes spend/confirm states from 4, info states from 1)
6. prepare manifest for the current transaction - differs depending on what transaction we're in:
  - if it's a lock creation, spend states from 2/4 are actually newly confirmed outputs here (we're creating the locked states to be spent later), and it will be spending some unlocked coins to create the lock
  - if it's a prepare on an existing lock, it doesn't spend or confirm any coins
  - it will always include a lock transition (the spend/confirm of the lock state itself, separate from the coins)
  - everything else is included as info states:
    - common info states from 1
    - prepared "confirm" states from 2 and 4 are info states (they aren't yet ready to confirm, they're just created and shared)
    - spend/cancel manifests from 3 and 5 are also info states

I think most of this logic won't be able to go directly in this common method, because it varies a little by transaction.

domains/noto/internal/noto/handler_unlock_common.go
shorsher: This was the start of me separating out spend/cancel encodings, but decided to do that in a follow up. I didn’t mean to leave in, I can remove.

domains/noto/internal/noto/handler_unlock_common.go
shorsher: perhaps we can do !V0?

shorsher:
Thanks for the review, I went ahead and put spend/cancel manifests in this PR.

domains/noto/internal/noto/handler_unlock_common.go
```diff
@@ -91,29 +99,88 @@ func (h *unlockCommon) init(ctx context.Context, tx *types.ParsedTransaction, pa
...
-	infoDistribution, err = h.getAllRecipientsDistribution(ctx, tx, notaryID, senderID, fromID, recipients, resolvedVerifiers)
-	if err == nil {
-		infoStates, err = h.noto.prepareDataInfo(ctx, unlockData, tx.DomainConfig.Variant, infoDistribution.identities(), tx.Transaction, resolvedVerifiers)
-	}
-	if err == nil {
-		// We need to know the IDs of the states at this point
-		err = h.noto.allocateStateIDs(ctx, stateQueryContext, infoStates)
+// buildUnlockOperationData builds a manifest for one operation (spend or cancel) and encodes
+// the transaction data referencing the operation manifest + info states.
+// v1 Paladin always builds a manifest
```
awrichar, line 71:
```suggestion
// buildUnlockOperationData encodes the bytes to be passed as "data" in a later operation
// (spend or cancel). This includes references to all info states (including manifest generation
// for Noto V1+).
```

domains/noto/internal/noto/handler_create_burn_lock.go
```diff
@@ -128,51 +128,50 @@ func (h *createBurnLockHandler) Assemble(ctx context.Context, tx *types.ParsedTr
...
 
+	// Build the cancel outputs before unlock data so they can be referenced in the cancel manifest
+	cancelOutputs, err := h.noto.prepareOutputs(fromID, (*pldtypes.HexUint256)(params.Amount), identityList{notaryID, senderID, fromID})
+	if err != nil {
+		return nil, err
+	}
+
 	// Build and encode the unlock data (separate to the data for this TX)
-	encodedUnlockData, infoStates, infoDistribution, err := h.buildUnlockData(ctx, notaryID, senderID, nil, tx, nil, req.ResolvedVerifiers, req.StateQueryContext, params.UnlockData)
+	unlockResult, err := h.buildUnlockResult(ctx, notaryID, senderID, nil, tx, nil /* no recipients */, req.ResolvedVerifiers, req.StateQueryContext, params.UnlockData, nil /* no spend outputs for burn */, cancelOutputs)
```
awrichar: Shouldn't we pass `fromID` here?

domains/noto/internal/noto/handler_unlock_common.go
```diff
@@ -91,29 +99,88 @@ func (h *unlockCommon) init(ctx context.Context, tx *types.ParsedTransaction, pa
...
+	if err != nil {
+		return nil, err
+	}
+
+	cancelEncoded, cancelManifest, err := h.buildUnlockOperationData(ctx, tx, stateQueryContext, cancelOutputs, infoDistribution, infoStates)
+	if err != nil {
+		return nil, err
+	}
+
+	if tx.DomainConfig.IsV1() || tx.DomainConfig.IsV2() {
```
awrichar: `!tx.DomainConfig.IsV0()` might be better, to be consistent with elsewhere in this file

domains/noto/internal/noto/handler_unlock_common.go
```diff
@@ -31,6 +32,13 @@ type unlockCommon struct {
 	noto *Noto
 }
 
+type unlockDataResult struct {
+	spendEncoded     []byte
+	cancelEncoded    []byte
```
awrichar: Maybe we should call these `spendData` and `cancelData` (ultimately that's what they're named when the caller uses them)

domains/noto/internal/noto/handler_unlock_common.go
```diff
@@ -91,29 +99,88 @@ func (h *unlockCommon) init(ctx context.Context, tx *types.ParsedTransaction, pa
...
+			StateDataJson: s.StateDataJson,
 		}
-		encodedUnlockData, err = h.noto.encodeTransactionData(ctx, tx.DomainConfig, tx.Transaction, endorsableInfoStates)
 	}
+	encodedData, err = h.noto.encodeTransactionData(ctx, tx.DomainConfig, tx.Transaction, endorsableInfoStates)
 	return
 }
 
+// buildUnlockResult builds the encoded data for spend and cancel operations
+func (h *unlockCommon) buildUnlockResult(ctx context.Context, notaryID, senderID, fromID *identityPair, tx *types.ParsedTransaction, recipients []*types.UnlockRecipient, resolvedVerifiers []*prototk.ResolvedVerifier, stateQueryContext string, unlockData []byte, spendOutputs *preparedOutputs, cancelOutputs *preparedOutputs) (*unlockDataResult, error) {
```
awrichar: `buildUnlockResult` feels a bit wrong... this is really about parsing the info states (including manifests). Maybe `buildUnlockInfo`?

awrichar: nit: This is also a whole lot of parameters, and they're interspersed a bit randomly. I'd say these are related to the overall transaction state:
`ctx, tx, resolvedVerifiers, stateQueryContext`

while these are specifically related to the unlock info parsing:
`notaryID, senderID, fromID, recipients, unlockData, spendOutputs, cancelOutputs`

Could consider wrapping the latter in an input struct, or at least reordering the params and breaking up the line.

domains/noto/internal/noto/handler_unlock_common.go
```diff
@@ -91,29 +99,88 @@ func (h *unlockCommon) init(ctx context.Context, tx *types.ParsedTransaction, pa
...
+
+	return &unlockDataResult{
+		spendEncoded:     spendEncoded,
+		cancelEncoded:    cancelEncoded,
+		infoStates:       infoStates,
+		infoDistribution: infoDistribution,
+	}, nil
+}
+
 func (h *unlockCommon) assembleStates(ctx context.Context, tx *types.ParsedTransaction, spendTxId *pldtypes.Bytes32, params *types.UnlockParams, req *prototk.AssembleTransactionRequest, unlockData []byte) (*prototk.AssembleTransactionResponse, *manifestBuilder, *unlockStates, error) {
```
awrichar: This helper is used from `unlock` and `prepare_unlock`, but not `prepare_mint_unlock` or `prepare_burn_unlock` or any of the `create_lock` methods. Might be worth noting that in a comment (I was confused until I looked back at the code).

awrichar: I guess it's also worth re-evaluating if the method is useful. It's commonizing _some_ logic for these two calls, but ultimately returns a complex struct that has to be decomposed differently by both, and there are a bunch of other methods that assemble unlock states in a completely different manner.

Seems like it should either be made common to more paths, or should possibly just be inlined back to the two places that are using it...

domains/noto/internal/noto/handler_unlock_common.go
```diff
@@ -177,13 +244,31 @@ func (h *unlockCommon) assembleStates(ctx context.Context, tx *types.ParsedTrans
...
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	spendEncoded, spendManifest, err := h.buildUnlockOperationData(ctx, tx, req.StateQueryContext, unlockedOutputs, infoDistribution, infoStates)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	if tx.DomainConfig.IsV1() || tx.DomainConfig.IsV2() {
```
awrichar: Feels like this should be a `!IsV0()`, combined with an `else` on line 272.

domains/noto/internal/noto/handler_create_burn_lock.go
shorsher: We didn't originally, so I was just keeping that behavior. Let me take a look

domains/noto/internal/noto/handler_unlock_common.go
shorsher: It seemed to me that `assembleStates` was the original implementation before we added the the new burn/mint locks, which then introduced `buildUnlockData`. I can try to break out  some common functionality if you think that's required for this PR.

domains/noto/internal/noto/handler_unlock_common.go
shorsher: Yeah we can be more consistent

domains/noto/internal/noto/handler_prepare_unlock.go
```diff
@@ -52,118 +50,142 @@ func (h *prepareUnlockHandler) Init(ctx context.Context, tx *types.ParsedTransac
...
 		return nil, err
 	}
-	fromID, err := h.noto.findEthAddressVerifier(ctx, "from", params.From, req.ResolvedVerifiers)
+
+	unlockInfo, err := h.buildUnlockInfo(ctx, tx, req.ResolvedVerifiers, req.StateQueryContext, &unlockInfoInput{
+		resolvedIdentities: ids,
+		recipients:         unlockParams.Recipients,
+		unlockData:         unlockData,
+		spendOutputs:       outputs,
+		omitCancelManifest: true,
```
awrichar: I think we do need to prepare a "cancel" here (which means some more reordering). Not sure if I broke it or if it was missed in the first pass.

domains/noto/internal/noto/handler_prepare_unlock.go
```diff
@@ -52,118 +50,143 @@ func (h *prepareUnlockHandler) Init(ctx context.Context, tx *types.ParsedTransac
...
+		}
+	}
+
+	unlockInfo, err := h.buildUnlockInfo(ctx, tx, req.ResolvedVerifiers, req.StateQueryContext, &unlockInfoInput{
+		resolvedIdentities: ids,
+		recipients:         unlockParams.Recipients,
+		unlockData:         unlockData,
+		spendOutputs:       outputs,
+		cancelOutputs:      cancelOutputs,
+		omitCancelManifest: tx.DomainConfig.IsV0(),
```
awrichar: This isn't strictly needed (`buildUnlockInfo` will already omit all manifests for V0)

domains/noto/internal/noto/handler_prepare_unlock.go
```diff
@@ -52,118 +50,143 @@ func (h *prepareUnlockHandler) Init(ctx context.Context, tx *types.ParsedTransac
...
+		lock, err := h.noto.prepareLockInfo_V0(unlockParams.LockID, fromID.address, nil, unlockInfo.infoDistribution)
+		if err != nil {
+			return nil, err
+		}
+		infoStates = append(infoStates, lock.state)
+	}
+
+	mb := h.noto.newManifestBuilder().
+		addOutputs(outputs).
+		addInfoStates(unlockInfo.infoDistribution, infoStates...)
```
awrichar: Feels like all the manifest builder code can go together (down on 155) rather than splitting it into two places

examples/spend-lock-tokens/src/create-mint-lock.ts
```diff
@@ -0,0 +1,184 @@
...
+      notaryMode: "basic",
+    })
+    .waitForDeploy(DEFAULT_POLL_TIMEOUT);
+  if (!token) {
+    logger.error("Failed to deploy Noto token!");
+    return false;
+  }
+  logger.log(`Noto token deployed at ${token.address}`);
+
+  // Step 2: Create a mint lock with a recipient
```
awrichar, line 74: Maybe it should be more obvious from the filename that this tests mint locks (as opposed to the other one that tests transfer locks)

domains/noto/internal/noto/handler_create_burn_lock.go
awrichar: Feels like an existing bug that we should tidy up.

awrichar (approved):

PR #1205, opened 2026-05-28
Async assembly and endorsement, block height tolerance rejections

annamcallister:
1. Endorsement is now done inside the coordinator rather than in the transport client.
2. Assembly and endorsement are performed in goroutines so the work does not block the state machine event loop
3. Assembly and endorsement requests have an expiry time, with the context for the assembly/endorsement goroutine cancelled at the expiry time
4. Delegation, assemble and endorsement requests are rejected if the receiver is not within block height tolerance of the sender
---
Introducing these rejections highlighted that we had a lot of inconsistency in the semantics of our state machine events and messages. I have attempted to consolidate on the following pattern.

* Response - The receiver has actioned the request and is returning with the result. The result may be successful or contain errors, but these are expected errors given the values in the request/conditions on the node (e.g. assembly/endorsement reverted, max inflight transactions reached)
* Error - The receiver has attempted to action the request but an unexpected error has occurred (e.g. failure to resolve a verifier, 
* RequestRejected - The receiver has not attempted to action the request (e.g. transaction unknown, block height tolerance exceeded)

I have folded the previous `NotActiveCoordinator` and `TransactionUnknown` messages/event types into rejection types for consistency. 

I think it should be a point of discussion for this PR though as to whether this is the correct approach.

---
This PR also addresses some debt
* Address review comments from #1191 for indentation and always specifying which matching semantics to use
* Clean up unused proto definitions
* Address TODO about using toolkit proto definitions instead of pb.Any with custom parsing. This has resulted in a slight change of behaviour where the absence of a repeating field now produces nil rather than an empty slice. We previously had a check that all of these repeating proto fields were not nil when marshalled into the EndorsementRequest go struct. However, since the parsing meant that they could never be nil, these checks were effectively a no op- so removing them to allow nil fields now does not result in looser behaviour. When the request is sent to the domain over the gRPC interface, an empty slice and nil are treated identically. Only the simple test domain has had to be changing to handle nil rather than empty slices.


core/go/internal/sequencer/common/rejection_reasons.go
```diff
@@ -0,0 +1,28 @@
...
+// RejectionReason is the reason a sequencer message was rejected. The same set of reasons
+// applies to all rejection message types (endorsement, delegation, assemble, pre-dispatch).
+// The integer values are carried over the wire as int32 in the proto rejection_reason field.
+type RejectionReason int
+
+const (
+	RejectionReason_BlockHeightTolerance        RejectionReason = iota // 0 — sender and receiver block heights differ by more than the configured tolerance
+	RejectionReason_NotCurrentDelegate                                 // 1 — receiver does not recognise the sender as its current active coordinator/delegate
+	RejectionReason_TransactionUnknown                                 // 2 — receiver does not recognise the transaction (already cleaned up)
+	RejectionReason_EndorserIsActiveCoordinator                        // 3 — endorser is currently the active coordinator; the sender should re-route to a different endorser
```
matthew1001: Just wanting to check why the active coordinator can't also endorse - is that requirement always true?

core/go/internal/sequencer/coordinator/transaction/assembling.go
```diff
@@ -224,3 +218,21 @@ func action_NudgeAssembleRequest(ctx context.Context, txn *coordinatorTransactio
...
 	return txn.nudgeAssembleRequest(ctx)
 }
+
+func action_HandleAssembleBlockHeightRejection(ctx context.Context, txn *coordinatorTransaction, event common.Event) error {
+	e := event.(*AssembleRequestRejectedEvent)
+	log.L(ctx).Warnf("assemble request rejected due to block height tolerance: coordinator block height=%d, assembler block height=%d, tolerance=%d", e.CoordinatorBlockHeight, e.AssemblerBlockHeight, txn.blockHeightTolerance)
+	return nil
+}
+
+func validator_IsAssembleBlockHeightRejection(_ context.Context, _ *coordinatorTransaction, event common.Event) (bool, error) {
```
matthew1001, line 224: I think `guard` vs `validator` is an area we've discussed before, but this instance feels odd again to me. I tend to think of them as:
 - `validator` -> this always applies except in some particular circumstances e.g. it has expired
 - `guard` -> a conditional

For event types I think we're really talking about the latter?

core/go/internal/sequencer/coordinator/transaction/confirming_dispatch.go
```diff
@@ -117,3 +117,11 @@ func action_SendPreDispatchRequest(ctx context.Context, txn *coordinatorTransact
 func action_NudgePreDispatchRequest(ctx context.Context, txn *coordinatorTransaction, _ common.Event) error {
 	return txn.nudgePreDispatchRequest(ctx)
 }
+
+func validator_IsPreDispatchNotCurrentDelegateRejection(_ context.Context, _ *coordinatorTransaction, event common.Event) (bool, error) {
+	return event.(*PreDispatchRequestRejectedEvent).RejectionReason == common.RejectionReason_NotCurrentDelegate, nil
```
matthew1001: Same discussion/comment here re. guard vs validator

core/go/internal/sequencer/statemachine/statemachine.go
```diff
@@ -409,6 +402,14 @@ func (sm *StateMachine[S, E]) evaluateTransitions(
 			sm.transitionCallback(ctx, entity, previousState, newState, event)
 		}
 
+		newStateDefinition, exists := sm.definitions[newState]
```
matthew1001, line 405: Just wanted to check:
1. Was this moved just to make sure the extra debug line is written before returning, and/or that the transition callback is invoked?
2. We return `true` from this function in error cases, although the function comment is "Returns (true, nil) when a transition was taken, (false, nil) when none matched." Is that correct?

core/go/internal/sequencer/transport/message_types.go
```diff
@@ -20,20 +20,23 @@ const (
 	MessageType_AssembleRequest                  = "AssembleRequest"
 	MessageType_AssembleResponse                 = "AssembleResponse"
 	MessageType_AssembleError                    = "AssembleError"
+	MessageType_AssembleRejection                = "AssembleRejection"
 	MessageType_CoordinatorHeartbeatNotification = "CoordinatorHeartbeatNotification"
 	MessageType_DelegationRequest                = "DelegationRequest"
 	MessageType_DelegationResponse               = "DelegationResponse"
```
matthew1001, line 26: I note we don't have `MessageType_DelegationError` - not needed for anything?

core/go/internal/sequencer/common/rejection_reasons.go
matthew1001: Also just a general question/discussion point on protofbuf vs go constants. We have message types defined as protofbuf messages, but the actual errors are defined here. Is there a reason we should define these as protobuf enums?

core/go/internal/sequencer/coordinator/transaction/transaction.go
```diff
@@ -69,6 +69,8 @@ type coordinatorTransaction struct {
 	revertCount                        int
 	lastCanRetryRevert                 bool
 	assembleErrorCount                 int
+	endorseToleranceByRequirement      map[string]int // attReqName → max tolerable failures; computed once on entering Endorsement_Gathering
```
matthew1001: Do we want these comments?

core/go/internal/sequencer/originator/transaction/assembling.go
```diff
@@ -63,63 +66,85 @@ func action_AssembleError(ctx context.Context, t *originatorTransaction, event c
...
-	// However, we do pass the preAssembly by pointer so there may be a need to add locking or pass by value if we off load to a separate thread
-	// lets keep it synchronous for now given that the whole contract is single threaded on the assemble stage anyway, this is unlikely to have a huge negative impact
-	// but from a flow of data perspective and the state machine logic, it _could_ be converted to async
-	postAssembly, err := txn.engineIntegration.AssembleAndSign(ctx, txn.pt.ID, txn.pt.PreAssembly, txn.latestAssembleRequest.stateLocksJSON, txn.latestAssembleRequest.coordinatorsBlockHeight)
+	assembleCtx := ctx
+	cancel := func() {}
+	if !req.expiry.IsZero() {
+		assembleCtx, cancel = context.WithDeadline(ctx, req.expiry)
+	}
+	go func() {
```
matthew1001, line 94: Just checking if you're happy that the previous comment:

```
// However, we do pass the preAssembly by pointer so there may be a need to add locking or pass by value if we off load to a separate thread
```

has been considered with this change?

core/go/internal/sequencer/sequencer.go
```diff
@@ -538,17 +540,33 @@ func (sMgr *sequencerManager) OnNewBlockHeight(ctx context.Context, blockHeight
 }
 
 func (sMgr *sequencerManager) notifySequencersNewBlock(ctx context.Context, blockHeight int64) {
```
matthew1001, line 542: I did a quick AI analysis of the contexts and locks introduced here and it suggested that the `ctx` passed in will be in the process of being torn down from the DB post commit context that invokes `OnNewBlockHeight`. Could you check if it's safer to use `sMgr.ctx`, or something else?

matthew1001 (commented):
Thanks Anna, finished a first review with some comments.

Re. the general PR description:

```
 - Response - The receiver has actioned the request and is returning with the result. The result may be successful or contain errors, but these are expected errors given the values in the request/conditions on the node (e.g. assembly/endorsement reverted, max inflight transactions reached)
 - Error - The receiver has attempted to action the request but an unexpected error has occurred (e.g. failure to resolve a verifier,
 - RequestRejected - The receiver has not attempted to action the request (e.g. transaction unknown, block height tolerance exceeded)
```

These generally feel like they make sense, but also slightly arbitrary distinctions between e.g. assembly reverted (regular response) vs verifier couldn't be resolved (error response). Is there a clear-ish set of rules we think we can describe the different scenarios with?

core/go/internal/sequencer/common/rejection_reasons.go
annamcallister: The active coordinator could endorse here. Rejecting here though forces resolution on to a single coordinator to occur much earlier, and prevents transactions from reaching the base ledger when we know there will be. state contention.

annamcallister: I think these reasons could be defined in the protos and would make more sense there. I went on a few iterations around this and yes it doesn't look like I landed in the right place.

core/go/internal/sequencer/coordinator/transaction/assembling.go
annamcallister: I don't think there's an "always applies" element in a validator. Both validators and guards are conditional. I think the distinction (since #1191) is:

A validator operates on the subject of the state machine _and_ an event to determine if an event handler should run
A guard operates only on the subject of the state machine and determines whether an action/transition should run

In this case the rejection reason is part of the event so this can only be a validator.

core/go/internal/sequencer/statemachine/statemachine.go
annamcallister: I think this should have always been the semantically correct order, but it hasn't mattered until now. I found that one of the new transitions added did care that the transition callback had been run before the on transition to actions in the new state ran.

It looks like that is the long standing behaviour (just moved to a different line)- we return true with an error because the state has actually changed. Possibly a discussion separate to this PR as to whether that is correct

core/go/internal/sequencer/transport/message_types.go
annamcallister: Yes we don't currently have an error path for delegate - just responses and rejections

core/go/internal/sequencer/originator/transaction/assembling.go
annamcallister: Yes I definitely remember considering it- I'll need to have a look to remind myself why I remember it being OK though!

core/go/internal/sequencer/coordinator/transaction/assembling.go
annamcallister: Noting our offline discussion about how we should make this distinction clearer in a future revision of state machine definitions.

core/go/internal/sequencer/originator/transaction/assembling.go
annamcallister: I believe it was alright as assembly doesn't modify PT. Turns out that wasn't true - it is now.

core/go/internal/sequencer/sequencer.go
annamcallister: I'm about to remove block height notifications in another PR, which will fix this

core/go/internal/sequencer/statemachine/statemachine.go
matthew1001: OK, thanks for checking on both counts.

core/go/pkg/proto/engine.proto
```diff
@@ -57,6 +57,13 @@ message TransactionConfirmed {
     string failure_message = 8;
 }
 
+enum RejectionReason {
+    BLOCK_HEIGHT_TOLERANCE = 0;         // sender and receiver block heights differ by more than the configured tolerance
```
matthew1001, line 61: 👍

matthew1001 (approved):
Thanks Anna, looks great. 

Just noting for posterity, re. my PR comment

```
These generally feel like they make sense, but also slightly arbitrary distinctions between e.g. assembly reverted (regular response) vs verifier couldn't be resolved (error response). Is there a clear-ish set of rules we think we can describe the different scenarios with?
```

that I think we landed on:

`Error` = unexpected (i.e. something should have caught whatever the issue was somewhere else in the transaction's lifecycle
`Response` containing errors = expected errors (i.e. assembly revert)

I think this is more or less as you described it, but wanted to clarify based on our in-person discussion.

Issue #1206, opened 2026-05-28
Pente private state completion tracking for assembly and endorsement


annamcallister:
An assemble request sent by a coordinator includes
* The coordinators confirmed block height (`C`)
* State locks and private state data for assembled transactions not included in the blockchain at block `C`
* State locks and private state data for confirmed transactions in the previous `N `blocks, where `N` is the configured block height tolerance

A node may only accept an assemble request if it's own confirmed block height is within `N` blocks of the coordinator's.

Private state distribution is independent of block height indexing. This means that it is possible for the assembling node to have indexed to within the block height tolerance of the coordinator, but to still be missing a private state, which it should have visibility of, confirmed in a block lower than `C-N`. This will likely result in an invalid assembly.

Understanding which confirmed states a node should have received private state data for is different for different domains
* Pente - all nodes in the privacy group receive all private state data. For a given transaction we can give a definitive yes or no as to whether a node has received all private state data.
* Noto
  * v1: a manifest distributed as an info state to nodes participating in the transaction details which states those nodes should receive private state data for. We can only answer definitively for transactions the node has participated in, and only once the manifest has been received. We can not distinguish between have not yet received the manifest vs will never receive the manifest.
  * v0: no manifests and no way to determine if a node has received all private states for a transaction.
* Zeto: no manifests and no way to determine if a node has received all private states for a transaction.

Paladin currently only tracks private state completeness for receipt listeners. Transactions receipts for listeners configured with blocking behaviour are not sent when both
* the node has not yet received all private state data
* the node will never receive all private state data
i.e. receipts are only sent when there is certainty that all the private state data has been received. Receipt listeners tolerate uncertainty about whether private state data are missing by simply not sending receipts. 

For assembly and endorsement there can be no uncertainty in this negative state. Assembly and endorsement can only be blocked when there is certainty that private state data is missing.

There are also significant differences with respect to cost overheads. The current receipt listener gap behaviour is to track private state completeness only for contracts which are a configured source for a receipt listener, and on the listener's dedicated goroutine. For assembly and endorsement we need to track private state completeness for every contract in the domain, and on the main blockindexer goroutine - since we must be able to answer the question of whether we have all private states relative to the confirmed block height

[...]

PR #1209, opened 2026-06-02
Track private state completeness

annamcallister:
Addresses #1206 

Domains which are able to assert the difference between "I have not yet received a state but I will" vs "I will never receive this state" (currently just Pente), are able to opt into a new persistence layer which tracks private states they have not yet received. Rows in this table are updated/deleted as new states arrive.

This persistence layer is queried at assembly and endorsement time, with assembly/endorsement blocked if the node is missing private states in the block range not covered by the in memory private state data and locks passed by the coordinator.

The linked issue expressed a desire to optimise receipt listener gaps, by making use of this new persistence layer, where available, to query state availability. I haven't upon implementation been able to see a compelling reason for doing this. 
* The states for the state receipt are already calculated as part of the receipt listener logic. The change in behaviour would be for domains which have the global completeness index to make a DB query, rather than pass those states into the domain. Where we are not building a global index for every contract this makes the receipt listener logic more complex for little gain.
* The next missing state still needs to be persisted as the receipt listener gap - a DB join with this table is what wakes the receipt listener up again when it becomes unblocked on state arrival. The new persistence layer could be used to query completeness but it can't replace the existing receipt gap rows for waking the listener.


peterbroadhurst (changes requested):
I really like the justification in the PR header, and the decision to make this a separate concern to receipt checking.

However, because this code is so performance critical on the block listener - I think we should go even further if you don't mind:

- We lean in to this only works for all-or-nothing cases. Domains (like Pente) that say they expect all states.
- With this, we don't need to make multiple gRPC calls to the domain - the answer that `SupportsCompletionIndex()` returns true is enough
    - Have a think maybe about the name of that in the context of this - maybe "require all states for assembly" is what we're really saying here, so `FullStateAvailablityRequired`
- Then I also suggest we take the effort to do a single DB call for the aggregate set of all TXs we need to confirm state availability for.
   - This could be by passing a list of  TXIDs and getting all state IDs back to compare
   - It could be by passing all state IDs we expect, and getting back a list of misses

annamcallister:
@peterbroadhurst this is ready for review again. I believe I've addressed your comments, plus following our in person discussion:
* block height for a sequencer is now taken from its domain block height
* event streams now have a function on their interface for retrieving their in memory block height
* an event stream's in memory block height is moved forwards for empty blocks

core/go/internal/domainmgr/domain.go
```diff
@@ -65,7 +65,7 @@ type domain struct {
 	initRetry   *retry.Retry
 	config      *prototk.DomainConfig
 	schemasByID map[string]components.Schema
-	eventStream *blockindexer.EventStream
+	eventStream blockindexer.EventStream
```
peterbroadhurst, line 68: See this changed to an interface 👍

domains/pente/src/main/java/io/kaleido/paladin/pente/domain/PenteDomain.java
```diff
@@ -72,10 +72,11 @@ protected CompletableFuture<ConfigureDomainResponse> configureDomain(ConfigureDo
...
          config.initFromConfig(request);
 
-         var domainConfig = DomainConfig.newBuilder()
-                 .addAllAbiStateSchemasJson(config.allPenteSchemas())
-                 .setAbiEventsJson(config.getEventsABI().toString())
-                 .build();
+        var domainConfig = DomainConfig.newBuilder()
+                .addAllAbiStateSchemasJson(config.allPenteSchemas())
+                .setAbiEventsJson(config.getEventsABI().toString())
+                .setFullStateAvailablityRequired(true)
```
peterbroadhurst, line 78: Cool, so just a simple switch

core/go/pkg/blockindexer/event_streams.go
```diff
@@ -646,7 +684,21 @@ func (es *eventStream) dispatcher() {
...
-		case d := <-es.dispatch:
+		case msg := <-es.dispatch:
+			if msg.confirmed != nil {
+				// A block (or catchup range) was confirmed to contain no matching events- advance the in-memory checkpoint.
+				// This checkpoint could also be persisted for use on restart. It's a tradeoff between writing a checkpoint
+				// per block, per event indexer in normal operation vs. needing to catch up over a bigger range of blocks
+				// on node restart. I've opted to take the hit at start up but this could be changed in the future.
+				if msg.confirmed.blockNumber > es.checkpoint.Load() {
+					es.checkpoint.Store(msg.confirmed.blockNumber)
+				}
```
peterbroadhurst: Want to make sure it's already true that we write checkpoints during catchup, at regular intervals.
I thought that was true already - but if it isn't it's a significant problem.

core/go/pkg/blockindexer/event_stream_types.go
```diff
@@ -123,6 +123,20 @@ type EventStreamStatus struct {
 	Catchup         bool
 }
 
+// EventStream is the handle returned to components that register an internal event stream.
+// It provides access to the stream's definition and its current confirmed processing position.
+type EventStream interface {
```
peterbroadhurst, line 128: Cool, so changed to an interface 👍

core/go/internal/domainmgr/event_indexer.go
```diff
@@ -309,13 +309,23 @@ func (d *domain) handleEventBatchForContract(ctx context.Context, dbTX persisten
 		stateReads[i] = &pldapi.StateReadRecord{DomainName: d.name, State: stateID, Transaction: txUUID}
 	}
 
+	// Only confirmed states are tracked for the completion index
+	confirmedStateIDsByTX := make(map[uuid.UUID]map[string]pldtypes.HexBytes)
```
peterbroadhurst, line 313: Because only those that are required are added to this map, the name is confusing.
Maybe `stateConfirmationsToRecordByTX`

peterbroadhurst: Sorry it's a bit of a pain, but I actually think the spelling problem is bigger - the `StateCompletionEntry` DB/Structs are implying this is a record of state completion - but we have a completely separate table for that which is already updated.

This is actually "state ids that must be present to confirm a transaction"

core/go/internal/sequencer/common/engine_integration.go
```diff
@@ -94,15 +89,29 @@ func (e *engineIntegration) WriteStatesForTransaction(ctx context.Context, txn *
...
+}
+
+func (e *engineIntegration) CheckStateCompletion(ctx context.Context, block int64) (bool, error) {
+	if !e.domainSmartContract.Domain().FullStateAvailablityRequired() {
+		return true, nil
+	}
+	return e.components.StateManager().CheckStateCompletionForContract(
+		ctx, e.components.Persistence().NOTX(),
+		e.domainSmartContract.Address().String(), block,
+	)
```
peterbroadhurst: To confirm I understand the question asked here is:

- There must be no (name TBD) "required state entries" unavailable with a block number <= X

peterbroadhurst (changes requested):
The code looks great @annamcallister - but I have I have opened a spelling question which I think we should close on before we merge this (as it affects the DB migrations)

core/go/pkg/blockindexer/event_streams.go
annamcallister: Yes we write checkpoints during catchup. This comment was attempting to capture something different.

Let's say we have a domain event in block 0, we then have 100 blocks with no domain events, and then the node restarts. The event stream checkpoint will be at 0, because that's the last block we saw an event in, we did not progress the persisted checkpoint for empty blocks.

When the node comes back the event stream will go into catch up mode, including over these 100 blocks that it has already seen. Assuming these are not truly empty blocks, they contain events, just not ones that match the ES definition, the node has to repeat the work to confirm that none of these events match the event stream.

My gut feel is that it is better to bear this cost at start up (which is actually the current behaviour, but adding the in memory checkpoint which does move forwards over empty blocks really highlights that this one doesn't), than to add in a new DB write on every block (or every N blocks if we opted to batch).

core/go/internal/sequencer/common/engine_integration.go
annamcallister: Yes that's the question 👍

core/go/internal/domainmgr/event_indexer.go
peterbroadhurst: Thanks for the update to `pending_private_state_data` 👍

core/go/pkg/blockindexer/event_streams.go
annamcallister: I've changed the behaviour of this. We now persist every checkpoint update. There will be no repeat work on start up of non matching blocks.

peterbroadhurst: Perfect, thanks. The potential unbounded scope of startup processing "debt" in a catch-up/re-sync scenario could be large. So I think now is the right time to close the gap on this.

peterbroadhurst (approved):
Thanks @annamcallister - really like where the spelling landed

PR #1214, opened 2026-06-04
Flex query

gabriel-indik:
Allow fields of type UUID and hex to support starts with/ends with/contains/etc

https://github.com/user-attachments/assets/4e239c4d-2f96-495d-9ba3-183b9b291c47


gabriel-indik:
Important: this PR should be considered/merged after https://github.com/LFDT-Paladin/paladin/pull/1218.

core/go/internal/filters/query_traverser_gorm.go
```diff
@@ -104,18 +104,27 @@ func (t *gormTraverser) IsEqual(e *query.OpSingleVal, fieldName string, field Fi
 	return t
 }
 
+func likeSQLColumn(field FieldResolver) string {
+	col := field.SQLColumn()
+	if _, ok := field.(UUIDField); ok {
+		return fmt.Sprintf("CAST(%s AS TEXT)", col)
```
matthew1001, line 110: I spent a bit of time understand the affect of this on index-use in the DB. I think it shouldn't affect any existing full-UUID queries which will use existing indexes. For wildcard searches it won't be able to use the index, but so using the new wildcard option to find by UUID won't be efficient - but then for many wildcard lookups (e.g. "%matchthistext") an index could never be used as far as I'm aware.

The only thing that might be worth exploring at some point is an index that would make suffix-wildcard searches most efficient (e.g. "matchthistext%"). I don't think that's critical for this PR, since the main thing is we don't want to have regressed any existing search performance.

matthew1001 (approved):
Happy to approve this having taken a look at any possible DB performance impact. There's probably an optimization we could make for suffix wildcards, but this PR doesn't regress the performance of any existing queries.

core/go/internal/filters/query_traverser_gorm.go
gabriel-indik: @matthew1001 Note that before one could only use the "equal" operator in the JSON query for UUIDs. Using "like" was not supported, the API call would fail. When using the "equal" operator one has to provide a valid UUID as the value argument, otherwise the API would reject the call while attempting to parse the UUID. With the changes in place, if we use the "like" operator, the value will be casted to string. However, if we use the "equal" operator, the code will behave like it did before the change: the user must provide a valid UUID, otherwise the call will be rejected. There would be no casting in that case. In other words, the queries that were supported before the change should not be negatively impacted performancewise.

PR #1215, opened 2026-06-05
Auto generate state machine docs

matthew1001:
## PR description

The state machines in the distributed sequencer need documenting clearly so users and contributors have a good understanding of the code architecture, but without overloading the reader with all of the detail in a single page.

Since the state machines are very liable to changes over time it is much more sustainable if we generate the documentation from the code rather than having to remember to manually update it when code changes are made.

## Generation script

This PR includes an AI-generated script that built the `.md` files also in the PR, including the mermaid diagrams and state descriptions taken from the comments along side the `State_*` constants.

Since the page is now using those comments I went through them all to make them more consistent (re. casing, sentence style etc) and (I think) more accurate. These comment changes probably need the closest review for accuracy.

The script also builds a second page which draws similar mermaid diagrams but includes **all** the events that cause the state transitions. The diagrams in this page are very busy to say the least. For now I've included the `.md` file (because I don't think it's inaccurate at least), but I've left it off the docs nav on the LHS while we decide what we really want to show the user. Perhaps for the full detail we want a page per-state-machine?

## Previews of the new page

image

---

image

## Preview of the full detail page

As you can see the full detail view is pretty hard to consume, hence not including in the nav for now:

image

---

image

## Gradle tasks

The PR adds 2 new gradle tasks for a) checking if the docs are out of date based on code changes, and b) generating updated docs using the new script:

 - `./gradlew checkStateMachineDocs`
 - `./gradlew generateStateMachineDocs`

The `checkStateMachineDocs` task is a `dependsOn` entry for `:core:go:build`

Below is the build action for this PR with new state machine changes merged in but before running the generate task:

image


annamcallister:
Could we have a new descriptions file to be used with i18n translations rather than using comments directly? 

It feels like there's a level of detail we might want to include in code comments that we don't want reflected in the docs.

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -25,22 +25,24 @@ import (
...
-	State_Confirming_Dispatchable = common.CoordinatorTransactionState_Confirming_Dispatchable // Endorsed; waiting for originator dispatch confirmation; originator can still decline
-	State_Ready_For_Dispatch      = common.CoordinatorTransactionState_Ready_For_Dispatch      // Dispatch confirmation received; point of no return; waiting to be collected by the dispatcher
-	State_Dispatched              = common.CoordinatorTransactionState_Dispatched              // Collected by the dispatcher/public TX manager; in-flight on the base ledger
-	State_Confirmed               = common.CoordinatorTransactionState_Confirmed               // Recently confirmed on the base ledger; not held in memory indefinitely
-	State_Final                   = common.CoordinatorTransactionState_Final                   // Final state; transactions are removed from memory on entry
-	State_Evicted                 = common.CoordinatorTransactionState_Evicted                 // Evicted for memory/in-flight slot management; distinct from Final
+	State_Initial                 State = iota // Transaction state machine created
```
annamcallister:
```suggestion
	State_Initial                 State = iota // Transaction state machine has been created
```

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -25,22 +25,24 @@ import (
...
+	State_Pooled                               // Waiting in the pool to be selected and sent for assembly to the transaction's originator
+	State_PreAssembly_Blocked                  // Has not been assembled yet and cannot be assembled because a dependency never got assembled successfully, typically because it was parked or reverted
```
annamcallister:
```suggestion
	State_PreAssembly_Blocked                  // The transaction cannot yet be put in the pool to be selected for assembly because a dependency must be assembled first
```

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -25,22 +25,24 @@ import (
...
+	State_Assembling                           // An assemble request has been sent to the originator and we are waiting for the response
+	State_Reverted                             // The transaction has been reverted at assembly time by the originator
```
annamcallister:
```suggestion
	State_Reverted                             // The transaction has been reverted, either at assembly time by the originator or on the base ledger
```

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -25,22 +25,24 @@ import (
...
+	State_Endorsement_Gathering                // Assembly completed successfully and we are now waiting for endorsement of the assembled transaction
```
annamcallister:
```suggestion
	State_Endorsement_Gathering                // The transaction has been successfully assembled and endorsement requests have been sent
```

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -25,22 +25,24 @@ import (
...
+	State_Blocked                              // All endorsements have been received but the transaction cannot proceed due to dependencies not being ready for dispatch
+	State_Confirming_Dispatchable              // Endorsed and waiting for confirmation that were are OK to dispatch. The originator can still request not to proceed at this point.
```
annamcallister:
```suggestion
	State_Confirming_Dispatchable              // The transaction has been endorsed. Confirmation from the originator is required before the transaction can be dispatched. The originator may still request not to proceed at this point.
```

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -25,22 +25,24 @@ import (
...
+	State_Ready_For_Dispatch                   // Dispatch confirmation has been received from the originator and we are waiting to be collected by the dispatcher thread. Going into this state is the point of no return to the base ledger.
```
annamcallister:
```suggestion
	State_Ready_For_Dispatch                   // Dispatch confirmation has been received from the originator and the transaction is waiting to be collected by the dispatch goroutine
```

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -25,22 +25,24 @@ import (
...
+	State_Dispatched                           // Collected by the dispatcher thread and submitted by the public TX manager to the base ledger
```
annamcallister:
```suggestion
	State_Dispatched                           // The transaction has been dispatched as either a public transaction to be picked up by the public TX manager, or as a chained private transaction, to be picked up by another sequencer. Entry to this state represents the point of no return to the base ledger.
```

^^ It used to be true that ready for dispatch was the absolute point of no return- but with the ability to stop the dispatch loop this is no longer true. If the dispatch loop is running then the state machine event loop cannot halt the progress of a ready to dispatch transaction through to dispatched, but it is now possible to stop the dispatch loop, and know that only transactions that are in dispatched state are past the point of no return. This is the behaviour that active -> active flush uses for managing key rotation

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -25,22 +25,24 @@ import (
...
-	State_Ready_For_Dispatch      = common.CoordinatorTransactionState_Ready_For_Dispatch      // Dispatch confirmation received; point of no return; waiting to be collected by the dispatcher
-	State_Dispatched              = common.CoordinatorTransactionState_Dispatched              // Collected by the dispatcher/public TX manager; in-flight on the base ledger
-	State_Confirmed               = common.CoordinatorTransactionState_Confirmed               // Recently confirmed on the base ledger; not held in memory indefinitely
-	State_Final                   = common.CoordinatorTransactionState_Final                   // Final state; transactions are removed from memory on entry
-	State_Evicted                 = common.CoordinatorTransactionState_Evicted                 // Evicted for memory/in-flight slot management; distinct from Final
+	State_Initial                 State = iota // Transaction state machine created
+	State_Pooled                               // Waiting in the pool to be selected and sent for assembly to the transaction's originator
```
annamcallister:
```suggestion
	State_Pooled                               // The transaction is waiting in the pool to be selected and sent for assembly to the its originator
```

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -25,22 +25,24 @@ import (
...
+	State_Endorsement_Gathering                // Assembly completed successfully and we are now waiting for endorsement of the assembled transaction
+	State_Blocked                              // All endorsements have been received but the transaction cannot proceed due to dependencies not being ready for dispatch
+	State_Confirming_Dispatchable              // Endorsed and waiting for confirmation that were are OK to dispatch. The originator can still request not to proceed at this point.
+	State_Ready_For_Dispatch                   // Dispatch confirmation has been received from the originator and we are waiting to be collected by the dispatcher thread. Going into this state is the point of no return to the base ledger.
+	State_Dispatched                           // Collected by the dispatcher thread and submitted by the public TX manager to the base ledger
+	State_Confirmed                            // Confirmed on the base ledger in recent blocks.  NOTE: confirmed transactions are not held in memory for ever so getting a list of confirmed transactions will only return those confirmed recently
```
annamcallister:
```suggestion
	State_Confirmed                            // The transaction has been confirmed on the base ledger. It will remain in this state for a number heartbeat intervals before moving to State_Final to removed from memory.
```

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -25,22 +25,24 @@ import (
...
+	State_Final                                // Final state for the transaction. Transactions are removed from memory as soon as they enter this state
```
annamcallister:
```suggestion
	State_Final                                // The transaction will be removed from memory upon entry to this state
```

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -25,22 +25,24 @@ import (
...
+	State_Evicted                              // Evicted state for a problematic transaction. Transactions are removed from memory as soon as they enter this state. Distinct from State_Final because it might just be used for memory or in-flight slot management
```
annamcallister:
```suggestion
	State_Evicted                              // A problematic transaction is being evicted. Transactions are removed from memory upon entry to this this state. Distinct from State_Final because it might just be used for memory or in-flight slot management
```

core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -25,714 +25,348 @@ import (
...
-	State_Observing     = common.CoordinatorState_Observing     // Not acting as a coordinator but aware of another node acting as a coordinator
-	State_Elect         = common.CoordinatorState_Elect         // Sent HandoverRequest to the active coordinator; waiting to see it start flushing
-	State_Prepared      = common.CoordinatorState_Prepared      // Confirmed the active coordinator is flushing; waiting for its Closing heartbeat before taking over
-	State_Active        = common.CoordinatorState_Active        // Assembling and dispatching transactions; may remain active across epoch boundaries
-	State_Active_Flush  = common.CoordinatorState_Active_Flush  // Draining dispatched transactions while still the active coordinator (key-rotation)
-	State_Closing_Flush = common.CoordinatorState_Closing_Flush // Draining dispatched transactions after stepping down (preemption)
-	State_Closing       = common.CoordinatorState_Closing       // Flush complete; sending closing heartbeats through the grace period
+	State_Initial   State = iota // Coordinator created
+	State_Idle                   // Not acting as a coordinator and not aware of any other active coordinators
```
annamcallister:
```suggestion
	State_Idle                   // Not actively coordinating and not aware of any other active coordinators
```

core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -25,714 +25,348 @@ import (
...
+	State_Observing              // Not acting as a coordinator but aware of another node acting as a coordinator
```
annamcallister:
```suggestion
	State_Observing              // Not actively coordinating but aware of another node actively coordinating
```

core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -25,714 +25,348 @@ import (
...
+	State_Elect                  // Elected to take over from another coordinator and waiting for handover information
```
annamcallister:
```suggestion
	State_Elect                  // Has sent a handover request to an active coordinator and is waiting for that node to stop coordinating
```

core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -25,714 +25,348 @@ import (
...
+	State_Standby                // Going to be coordinator on the next block range but local indexer is not at that block yet
```
annamcallister: This state will be removed after a merge with main

core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -25,714 +25,348 @@ import (
...
+	State_Prepared               // Have received the handover response but haven't seen the flush point confirmed
```
annamcallister:
```suggestion
	State_Prepared               // Has seen the previous active coordinator begin to flush and is waiting for the flush to complete
```

core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -25,714 +25,348 @@ import (
...
+	State_Active                 // Actively coordinating transactions for this domain instance
+	State_Flush                  // Stopped dispatching new transactions but continuing to process transactions that are already dispatched
```
annamcallister: A merge will replace this with 

`State_Active_Flush` // Actively coordinating but waiting for transactions already dispatched to be processed before rotating default signing key and resuming dispatches

`State_Closing_Flush` // Stepping down as active coordinator but continuing to process transactions that have already been dispatched

core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -25,714 +25,348 @@ import (
...
+	State_Closing                // Have flushed and am continuing to send closing status for configured number of heartbeats
```
annamcallister:
```suggestion
	State_Closing                // Has flushed and is continuing to send closing status for configured number of heartbeats
```

core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -25,714 +25,348 @@ import (
...
-	State_Observing     = common.CoordinatorState_Observing     // Not acting as a coordinator but aware of another node acting as a coordinator
-	State_Elect         = common.CoordinatorState_Elect         // Sent HandoverRequest to the active coordinator; waiting to see it start flushing
-	State_Prepared      = common.CoordinatorState_Prepared      // Confirmed the active coordinator is flushing; waiting for its Closing heartbeat before taking over
-	State_Active        = common.CoordinatorState_Active        // Assembling and dispatching transactions; may remain active across epoch boundaries
-	State_Active_Flush  = common.CoordinatorState_Active_Flush  // Draining dispatched transactions while still the active coordinator (key-rotation)
-	State_Closing_Flush = common.CoordinatorState_Closing_Flush // Draining dispatched transactions after stepping down (preemption)
-	State_Closing       = common.CoordinatorState_Closing       // Flush complete; sending closing heartbeats through the grace period
+	State_Initial   State = iota // Coordinator created
```
annamcallister:
```suggestion
	State_Initial   State = iota // Coordinator state machine created
```

core/go/internal/sequencer/originator/transaction/state_machine.go
```diff
@@ -25,22 +25,26 @@ import (
...
-	State_Submitted             = common.OriginatorTransactionState_Submitted             // Submitted to the blockchain
-	State_Confirmed             = common.OriginatorTransactionState_Confirmed             // Confirmed by the blockchain as successful
-	State_Reverted              = common.OriginatorTransactionState_Reverted              // Domain code determined the intent is invalid; finalized as reverted
-	State_Parked                = common.OriginatorTransactionState_Parked                // Domain code determined not ready for assembly; parked for later; remaining transactions may continue unless explicitly dependent
-	State_Final                 = common.OriginatorTransactionState_Final                 // Final state; transactions are removed from memory on entry
+	State_Initial               State = iota // Transaction state machine created
+	State_Pending                            // Intent for the transaction has been created in the database and has been assigned a unique ID but is no confirmation yet that a coordinator is processing it
```
annamcallister:
```suggestion
	State_Pending                            // The transaction has not yet been delegated to a coordinator
```

core/go/internal/sequencer/originator/transaction/state_machine.go
```diff
@@ -25,22 +25,26 @@ import (
...
+	State_Delegated                          // The transaction has been sent to the current active coordinator - we do not know that the coordinator has accepted the transaction as there is no confirmation response to a delegation request but heartbeats will confirm this indirectly
```
annamcallister:
```suggestion
	State_Delegated                          // The transaction has been sent to the current active coordinator
```

The previous comment is fairly out of date as we do have delegation rejections now. And there is a response but we don't listen for it. There is more nuance here than it feels can be captured in a single line comment.

core/go/internal/sequencer/originator/transaction/state_machine.go
```diff
@@ -25,22 +25,26 @@ import (
...
+	State_Assembling                         // The coordinator has sent an assemble request to us and we have not yet sent the assembled transaction back to the coordinator
+	State_Endorsement_Gathering              // We have responded to an assemble request and are waiting the coordinator to gather endorsements before sending us a dispatch confirmation request
```
annamcallister:
```suggestion
	State_Endorsement_Gathering              // An assemble response has been sent to the active coordinator, who should now be gathering endorsements for the transaction. A dispatch confirmation request is expected in this state.
```

annamcallister (commented):
The format of the generated docs look great.

I've made set of suggestions to the comments. Some are to make them more accurate, but the majority are to keep style, particularly wrt 1st/3rd person, consistent. 

I have stopped part way through commenting the originator description as it was getting quite challenging to make them accurate. I'm not sure we have the correct set of states here and I think this would be a helpful conversation to have in person.

For the generation script, is there a reason you've chosen to use python? All other docs generation in the project is in golang, and happens automatically as part of the pipeline, ensuring that docs stay up to date as the code changes. It will add extra challenges to bring another language into the gradle build, which is where this script needs to run from.

core/go/internal/sequencer/originator/transaction/state_machine.go
matthew1001: Yeah I think for the docs it's a balance of not trying to describe the entire state machine logic while still being useful. I prefer your shorter description because it's forcing the reader to go and understand the detail from the code, rather than think they have it all from the one-liner

core/go/internal/sequencer/coordinator/transaction/state_machine.go
matthew1001: This is an interesting one because the whole point-of-no-return really only applies to the non-chained case, and also it sounds somewhat over the top as a characteristic to highlight in the docs. It really means "point of no return for this specific public TX but might fail in a perfectly safe way and then get re-assembled later". So I sort of feel we're over-playing the importance of it. Perhaps we slim this down to your suggestion, minus the point-of-no-return part?

annamcallister (approved):
Approving so you can merge when ready, but there's a final update needed to the top README to specifiy the new python dependency https://github.com/LFDT-Paladin/paladin#building-locally

PR #1217, opened 2026-06-07
Privacy group, State and Message panels

gabriel-indik:
New Privacy groups panel:

image

with lookup options:

image

and details section:

image

New States panel:

image

with details section:

image

New Messages panel:

image

with details section:

image

PR #1216, opened 2026-06-07
Update Domains panel

gabriel-indik:
Update domains panel:
 - Introduced pagination
 - Introduce sorting
 - Introduce lookup
 - Introduce details panel
 - Address issues with Balance, Mint and Transfer dialog boxes
 - Introduce Burn dialog box
 
 
Updated domains panel:

image

New lookup option:

image

New details panel:

image

Sample of addressed dialog box (transfer):

image

New burn option:

image

PR #1218, opened 2026-06-07
State transfer (API + UI)

gabriel-indik:
New `pstate_transferState` JSON RPC method makes it possible to transfer states between nodes. UI panel for states has been updated with the `Send` section that invokes the new API.

https://github.com/user-attachments/assets/b40c4b45-464f-47be-9318-1493962fb548

Note: based on feedback the `Push` action in the screen capture above has been changed to `Send`.


gabriel-indik:
Important: this PR should only be considered/merged after https://github.com/LFDT-Paladin/paladin/pull/1217, which itself depends on https://github.com/LFDT-Paladin/paladin/pull/1216

gabriel-indik:
Replaced by https://github.com/LFDT-Paladin/paladin/pull/1221

PR #1219, opened 2026-06-08
Fix WS reconnect deadlock

annamcallister:
The `wsClient` send loop in `wsclient.go` could exit prematurely due to a network write failure (e.g. `close 1006 abnormal closure`) while `afterConnect` was still running. Because `afterConnect` calls `Send()` to re-establish subscriptions, and `Send()` blocked indefinitely waiting for the send loop to read from its channel, the entire reconnect goroutine deadlocked — no further reconnection attempts were made and the block indexer stopped processing blocks permanently.

The fix adds a `sendDone` channel that the send loop closes on exit, and a `case <-sendDone` arm to `Send()` so callers return `PD021105` immediately instead of blocking. The send loop cleanup (`close(receiverDone)`, drain `sendDone`, close `wsconn`) is now unconditional so it runs even when `afterConnect` returns an error. Finally, `sendDone` is never set back to `nil` between reconnect iterations, so any `Send()` call during the reconnect window picks up the previous iteration's already-closed channel and also returns immediately rather than waiting for a 2-minute RPC timeout.

(Also fixed an `ERROR` log which is not an error and made debug hard)


matthew1001 (approved):
LGTM Anna

PR #1220, opened 2026-06-08
Pull block height to sequencer

annamcallister:
Moves block height updates back to a pull rather than a push model. This ensures we always have an up to date block height when needed, without driving work when its not needed, and without the need for the sequencer manager to queue events for every single coordinator and originator.

### DIFF
--- a/core/go/internal/sequencer/originator/state_machine.go
+++ b/core/go/internal/sequencer/originator/state_machine.go
@@ -92,21 +92,16 @@ var stateDefinitionsMap = StateDefinitions{
 				Match: statemachine.MatchFirst,
 				Handlers: []EventHandler{{
 					Validator: validator_TransactionDoesNotExist,
-					Actions:   []ActionRule{{Action: action_TransactionCreated}},
+					Actions: []ActionRule{
+						{Action: action_TransactionCreated},
+						// If we are idle we have not been receiving heartbeats from a coordinator so we
+						// need to make sure our delegation goes to the current top priority coordinator
+						{Action: action_RefreshBlockHeight},
+						{Action: action_ResetToTopPriorityCoordinator},
+					},
 					Transitions: []Transition{{
 						To: State_Sending,
 					}},
-				}},
-			},
-			common.Event_NewBlock: {
-				Match: statemachine.MatchFirst,
-				Handlers: []EventHandler{{
-					Actions: []ActionRule{
-						{Action: action_UpdateBlockHeight},
-						{If: guard_IsOnEpochBoundary, Action: action_CalculateCoordinatorPriorities},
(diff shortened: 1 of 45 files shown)


core/go/internal/sequencer/originator/transaction/state_machine.go
```diff
@@ -143,8 +143,11 @@ var stateDefinitionsMap = StateDefinitions{
 				}},
 			},
 			Event_AssembleRequestReceived: {
-				Match: statemachine.MatchFirst,
+				Match: statemachine.MatchAll,
 				Handlers: []EventHandler{{
+					// Always runs first: refresh the cached block height before any validator reads it.
+					Actions: []ActionRule{{Action: action_RefreshBlockHeight}},
```
matthew1001, line 149: I know this is sort of what multiple event handlers are for, but I notice we're doing this for a very large number of event types. The PR is intentionally changing to an on-demand/pull approach, so these changes do make sense, but given how many event types it applies to is it worth considering a more central place e.g. in `propagateEventToTransaction`?

core/go/internal/sequencer/originator/state_machine.go
```diff
@@ -165,15 +160,6 @@ var stateDefinitionsMap = StateDefinitions{
 					}},
 				}},
 			},
-			common.Event_NewBlock: {
-				Match: statemachine.MatchFirst,
-				Handlers: []EventHandler{{
-					Actions: []ActionRule{
-						{Action: action_UpdateBlockHeight},
-						{If: guard_IsOnEpochBoundary, Action: action_CalculateCoordinatorPriorities},
```
matthew1001, line 173: Just wanting to confirm. Previously we invoked `action_CalculateCoordinatorPriorities` on every new-block event. With this PR, I think we're calling it just once (for `Event_EndorserNodesDiscovered`). Mainly just confirming that's an expected change here?

core/go/internal/sequencer/originator/transaction/state_machine.go
matthew1001: It would mean calling it in a few places it's not needed, but the overhead vs code simplicity feels like it's worth considering

matthew1001 (approved):
Happy with the changes. Repetition in the state machine continues to be something that PRs like this really highlight. I'm OK with merging in its current state, but it felt worth suggesting that in some cases where low-cost common actions occur when handling many event types, that perhaps a "run before all handlers for all events" hook might emerge as a reasonable compromise of state machine complexity vs code overheads.

core/go/internal/sequencer/originator/state_machine.go
annamcallister: It's called as part of refreshing block height too, if the effective block height has changed. It's hard to surface this in the action names without making them very long.

PR #1221, opened 2026-06-08
State transfer API + UI

gabriel-indik:
New `pstate_transferState` JSON RPC method makes it possible to transfer states between nodes. UI panel for states has been updated with the `Send` section that invokes the new API.

https://github.com/user-attachments/assets/b40c4b45-464f-47be-9318-1493962fb548

Note: based on feedback the `Push` action in the screen capture above has been changed to `Send`.


core/go/internal/statemgr/statestore_rpc.go
```diff
@@ -40,7 +41,8 @@ func (ss *stateManager) initRPC() {
 		Add("pstate_queryStates", ss.rpcQueryStates()).
 		Add("pstate_queryContractStates", ss.rpcQueryContractStates()).
 		Add("pstate_queryNullifiers", ss.rpcQueryNullifiers()).
-		Add("pstate_queryContractNullifiers", ss.rpcQueryContractNullifiers())
+		Add("pstate_queryContractNullifiers", ss.rpcQueryContractNullifiers()).
+		Add("pstate_transferState", ss.rpcTransferState())
```
matthew1001: I have a conservative view here, which is that we should make sure it's very clear to the RPC user/app developer that this sends private state. Not comittments/hashes/state IDs - but actual private state data. So I'd make a case for calling the RPC method `pstate_transferPrivateState`

core/go/internal/statemgr/state_transfer.go
```diff
@@ -0,0 +1,72 @@
...
+		return uuid.Nil, err
+	}
+
+	states, err := ss.GetStatesByID(ctx, dbTX, domain, nil, []pldtypes.HexBytes{stateID}, true, false)
+	if err != nil {
+		return uuid.Nil, err
+	}
+	state := states[0]
+
+	if node == ss.transportManager.LocalNodeName() {
```
matthew1001: Maybe move this to before the DB lookup?

ui/client/src/components/StateActions.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { SendStateDialog } from "../dialogs/SendState";
+import { useState } from "react";
+
+type Props = {
+  state: IState
+}
+
+export const StateActions: React.FC<Props> = ({ state }) => {
+
+  const [sendStateDialogOpen, setSendStateDialogOpen] = useState(false);
```
matthew1001, line 29: Similar to my comment about naming the RPC method to be clear what you are sending, I think we should make sure the UI is similarly clear and gives you plenty of opportunity to understand what it is you're sending. I don't know what the best design approach for that is - maybe a warning/confirmation popup or something that clearly says "Send Private State", but I do think we should make it super obvious what you're sending

core/go/internal/statemgr/state_transfer.go
```diff
@@ -0,0 +1,72 @@
+// Copyright © 2026 Kaleido, Inc.
```
matthew1001, line 1: It looks like this doesn't have 100% test coverage. Do you mind getting it up to 100% and maybe check the other changes files haven't dropped their coverage

ui/client/src/components/StateActions.tsx
gabriel-indik: Made the following changes to the dialog box:

image

core/go/internal/statemgr/state_transfer.go
gabriel-indik: Expanded code coverage to 100%

core/go/internal/statemgr/state_transfer.go
gabriel-indik: Great point, done.

core/go/internal/statemgr/statestore_rpc.go
gabriel-indik: pstate_transferState renamed to pstate_transferPrivateState.

gabriel-indik:
@matthew1001 thank you for the review, I applied the changes based on the feedback.

ui/client/src/components/StateActions.tsx
matthew1001: Think this is super clear, thanks. I'd much rather we start off with something like this and slowly relax it based on feedback

sdk/go/pkg/pldclient/statestore.go
```diff
@@ -33,7 +33,7 @@ type StateStore interface {
 	QueryContractStates(ctx context.Context, domain string, contractAddress pldtypes.EthAddress, schemaRef pldtypes.Bytes32, query *query.QueryJSON, qualifier pldapi.StateStatusQualifier) (states []*pldapi.State, err error)
 	QueryNullifiers(ctx context.Context, domain string, schemaRef pldtypes.Bytes32, query *query.QueryJSON, status pldapi.StateStatusQualifier) (states []*pldapi.State, err error)
 	QueryContractNullifiers(ctx context.Context, domain string, contractAddress pldtypes.EthAddress, schemaRef pldtypes.Bytes32, query *query.QueryJSON, status pldapi.StateStatusQualifier) (states []*pldapi.State, err error)
-	TransferState(ctx context.Context, domain string, stateID pldtypes.HexBytes, recipient pldtypes.PrivateIdentityLocator) (reliableMessageID uuid.UUID, err error)
+	TransferPrivateState(ctx context.Context, domain string, stateID pldtypes.HexBytes, recipient pldtypes.PrivateIdentityLocator) (reliableMessageID uuid.UUID, err error)
```
matthew1001, line 36: I'm happy to approve as-is, but it might make sense to change `TransferPrivateState` back to `TransferState` (which `pstate_transferPrivateState` calls). I think the code in this file is more for Paladin contributors than app developers, and since the rest of the codebase tends to use the word `State` for private state I'm happy if this one is changed back.

sdk/go/pkg/pldclient/statestore.go
```diff
@@ -64,7 +64,7 @@ var stateStoreInfo = &rpcModuleInfo{
 			Inputs: []string{"domain", "contractAddress", "schemaRef", "query", "qualifier"},
 			Output: "states",
 		},
-		"pstate_transferState": {
+		"pstate_transferPrivateState": {
```
matthew1001, line 67: 👍

matthew1001 (approved):
Thanks @gabriel-indik one minor comment but happy to approve as it stands

sdk/go/pkg/pldclient/statestore.go
gabriel-indik: I'll include this item in the next PR.

PR #1222, opened 2026-06-09
Include all transactions in delegation message

annamcallister:
This must have been omitted earlier on. The transport writer interface and transactions delegated event in the coordinator accept multiple transactions, but then at the point of sending each transaction was being put into its own message severing the FIFO ordering between them.


matthew1001 (approved):
Unrelated to the PR changes, but we should potentially check if we're compressing payloads across the wire

PR #1223, opened 2026-06-09
Add filter to indexed txs

gabriel-indik:
The JSON RPC method `bidx_queryIndexedTransactions` has been updated to include an additional (boolean) argument called `hasPaladinReceipt`. If the value of the new argument is `true`, the entries returned will be filtered to only include indexed transactions that contain at least one Paladin receipt.


core/go/pkg/blockindexer/block_indexer.go
```diff
@@ -1087,13 +1087,16 @@ func (bi *blockIndexer) QueryIndexedBlocks(ctx context.Context, jq *query.QueryJ
...
-func (bi *blockIndexer) QueryIndexedTransactions(ctx context.Context, jq *query.QueryJSON) ([]*pldapi.IndexedTransaction, error) {
+func (bi *blockIndexer) QueryIndexedTransactions(ctx context.Context, jq *query.QueryJSON, hasPaladinReceipt bool) ([]*pldapi.IndexedTransaction, error) {
 	ctx = log.WithComponent(ctx, "blockindexer")
 	if jq.Limit == nil || *jq.Limit == 0 {
 		return nil, i18n.NewError(ctx, msgs.MsgBlockIndexerLimitRequired)
 	}
 	db := bi.persistence.DB()
 	q := db.Table("indexed_transactions").Joins("Block").WithContext(ctx)
+	if hasPaladinReceipt {
+		q = q.Where(`EXISTS (SELECT 1 FROM transaction_receipts WHERE "transaction_receipts"."tx_hash" = "indexed_transactions"."hash")`)
```
matthew1001, line 1098: This matches an index we already have on transaction_receipts so it looks fine to me

ui/client/src/queries/rpcMethods.ts
```diff
@@ -56,5 +56,6 @@ export const RpcMethods = {
   pstate_listSchemas: 'pstate_listSchemas',
   pstate_queryStates: 'pstate_queryStates',
   transport_queryReliableMessages: 'transport_queryReliableMessages',
-  pstate_transferState: 'pstate_transferState'
+  pstate_transferState: 'pstate_transferState',
+  bidx_QueryIndexedTransactionsWithReceipt: 'bidx_queryIndexedTransactionsWithReceipt'
```
matthew1001: nit: maybe bump to the top with the other `bidx_*` methods

matthew1001 (approved):

ui/client/src/queries/rpcMethods.ts
gabriel-indik: Sorted all RPC methods.

PR #1225, opened 2026-06-11
Update state machine doc generator after merge

matthew1001:
Follows on from https://github.com/LFDT-Paladin/paladin/pull/1215. When merging `main` into that PR the state machine changes (in code) required an update to the generation script.

### DIFF
--- a/doc-site/docs/architecture/distributed_sequencer_state_machine.md
+++ b/doc-site/docs/architecture/distributed_sequencer_state_machine.md
@@ -10,12 +10,50 @@ The distributed sequencer is designed as a set of state machines, each of which 
 %%{init: {'themeVariables': {'background': 'transparent'}}}%%
 stateDiagram-v2
     direction LR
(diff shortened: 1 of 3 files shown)


annamcallister (approved):
Approving as we need these changes.

But to me I think this highlights that it is not sustainable to generate from code and comment structure alone and we do need to reconsider on using gostruct tags tied to proper descriptions. I think the risk that a description could get out of date is less significant than the entire page becoming broken due to a refactor.

matthew1001:
> Approving as we need these changes.
> 
> But to me I think this highlights that it is not sustainable to generate from code and comment structure alone and we do need to reconsider on using gostruct tags tied to proper descriptions. I think the risk that a description could get out of date is less significant than the entire page becoming broken due to a refactor.

100% agree. I think we'll need to continue iterating on it to land on the most sustainable approach for these sections of docs

PR #1227, opened 2026-06-11
Fix reliable message post commit race

annamcallister:
Addresses two bugs found in transport manager

---
When multiple reliable messages are being inserted, there is no guarantee that their commits will complete in the same order as their assigned sequence numbers. The post commit notifies the peer that there are messages to send, and if it gets notified before This could result in the `lastDrainHWM` being advanced past a message sequence that is not yet committed. This message will not be sent until the full scan runs, as the sequence is below the `lastDrainHWM`.

This PR replaces `lastDrainHWM` with a `lowestPendingSeq`. This is reset between each triggered scan, so it can always be lowered for the lowest possible sequencer the triggered scan may need to send. In the case where we've had out of order commits, this will likely result in the message with the later sequencer/earlier commit being immediately resent (where we usually enforce the resend interval before a retry), but this is a) tolerated by at least once delivery and b) isn't going to be the sort of endless spamming of unacked messages that the age gate is supposed to prevent.

---
If there is a problem in draining the fire and forget send queue (likely because of a bug somewhere else in the system), a peer can be reaped while there are still messages on its queue to send. This is not a problem, since the semantics are at most once delivery. What is a problem is that if a sender is waiting in a select on the buffered channel, once the peer is reaped, there is no way out of that select. This PR adds an additional `senderDone` check into the select to prevent this blocking behaviour.

### DIFF
--- a/core/go/internal/transportmgr/manager.go
+++ b/core/go/internal/transportmgr/manager.go
@@ -316,6 +316,9 @@ func (tm *transportManager) queueFireAndForget(ctx context.Context, nodeName str
 		return nil
 	case <-ctx.Done():
 		return i18n.NewError(ctx, msgs.MsgContextCanceled)
+	case <-p.senderDone:
+		log.L(ctx).Warnf("peer %s sender stopped before message %s/%s could be queued (discarding)", p.Name, msg.MessageType, msg.MessageId)
+		return nil
 	}
 }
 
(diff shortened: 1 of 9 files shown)


core/go/internal/transportmgr/manager.go
```diff
@@ -316,6 +316,9 @@ func (tm *transportManager) queueFireAndForget(ctx context.Context, nodeName str
 		return nil
 	case <-ctx.Done():
 		return i18n.NewError(ctx, msgs.MsgContextCanceled)
+	case <-p.senderDone:
```
matthew1001, line 319: Is this effectively the same fix we made in https://github.com/LFDT-Paladin/paladin/pull/1219/changes#diff-248c293a6af676c5462f146cecc009718485103499562fded53131a4bea662c4? Just wondering if there are other places in the codebase that currently use a similar (broken) pattern

matthew1001 (approved):

core/go/internal/transportmgr/manager.go
annamcallister: yes it is!

I can't think of anywhere else we have this idea of reconnecting to the same target of the top of my head but will do a scan of the code.

PR #1228, opened 2026-06-11
Calculate inactive grace period based on full heartbeat interval periods

annamcallister:
We have been calculating the inactive grace period based on number of heartbeat interval events received since something last occurred. I'm proposing that we change this to being full heartbeat interval periods instead. These are almost equivalent, you just need to see one more event when calculating based on full periods, but it gives a neater experience around specifying minimums.

We currently have a minimum of `1` but this is broken in the current model. Even in a well functioning system we are guaranteed to get an interval event in between heartbeats, meaning an originator would always consider a coordinator inactive at that point and constantly redelegate.

With our default of `2`, there are still some problematic edge cases. If we see, for example, with a 10s heartbeat interval
- T - heartbeat arrives
- T + 1ms - heartbeat interval event fires
- T + 10,0001 ms - heartbeat interval event fires
- T + 10,0002 ms - heartbeat arrives

We're going to consider the coordinator inactive - yet given heartbeats are processed on what might be a busy event loop, and sent over the network, so it is expected that we might see some variation above and below the 10s in when we get heartbeats.

By counting full interval periods we can keep our minimum at 1 (probably not desirable but fine if you want very low tolerance for latency in a system that is not overloaded), and our default at 2. Whereas it feels harder to explain a minimum of 2 if we don't change how we do the counting.

### DIFF
--- a/core/go/internal/sequencer/coordinator/transaction/finalizing.go
+++ b/core/go/internal/sequencer/coordinator/transaction/finalizing.go
@@ -24,7 +24,9 @@ import (
 func guard_HasFinalizingGracePeriodPassedSinceStateChange(ctx context.Context, txn *coordinatorTransaction) bool {
 	// Has this transaction been in the same state for longer than the finalizing grace period?
 	// most useful to know this once we have reached one of the terminal states - Reverted or Committed
-	return txn.heartbeatIntervalsSinceStateChange >= txn.finalizingGracePeriod
+	// measure number of complete heartbeat interval periods - e.g. count of 2 means
+	// 1 full heartbeat interval has elapsed, hence use of > not >=
+	return txn.heartbeatIntervalsSinceStateChange > txn.finalizingGracePeriod
 }
 
 // action_FinalizeAsUnknownByOriginator is called when the originator reports that it doesn't recognize
--- a/core/go/internal/sequencer/coordinator/inactive_test.go
+++ b/core/go/internal/sequencer/coordinator/inactive_test.go
@@ -84,10 +84,18 @@ func Test_guard_InactiveGracePeriodExceeded_NotExceeded(t *testing.T) {
 	assert.False(t, guard_InactiveGracePeriodExceeded(ctx, c))
 }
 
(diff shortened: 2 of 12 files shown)


core/go/internal/sequencer/coordinator/transaction/finalizing.go
```diff
@@ -24,7 +24,9 @@ import (
 func guard_HasFinalizingGracePeriodPassedSinceStateChange(ctx context.Context, txn *coordinatorTransaction) bool {
 	// Has this transaction been in the same state for longer than the finalizing grace period?
 	// most useful to know this once we have reached one of the terminal states - Reverted or Committed
-	return txn.heartbeatIntervalsSinceStateChange >= txn.finalizingGracePeriod
+	// measure number of complete heartbeat interval periods - e.g. count of 2 means
```
matthew1001, line 27: I'm struggling to understand this description. Is it saying that if you have 2 heartbeat events, then 1 heartbeat interval (the time between them) has passed?

matthew1001 (commented):
I'm just trying to make sure I understand the change fully. My comment on one of the line changes is probably the crux of what I'm trying to understand.

Base on the PR description:

> We have been calculating the inactive grace period based on number of heartbeat interval events received since something last occurred. I'm proposing that we change this to being full heartbeat interval periods instead. These are almost equivalent, 

it sounds like we're saying we're changing from counting 2 x interval events = 2 x inactive period, to 2 x interval events = 1 inactive period? But it doesn't feel like these are almost equivalent so I'm not sure I've fully understood the change

core/go/internal/sequencer/coordinator/transaction/finalizing.go
annamcallister: Yes - that's it.

In terms of how its calculated it's very close - if you express number of events as `N`, the number of full periods is `N-1`, conceptually they're more different.

Full periods feels like it makes more sense to me though since
- the config parameter names are `inactiveGracePeriod` and `closingGracePeriod`
- they allow more intuitive minimum values

matthew1001 (approved):
LGTM

PR #1235, opened 2026-06-11
Updated UX layout and filters

gabriel-indik:
### Backend / API

- Expands flexible JSON query/filtering across resources (transactions, messages, states, privacy groups, submissions, keys, etc.).
- Adds address-field querying and key-address filtering in the key manager API.
- Includes transport manager updates and SDK/doc changes to support the new capabilities.

### UI platform
- Upgrades the client to a newer Vite toolchain and Node 20.19.0.
- Updates dependencies to address security vulnerabilities.

### UX overhaul
- Applies a consistent new layout across all panels.
- Adds collapsible, type-aware filters on every list view, with filter visibility persisted in app state.
- Adds lookup/search affordances on panels and drill-down navigation from lists to individual item detail views (registry entries, states, transactions, messages, etc.).
- Refactors major panels: domains, keys, submissions, registry, states, transactions, privacy groups, and peers.
- Introduces or expands dedicated views for privacy group details/messages, reliable messages, and a new transports panel.
- Removes polling in favor of more efficient data fetching, with assorted polish (empty states, sorting defaults, flicker fixes).


gabriel-indik:
The following screenshots show how the new style has been uniformly applied to all panels:

image

image

image

image

image

image

image

image

image

jimthematrix:
## Keys
1. For the "Keys" page, I think it's missing the "Key URI" field, as that's the value used to identify the signing key when submitting transactions.
2. Not sure if "Path" is the right term for the first column, or was it "Key name"?
3. Not sure if the "Index" column is worth calling out separately, as that's already indicated in the "Key Handle" string.

## Domains
1. Would be useful to display the type of domain (Noto, Zeto or Pente).
2. what is the significance of the "0x5b41..." label next to the "Noto" dropdown?
3. Not clear what `Is Notary` means.

## States
1. it seems the list would be more helpful if it's per-contract. such that the contract address is another filter selection on the top after the domain selection. don't think mixing states from different contracts in the same list is useful.

## Transports
1. "Node" column, is that meant to be the "peer node"? it's not clear as-is what it's representing.

gabriel-indik:
@jimthematrix thank you for the feeback.

### Keys

Here is an example of how data returned from the API is represented visually:

image

```json
{
            "isKey": true,
            "hasChildren": true,
            "parent": "",
            "path": "assetIssuer",
            "name": "",
            "index": 17,
            "wallet": "signer-1",
            "keyHandle": "m/44'/60'/17'",
            "verifiers": [
                {
                    "verifier": "0x050a7bb1df8708974da0ad09bb0943a42436c9bc",
                    "type": "eth_address",
                    "algorithm": "ecdsa:secp256k1"
                }
            ]
        }
```

Note also that the `Keys` panel can optionally be used in "Explorer" mode, which makes it possible to navigate entries like in a file system with files and folders:

https://github.com/user-attachments/assets/06026697-335b-42be-b66f-9a8240634fde

Note that as one explores the entries, the one picks "segments of the path" and the full path is shown at the top in the form of a breadcrumb. Note also that in the API, the value returned for those items is also called `path`. 

With regards to the index column, the reason it is there is so that the user can perform sorting ascending/descending by index.

### Domains

There is a dropdown in the `Domains` panel that is dynamically populated with the domains (by making the corresponding API call) and one can select a domain using that dropdown. When doing so, the rest of the panel updates. The tag that appears next to the domain is its registered address. The following capture shows this in action:


https://github.com/user-attachments/assets/f6a06dd9-c1cf-4b6e-9d50-f9de0fb3b755

With regards to including "Is Notary" that was requested in https://github.com/LFDT-Paladin/paladin/issues/699#issue-3147476120

### States

The way the `States` panel works is: one first has to select a domain, and then a schema. Once the schema is selected we can make the corresponding API calls to get the list of states. Here is a capture to show how it works:


https://github.com/user-attachments/assets/ff34e1f3-b21f-4d97-ab0b-36f5b4cd035b

### Transport

On the `Transport Active Peer Connections` table, "Node" shows the name of the peer that the connection has been established against. We could optionally change the column to be `Peer name` instead. On the `Transport Reliable Messages` table, the column `node` represent the node that the message was exchanged with, which is the same terminology used in the API:

```json
        {
            "sequence": 1032,
            "id": "244b52ef-1d9a-4921-894d-3e0468591432",
            "created": "2026-06-23T03:20:16.513896423Z",
            "node": "node2",
            "messageType": "state",
            "metadata": {
                "stateId": "0x2fe82416634c41e04ca18888141061bfe820a25d732bc2a0e08cfb3b45a7ffe6",
                "identityLocator": "a@node2",
                "domain": "noto",
                "contractAddress": "0xfc02f965c346918d5faf99d8940721577b74fc53",
                "schemaId": "0x39aa9c8bb0d781b43c3da1beff4ab34043e45e007316ce77412c243868815a02"
            },
            "ack": {
                "time": "2026-06-23T03:20:17.089645048Z"
            }
        },
```

Please let me know if you have any other feedback / items.

core/go/internal/filters/field_bytes32.go
```diff
@@ -1,4 +1,4 @@
-// Copyright © 2024 Kaleido, Inc.
+// Copyright © 2026 Kaleido, Inc.
```
matthew1001, line 1: I don't think we need to bump all of these copyright statements if they haven't been touched

matthew1001: FYI if you do touch a file and want to update the copyright, we've been given the steer from the Linux Foundation to use

```
Copyright contributors to Paladin, an LFDT project
```

So for new files we've started adopting that. When updating copyright because a file has been modified I'd suggest changing it to the above. If a file hasn't been touched I don't think we need to change the copyright

sdk/go/pkg/pldclient/transport.go
```diff
@@ -54,6 +55,10 @@ var transportInfo = &rpcModuleInfo{
 			Inputs: []string{},
 			Output: "peers",
 		},
+		"transport_peersWithQuery": {
```
matthew1001, line 58: Is there a reason to have a new JSON/RPC method, or could we extend `transpore_peers` to take an optional query, so without the query it would behave like it does today?

core/go/internal/filters/field_bytes32.go
gabriel-indik: This is what happened: there was another PR (commit within PR https://github.com/LFDT-Paladin/paladin/commit/a0fe9c451d1b8a64f292f422229fe7c5398bf984) where that file was modified, but the header was not updated. That PR and this one are based on the same branch, so they both contained the same changes to that file except that this PR also updates the header. After that other PR was merged, when calculating the delta between what's in `main` and this branch, the copyright is the only difference. In other words, the file was effectively modified, but the header was not updated. This PR rectifies that.

sdk/go/pkg/pldclient/transport.go
gabriel-indik: The reason I introduced a new method instead of modifying the existing one was to ensure that the code is backward compatible. Because of the way JSON RPC methods are implemented in Paladin, I thought there wasn't a simply way to add optional arguments, I'll check again.

core/go/internal/groupmgr/messages.go
```diff
@@ -51,6 +51,7 @@ func (persistedMessage) TableName() string {
 var messageFilters = filters.FieldMap{
 	"localSequence": filters.Int64Field("local_seq"),
 	"domain":        filters.StringField("domain"),
+	"node":          filters.StringField("node"),
```
matthew1001, line 54: It looks like we don't have an index that covers `node`. So existing queries that don't specify `node` won't be affected, but a new query by `node` will have very poor performance.

We ought to add an index to the DB to ensure efficient per-node queries

core/go/internal/keymanager/dbtypes.go
```diff
@@ -56,4 +56,5 @@ var KeyEntryFilters filters.FieldSet = filters.FieldMap{
 	"path":        filters.StringField("path"),
 	"wallet":      filters.StringField("wallet"),
 	"keyHandle":   filters.StringField("key_handle"),
+	"verifier":    filters.StringField("verifier"),
```
matthew1001, line 59: This is covered by an existing index, so should be safe to add

core/go/internal/keymanager/manager.go
```diff
@@ -331,6 +331,7 @@ func (km *keyManager) QueryKeys(ctx context.Context, dbTX *gorm.DB, jq *query.Qu
 	)
 
 	q.Joins("LEFT OUTER JOIN key_mappings ON key_paths.path = key_mappings.identifier")
+	q.Joins("LEFT OUTER JOIN key_verifiers ON key_paths.path = key_verifiers.identifier")
```
matthew1001, line 334: Analysing this effect on DB performance suggests that a key with N verifier rows produces N rows in the join result. It looks like the join should only be added conditionally based on whether `verifier` is stipulated in the query. Without that change I don't think this is safe to include

core/go/internal/transportmgr/manager.go
```diff
@@ -88,6 +88,7 @@ var reliableMessageFilters = filters.FieldMap{
 	"created":     filters.TimestampField("created"),
 	"node":        filters.StringField("node"),
 	"messageType": filters.StringField("msg_type"),
+	"ack.time":    filters.TimestampField(`"Ack"."time"`),
```
matthew1001, line 91: This also doesn't have an index, so I think this needs one for any new queries to be performant without a full table scan

core/go/pkg/blockindexer/internal_types.go
```diff
@@ -30,9 +30,10 @@ var IndexedBlockFilters filters.FieldSet = filters.FieldMap{
 var IndexedTransactionFilters filters.FieldSet = filters.FieldMap{
 	"hash":             filters.HexBytesField(`"indexed_transactions"."hash"`),
 	"blockNumber":      filters.Int64Field("block_number"),
+	"blockTimestamp":   filters.Int64Field(`"Block"."timestamp"`),
```
matthew1001, line 33: I think this also has some DB performance implications

matthew1001 (changes requested):
I think there are enough DB index issues to address that it's worth us taking the time to assess them one by one and convince ourselves we're happy we haven't regressed anything.

sdk/go/pkg/pldclient/transport.go
gabriel-indik: Solution proposed in https://github.com/kaleido-io/paladin/pull/74

gabriel-indik:
This PR has been replaced by https://github.com/LFDT-Paladin/paladin/pull/1276 and https://github.com/LFDT-Paladin/paladin/pull/1275

PR #1236, opened 2026-06-12
Switch to translatable files for state descriptions; use go app not python

matthew1001:
Following on from https://github.com/LFDT-Paladin/paladin/pull/1225 this is probably my last iteration on auto-generating state machine docs for a while, but it's probably a useful update:

 - Use a Go app to generate the docs, not python
    - Removed python from the pre-reqs in the readme
 - Copied the comments to translatable message files
    - I left the code comments since they are still valid & useful, but they are no longer used to create the `.md` files


annamcallister (changes requested):
The approach looks good but the file locations are inconsistent with existing code

* core messages are all currently [here](https://github.com/LFDT-Paladin/paladin/tree/main/core/go/internal/msgs) rather than colocated within packages
* docs generation all lives [here](https://github.com/LFDT-Paladin/paladin/tree/main/toolkit/go/pkg/reference)

matthew1001:
> The approach looks good but the file locations are inconsistent with existing code
> 
> * core messages are all currently [here](https://github.com/LFDT-Paladin/paladin/tree/main/core/go/internal/msgs) rather than colocated within packages
> * docs generation all lives [here](https://github.com/LFDT-Paladin/paladin/tree/main/toolkit/go/pkg/reference)

Thanks @annamcallister  I've addressed the latter. Re the messages location, I intentionally put them close to the state machine itself to help avoid forgetting to update it in future state machine changes. Do you think they are OK there for that reason, or still want me to move them?

annamcallister (approved):

PR #1237, opened 2026-06-12
Address tx lookup dialog

gabriel-indik:
The lookup dialog box in the transactions panel works when providing either a transaction hash or a Paladin transaction ID. With the changes in this PR, it will also work providing a Paladin receipt ID.


ui/client/src/dialogs/TransactionLookup.tsx
```diff
@@ -81,7 +81,7 @@ export const TransactionLookupDialog: React.FC<Props> = ({
         }
       });
     } else if (isValidUUID(hashOrId)) {
-      paladinTransactionById().then(result => {
+      paladinReceiptById().then(result => {
```
awrichar, line 89: Will this only work if I have the receipt? It might be good if it worked for pending transactions or failed transactions as well. I think receipts are the _more_ likely use case (usually I'm looking up confirmed things), but we might need a fallback.

awrichar (approved):

ui/client/src/dialogs/TransactionLookup.tsx
gabriel-indik: Excellent point @awrichar I updated the PR to make sure that we first try to get the receipt (if the data provided is a UUID as opposed to a hash in which case we would try to get the blockchain transaction), and if that fails, then we try to get the Paladin transaction. This covers the scenario of transactions that do not have a receipt.

PR #1243, opened 2026-06-16
Add missing originator notifications for chained dependency failure

annamcallister:
A missing notification for when a transaction had finalized as reverted because of a chained dependency failure resulted in the originator repeatedly redelegating the transaction.

### DIFF
--- a/core/go/internal/sequencer/coordinator/transaction/pooled.go
+++ b/core/go/internal/sequencer/coordinator/transaction/pooled.go
@@ -22,6 +22,7 @@ import (
 	"github.com/LFDT-Paladin/paladin/core/internal/msgs"
 	"github.com/LFDT-Paladin/paladin/core/internal/sequencer/common"
 	"github.com/LFDT-Paladin/paladin/core/internal/sequencer/syncpoints"
+	"github.com/LFDT-Paladin/paladin/core/pkg/proto/engine"
 	"github.com/google/uuid"
 )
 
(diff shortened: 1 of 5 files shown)


core/go/internal/sequencer/coordinator/transaction/pooled.go
```diff
@@ -183,6 +184,20 @@ func action_FinalizeOnRevertedChainedDependencyAtCreation(ctx context.Context, t
 	return nil
 }
 
+func action_NotifyOriginatorOfChainedDependencyFailureAtCreation(ctx context.Context, t *coordinatorTransaction, _ common.Event) error {
```
matthew1001, line 187: Just looking at the other actions we have for sending notifications to the originator:
 - `action_NotifyOriginatorOfRetryableRevert`
 - `action_NotifyOriginatorOfNonRetryableRevert`

I wonder if we are going to slowly add new `action_NotifyOriginatorOf*` methods for specific cases, or whether we can re-use generic ones in a way that allows a specific failure message not retrieved from `t.decodedRevertReason` to be returned.

Just pushing on the discussion of whether we add niche `action_*` functions for every similar-but-different case, so I'm happy if the answer is to stick with the way you've got it here.

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -175,7 +178,10 @@ var stateDefinitionsMap = StateDefinitions{
 			Event_ChainedDependencyFailed: {
 				Match: statemachine.MatchFirst,
 				Handlers: []EventHandler{{
-					Actions:     []ActionRule{{Action: action_FinalizeOnChainedDependencyFailure}},
+					Actions: []ActionRule{
+						{Action: action_FinalizeOnChainedDependencyFailure},
```
matthew1001, line 182: This pushes on a similar-but-different question. Could `action_FinalizeOnChainedDependencyFailure` just do the originator notifying as well?

matthew1001 (approved):
I'm fine with it as it stands, but left some comments for discussion

core/go/internal/sequencer/coordinator/transaction/pooled.go
annamcallister: Discussed in person- it's not possible to capture the right information from the event alone if we have a single action here

core/go/internal/sequencer/coordinator/transaction/state_machine.go
annamcallister: Discussed in person- in the retryable revert case there's a notification without finalizing- keeping separate actions for notification and finalization keeps things consistent.

PR #1244, opened 2026-06-16
Cancel inflight assembly if new request is received

annamcallister:
An originator is not told if a coordinator cancels the assembly it is currently actioning. Continuing with the assembly does not block the arrival of new assembly requests, but its successful completion will cause the originator transaction state machine to transition out of assembling state, which means it does not handle the assemble result of a second assembly, so does not send a response to the coordinator. The coordinator will time the assembly out at the state timeout, and the transaction will be repooled. This manifests as excessive latency from submit through to confirm of the transaction.

This PR ensures that
* An originator cancels an inflight assembly request if a new one is received
* Does not move out of assembling state if an assembly result is not for the most recently received request.

There a further possible enhancement here to send a notification when assembly is cancelled, rather than letting the only signal be that a new assembly has been request. In the case that there is a large pool of transactions, the new assembly request may be some time later, in which case the originator has wastefully continued on an assembly which isn't needed.

### DIFF
--- a/core/go/internal/sequencer/originator/transaction/assembling.go
+++ b/core/go/internal/sequencer/originator/transaction/assembling.go
@@ -74,6 +74,9 @@ func action_AssembleError(ctx context.Context, t *originatorTransaction, event c
 // transaction event loop unblocked while allowing the potentially slow AssembleAndSign
 // call to run concurrently.
 //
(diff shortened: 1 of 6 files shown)


matthew1001 (approved):
At some point I'm going to raise a PR just to swap all the instances of `txn` for `t` or vice versa :)

Issue #1230, opened 2026-06-11
Enhancement: Java SDK - core data models & primitives


divyam-jha123:
### What would you like to be added?

Implement the core Java types that mirror Paladin's data model, so all
higher-level modules (client, domains, txbuilder) have a shared type vocabulary.

### Reference implementations
- `sdk/go/pkg/pldtypes/` — primitive types (hex bytes, addresses, hashes, etc.)
- `sdk/go/pkg/pldapi/` — transaction, receipt, ABI, query, state types

### Scope
- [x] Java classes for transaction types (submit, receipt, status)
- [x] ABI types (function, event, parameter)
- [x] Query filter / pagination types
- [x] Address, hash, hex-bytes value types
- [x] Custom Jackson `ObjectMapper` with appropriate serializers/deserializers
- [x] Unit tests: round-trip JSON serialization for all types

### Out of scope
HTTP transport, RPC calls, domain-specific types.

### Acceptance criteria
- All core types serialize/deserialize correctly against sample JSON payloads
- No runtime Jackson configuration required by consumers

### Why is this needed?

Part of the Java SDK RFC #1224

Issue #1239, opened 2026-06-16
Enhancement: Java SDK - remaining RPC modules (blockindex, pgroup, registry, statestore, transport)


divyam-jha123:
### What would you like to be added?

Implement the remaining RPC namespace clients on top of the HTTP transport, so the
Java SDK reaches parity with the Go `pldclient` for all base RPC namespaces.

After Phase 1 delivered `ptx_*` and `keymgr_*` (#1232), the remaining namespaces are:

- `bidx_*` (block index) — mirror `sdk/go/pkg/pldclient/blockindex.go`
- `pgroup_*` (privacy groups) — mirror `sdk/go/pkg/pldclient/pgroup.go`
- `reg_*` (registry) — mirror `sdk/go/pkg/pldclient/registry.go`
- `statestore_*` (state store) — mirror `sdk/go/pkg/pldclient/statestore.go`
- `transport_*` (transport) — mirror `sdk/go/pkg/pldclient/transport.go`

### Reference implementations
- `sdk/go/pkg/pldclient/` — one file per namespace (`blockindex.go`, `pgroup.go`, `registry.go`, `statestore.go`, `transport.go`)

### Scope
- [ ] `BlockIndexClient` — all `bidx_*` RPC methods
- [ ] `PrivacyGroupClient` — all `pgroup_*` RPC methods
- [ ] `RegistryClient` — all `reg_*` RPC methods
- [ ] `StateStoreClient` — all `statestore_*` RPC methods
- [ ] `TransportClient` — all `transport_*` RPC methods
- [ ] Request/response types specific to these namespaces (shared types in `core`)
- [ ] Unit tests for each method using mock HTTP responses

### Out of scope
WebSocket subscriptions, domain helpers, the TxBuilder (delivered separately).

### Acceptance criteria
- Full method parity with the Go equivalents for each namespace
- Each method is covered by at least one unit test

### Why is this needed?

Part of the Java SDK RFC #1224

PR #1245, opened 2026-06-16
 Core data models & primitives — primitive value types

divyam-jha123:
Implements the foundational primitive value types for the Java SDK, mirroring
`sdk/go/pkg/pldtypes`. These give every higher-level module (client, domains,
txbuilder) a shared type vocabulary.

### Issue #1230 — group checklist
- [x] **Group 1 — Primitive value types** (this PR): `HexBytes`, `EthAddress`,
      `Bytes32`, `HexUint256`, `HexInt256`, `HexUint64`, `Timestamp` + round-trip tests
- [x] **Group 2 — Custom `ObjectMapper`**: `PaladinObjectMapper` bundling SDK-wide
      defaults (JavaTimeModule, unknown-property policy, null handling)
- [x] **Group 3 — ABI types**: `AbiEntry` (function/event/error/constructor),
      `AbiParameter`, type enums
- [x] **Group 4 — Transaction types**: `TransactionInput`, `TransactionReceipt`,
      transaction type/status enums (mirror `pldapi/transaction.go`)
- [x] **Group 5 — Query types**: filter / sort / pagination

### What's in this PR

Five groups of immutable, self-serializing types. Each carries its own Jackson annotations, so they round-trip through any ObjectMapper with no consumer-side configuration — the tests use a plain new ObjectMapper().

#### Group 1 — Primitive value types (core.types, mirrors pldtypes)

| Type       | Backing         | JSON form                     |
|------------|-----------------|-------------------------------|
| HexBytes   | byte[]          | 0x lower-case hex             |
| EthAddress | 20 bytes        | 0x + 40 hex                   |
| Bytes32    | 32 bytes        | 0x + 64 hex                   |
| HexUint256 | BigInteger      | 0x even-padded hex            |
| HexInt256  | BigInteger      | signed ±0x hex                |
| HexUint64  | long (unsigned) | 0x hex, full 2⁶⁴ range        |
| Timestamp  | long nanos      | RFC3339-nano UTC, zero ↔ null |

Integer types deserialize from a JSON string or number using BigInteger/long to avoid float precision loss (mirrors Go's UseNumber); exact hex formatting preserved; Timestamp reproduces the sec/milli/nano magnitude inference and the zero↔null round-trip.

#### Group 2 — Custom ObjectMapper (core.json)
PaladinObjectMapper bundles the SDK-wide defaults that higher-level modules rely on: JavaTimeModule, ISO-8601 dates, unknown-property tolerance (forward compatibility), null omission, and BigInteger/BigDecimal for untyped numbers.

#### Group 3 — ABI types (core.abi, mirrors firefly-signer)
AbiEntry (function/constructor/event/error/fallback/receive), AbiParameter (incl. nested tuple components), and the EntryType / StateMutability enums (lower-case tokens, case-insensitive parsing). Fluent builders for entries and parameters.

#### Group 4 — Transaction types (core.transaction, mirrors pldapi/transaction.go)
TransactionInput and TransactionReceipt (each flattening the Go embedded structs into the flat JSON wire form), plus the TransactionType (private/public) and SubmitMode (auto/external/call/prepare) enums. data is modelled as JsonNode (Go's RawJSON); receipt on-chain indices use boxed Long so "not finalized on-chain" is distinct from a genuine 0.

#### Group 5 — Query types (core.query, mirrors sdk/go/pkg/query)
QueryJSON (filter / sort / pagination) and a fluent QueryBuilder mirroring Go's builder API (equal/notEqual/gt/gte/lt/lte/in/notIn/isNull/isNotNull/like/notLike/or/limit/sort). Operands modelled by Op / OpSingleVal / OpMultiVal with a QueryModifier enum for the not/caseInsensitive flags.

> Subsequent groups will land as follow-up PRs to keep each reviewable.

### Acceptance criteria
- [x] All types in this PR serialize/deserialize correctly against sample JSON
- [x] No runtime Jackson configuration required by consumers (tests use a plain
      `new ObjectMapper()`)

ref #1230


divyam-jha123:
hey @matthew1001 need to discuss few design decisions: from  group 4 ( https://github.com/LFDT-Paladin/paladin/pull/1245/changes/4da54c430986909e4225ee91815e22839b23d2e1 ) - 

1. `Flattening vs. a reusable PublicTxOptions` — Go embeds TransactionBase/PublicTxOptions, but the JSON wire form is flat, and Jackson won't allow @JsonUnwrapped on @JsonCreator parameters. I flattened gas/value/maxPriorityFeePerGas/maxFeePerGas straight into TransactionInput, which matches the JSON exactly. The `trade-off`: PublicTxOptions is reused by PublicTx in the later RPC modules (#1239), so we may want a standalone type eventually. Happy to extract it now if reviewers prefer.

2. `data → JsonNode` as the Java equivalent of Go's RawJSON — flagging in case there's a preferred convention.

3. `On-chain receipt indices use boxed Long` (blockNumber/transactionIndex/logIndex) so "not finalized on-chain" (null/omitted) is distinguishable from a genuine 0.

4. `Zero Timestamp normalized to null` in the receipt constructor — the Timestamp deserializer yields a zero (not null) for an absent indexed field, so this keeps equality/round-trips clean and matches Go omitempty.

matthew1001 (commented):
Thanks @divyam-jha123 great to see the first lot of function dropping 👍 While I'm reviewing the code more closely, a couple of high-level things to look at that I think are worth including in this PR as the first drop of substantial Java code:

1. Can you look at incorporating code coverage checks into the build. I think we could look at JaCoCo or Sonar cube, but now would be a good time to have coverage reported, and to have the build fail when coverage is below a given threshold. Maybe we set the coverage threshold at whatever level the tests in the PR give us to start with.
2. With Java formatting being more subject to developer preference I quite like the idea of us adopting something like Spotless with Google's java format. I've found that helpful when working on projects like Besu to know that formatting decisions are firmly in place in the codebase, and the build fails when that formatting is broken.

divyam-jha123:
Thanks @matthew1001 ! I've added both gates to the Java SDK build in 
`2b89677` . Quick summary of what I did and the choices I made:

Code coverage — JaCoCo

I went with JaCoCo over SonarQube to start: it's a pure Gradle plugin with no external service or CI secrets, it generates HTML + XML reports, and it fails the build via `jacocoTestCoverageVerification` when coverage drops below the threshold. It also matches the coverage-gating pattern the repo already uses on the Go side (core/go, domains/noto, etc.). Sonar can be layered on top later if we want dashboards/trend tracking.

As suggested, I set the threshold to the level the current tests give us: `78% instruction coverage` (the core module's measured baseline of 78.1%, floored to the nearest whole percent for a small buffer). For reference the other counters are line 85.4% / method 82.9% / branch 61.6% — happy to switch to a different counter (e.g. line) or raise the bar as the suite grows; it's a one-line change.

Formatting — Spotless + Google Java Format

Adopted Spotless with standard Google Java Format (2-space) to match the `Besu-style` setup you mentioned, plus import ordering, unused-import removal, trailing-whitespace/newline normalization, and Apache-2.0 license-header enforcement. The build now fails on any formatting violation, and `./gradlew :sdk:java:spotlessApply` fixes them. This PR includes the one-time reformat of the existing files (the bulk of the diff), which is the point of locking it in now.

Wiring

Both hook into Gradle's check/build lifecycle in the shared sdk/java/build.gradle `subprojects {}` block, so they apply to all SDK modules and the existing CI (`./gradlew … build`) enforces them automatically — no workflow changes needed.

I verified the negative paths too: dropping below the coverage threshold, breaking formatting, and removing a license header each fail the build as expected. 87 tests still pass.

sdk/java/build.gradle
```diff
@@ -17,10 +27,50 @@ subprojects {
...
+                    // (instruction coverage 78.1%), floored to the nearest whole percent.
+                    counter = 'INSTRUCTION'
+                    minimum = 0.78
+                }
+            }
+        }
+    }
+
+    check.dependsOn jacocoTestCoverageVerification
+
```
matthew1001, line 64: Would you mind adding Javadoc linting so e.g. public Java methods are required to have params, return values etc. doc'd in their comments.

matthew1001 (approved):
@divyam-jha123 thanks for this. I think the structure is looking good. I've made one more request before merging, which is to add Javadoc linting to the build and add the necessary Javadoc to pass that check.

I've added an approval because if we really had to we could add that in a later PR - but I feel like it would be good to start with clean javadoc coverage.

Thanks for this PR 👍

sdk/java/build.gradle
divyam-jha123: done added: 
1. Added `Javadoc linting` to the build (`sdk/java/build.gradle`, in the subprojects {} block alongside Spotless/JaCoCo)
2. Added the Javadoc to pass it — filled all `259 doclint findings` across `21 core` files (getters, static factories, public constructors, and the Builder setters that were the biggest gap), matching the existing terse, Go-cross-referencing house style.

divyam-jha123:
image
@matthew1001 Everything is green — javadoc, spotless, and jacoco all pass .

I think this pr is good to merge

Issue #1247, opened 2026-06-18
Split domain context into two


annamcallister:
### What would you like to be added?

The domain context currently serves two purposes:

- A long lived domain context per sequencer used to stage state and nullifier writers ahead of flushing them to the database
- A short lived domain context per assemble/endorsement, used to import a coordinator snapshot and create an execution context for the assemble/endorse which matches the coordinators ahead of chain view of the domain instance

There is practically no overlap between these two functions, yet they are combined into a single construct. In some cases, the same fields within the struct serve a dual purpose.

We should break the domain context into two separate structs/interfaces, where each provides only the minimal set of capability needed for each purpose.

### Why is this needed?

The current code is difficult to understand, prone to bugs, and difficult to fix where a bug affects one of the uses, without having unintended effects on the other use.

annamcallister:
https://github.com/LFDT-Paladin/paladin/pull/1283

PR #1250, opened 2026-06-18
Fixes from load test

annamcallister:
* Don't use the logrus formatter as it compiles a new regex for every log statement. We are not even using the prefix capability. I've added our own version of the formatter we have been using to avoid making big changes to logging at this time.
* Call `ResetTransactions` on the domain context when a transaction is cleaned up, as even though domain context is not used from tracking coordinator locks anymore, they are still being stored there. #1247 is needed to resolve this properly as the code is fairly tangled right now and difficult to change
* Cancel the assembly context after every assemble- regardless of outcome
* Only export private state data where there isn't a corresponding spend lock
* Some test fixes to use mocks more consistently and stop setting max dispatch ahead to -1 when the test no longer starts the dispatch loop


common/go/build.gradle
```diff
@@ -14,13 +14,17 @@
...
+import org.apache.tools.ant.filters.LineContains
 
 ext {
     goFiles = fileTree(".") {
         include "pkg/**/*.go"
     }
     targetCoverage = 100
     maxCoverageBarGap = 1
+    coverageExcludedPackages = [
+        'pkg/log/formatter.go',
```
matthew1001, line 26: You think we ought to come back and add tests for the new formatter?

core/go/internal/sequencer/coordinator/grapher/grapher.go
```diff
@@ -440,7 +440,19 @@ func (g *grapher) ExportStatesAndLocks(ctx context.Context, node string) (Export
 	defer g.mu.RUnlock()
 
 	result := exportableStates{}
-	result.OutputState = g.stateVisibilityTracker.GetForNode(node)
+
+	// Exclude private state data for states that also have a spend lock — those states are
+	// already consumed ahead-of-chain by another transaction. Assemblers and incoming coordinators
+	// have no use for the private data since the state cannot be respent.
+	allStates := g.stateVisibilityTracker.GetForNode(node)
```
matthew1001, line 447: Think I agree with the logic going in here and not in `GetForNode` since it's really grapher behaviour that's choosing to exclude them

matthew1001 (approved):

common/go/build.gradle
annamcallister: If we ever change this file then yes, that should be the signal to add tests.

I think there is a high chance once we put more investment into perf testing that we actually want a completely new logging library.

PR #1254, opened 2026-06-22
Missing reset handlers in preassembly blocked

annamcallister:
Preassembly blocked was not handling dependencies being reset as it is already the state that transactions must transition to if their dependencies are reset.

However, a transaction may be in preassembly blocked for both preassembly and and chained dependencies, meaning that if the unassembled status of a chained dependency is not restored, it becomes possible for the transaction to transition to pooled, when its preassembly dependency is selected, and eventually get assembled ahead of its chained dependency. This can result in a circular dependency between transactions A and B, where B has A as a chained dependency, but B has A as a post assemble dependency. In this case neither transaction can be dispatched.


matthew1001 (approved):
LGTM

PR #1255, opened 2026-06-22
Test enhancements

annamcallister:
Reworks test runner to trigger the checks that were previously done at the end of the run in batches through out the test run. This puts a cap on how much needs to be stored in memory over the duration of the test run.

Prints the transaction ID at the end of each `RunOnce` action at INFO level


test/internal/perf/perf.go
```diff
@@ -787,8 +839,10 @@ func (pr *perfRunner) runLoop(tc testsuite.TestCase, workerID int, actionsPerLoo
 				actionCount := actionsCompleted
 				pendingActions++
 				go func() {
-					transactionID, err := tc.RunOnce(actionCount)
-					log.Debugf("%d --> %s action %d sent after %f seconds", workerID, testName, actionCount, time.Since(startTime).Seconds())
+				transactionID, err := tc.RunOnce(actionCount)
```
matthew1001, line 842: The formatting looks out here

test/internal/perf/perf.go
```diff
@@ -168,6 +172,11 @@ func New(config *conf.RunnerConfig, reportBuilder *util.Report) PerfRunner {
 	startTime := endRampTime
 	endTime := startTime + int64(config.Length.Seconds())
 
+	rollingCheckBatchSize := config.RollingCheckBatchSize
+	if rollingCheckBatchSize <= 0 {
```
matthew1001, line 176: It looks like we also default in `run.go`, can we drop one?

matthew1001 (approved):
LGTM, couple of minor comments

PR #1258, opened 2026-06-22
Reliable message test fix

annamcallister:
Messages are not guaranteed to arrive in order- test must tolerate this


matthew1001 (approved):
LGTM

PR #1260, opened 2026-06-23
Accept no revert data/retry failed finalize

### LINKED ISSUE #1257: bug: Transaction receipt not written to database
### What happened?

The sequencer state machines finalised a private transaction as failed, but the receipt was not written to the database. This caused the originator to resume the transaction every 5 mins, with new base ledger failures each time, but no receipt ever written.

The receipt to be persisted, meaning the transaction is finalised in the DB.

Run test `noto-pente-tracker` over an extended period of time.

_No response_

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
# paste output here
```

</details>

annamcallister:
An on chain revert may not have revert data. Previously the code would silently drop the finalize operation if this was the case. There is already code in place to handle no revert data being turned into an i18n message to explain this. I have also
- added an error message at the point where we currently silently drop finalize operations, should a future bug/regression cause us to drop other finalize operations
- moved a warn log down to debug level as it is not a system level warning
- removed an inaccurate comment about flushing domain context at finalize

Fixes https://github.com/LFDT-Paladin/paladin/issues/1257

---

Fix a bug where transaction finalization failures were not being consistently retried.

### DIFF
--- a/core/go/internal/sequencer/syncpoints/finalize.go
+++ b/core/go/internal/sequencer/syncpoints/finalize.go
@@ -19,10 +19,8 @@ package syncpoints
 import (
 	"context"
 
-	"github.com/LFDT-Paladin/paladin/common/go/pkg/i18n"
 	"github.com/LFDT-Paladin/paladin/common/go/pkg/log"
 	"github.com/LFDT-Paladin/paladin/core/internal/components"
-	"github.com/LFDT-Paladin/paladin/core/internal/msgs"
 	"github.com/LFDT-Paladin/paladin/core/pkg/persistence"
 	"github.com/LFDT-Paladin/paladin/sdk/go/pkg/pldapi"
 	"github.com/LFDT-Paladin/paladin/sdk/go/pkg/pldtypes"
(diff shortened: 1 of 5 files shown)


core/go/internal/sequencer/syncpoints/finalize.go
matthew1001, line 87: So just to confirm the updated behaviour, the UX is that if the EVM ledger doesn't include revert data you'll just get empty revert data in the receipts table, not a "PD*****" replacement message? We probably want to follow up with some docs updates to explain the importance of enabling it where it's available.

matthew1001 (approved):

core/go/internal/sequencer/syncpoints/finalize.go
annamcallister: no you'll get the replacement message- that's already implemented further down the stack

annamcallister: https://github.com/LFDT-Paladin/paladin/blob/main/core/go/internal/txmgr/persisted_receipt.go#L318

PR #1261, opened 2026-06-23
Snapshot improvements

annamcallister:
* Always include originator in snapshot - this was missing on confirmed. An originator removes a transaction from memory as soon as it is finalized, a coordinator keeps it in memory for a further number of heartbeat intervals. If the transaction is resumed in this time on the originator then processing the confirmed transaction in the heartbeat will cause the resumed transaction to complete also, preventing it from beign redelegated back to the coordinator once it has been removed from memory 
* Include reverted transactions in snapshots too, to allow the same fast forward logic
* Make use of multiple validators to stop storing a temporary will retry value on an originator transaction

### DIFF
--- a/core/go/internal/sequencer/common/types.go
+++ b/core/go/internal/sequencer/common/types.go
@@ -25,6 +25,7 @@ type CoordinatorSnapshot struct {
 	DispatchedTransactions []*SnapshotDispatchedTransaction `json:"dispatchedTransactions"`
 	PooledTransactions     []*SnapshotPooledTransaction     `json:"pooledTransactions"`
 	ConfirmedTransactions  []*SnapshotConfirmedTransaction  `json:"confirmedTransactions"`
+	RevertedTransactions   []*SnapshotRevertedTransaction   `json:"revertedTransactions"`
 	CoordinatorState       CoordinatorState                 `json:"coordinatorState"`
 	BlockHeight            uint64                           `json:"blockHeight"`
 	EndorserCandidates     []string                         `json:"endorserCandidates,omitempty"` // (COORDINATOR_ENDORSER selection mode only)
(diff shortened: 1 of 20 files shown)


core/go/internal/sequencer/coordinator/transaction/snapshot.go
```diff
@@ -57,27 +58,36 @@ func (t *coordinatorTransaction) GetSnapshot(ctx context.Context) (*common.Snaps
...
+				ID:         t.pt.ID,
+				Originator: t.originator,
+			},
+			RevertReason: t.revertReason,
+		}
 	}
 
-	// Reverted/final/other states are excluded from snapshots.
-	return nil, nil, nil
+	// Final/Evicted/Initial states are excluded from snapshots.
```
matthew1001: This feels like a good candidate to get out of date? Maybe just `All other states are excluded from snapshots."?

matthew1001 (changes requested):
In person discussion agreed that we shouldn't be informing all originators about other originators' TX IDs.

Also important to have sufficient comments in a few places to avoid someone inadvertently adding that back in thinking they've found an optimisation

core/go/internal/sequencer/common/types.go
```diff
@@ -39,8 +45,7 @@ type CoordinatorSnapshot struct {
 }
 
 type SnapshotPooledTransaction struct {
-	ID         uuid.UUID
-	Originator string
+	ID uuid.UUID
 }
```
matthew1001, line 49: 👍

doc-site/docs/architecture/distributed_sequencer_state_machine_transitions.md
```diff
@@ -231,43 +231,49 @@ stateDiagram-v2
     Initial --> Confirmed : ConfirmedSuccess
     Initial --> Pending : Created
     Pending --> Confirmed : ConfirmedSuccess
+    Pending --> Confirmed : ConfirmedReverted
```
matthew1001, line 234: ❤️

matthew1001 (approved):

Issue #1224, opened 2026-06-09
Enhancement:  RFC: Java SDK for Paladin


divyam-jha123:
# RFC: Java SDK for Paladin

**Type:** RFC / Architecture Discussion  
**Mentorship:** LFX 2026 — Java SDK Development and Release Pipeline  
**Mentors:** @matthew1001 @annamcallister  

---

## Summary

Add a first-class Java SDK to the Paladin repo (`sdk/java/`), published to Maven Central, with feature parity to the existing Go (`sdk/go`) and TypeScript (`sdk/typescript`) SDKs across all nine JSON-RPC namespaces. This RFC proposes the module layout, transport design, and the Java-specific API decisions, and lists the open questions needing maintainer input before implementation starts.

A working PoC validated end-to-end against a live Paladin devnet exists at https://github.com/divyam-jha123/paladin-java-sdk and informs the design below.

## Scope

Feature parity with the Go and TypeScript SDKs across: `ptx`, `keymgr`, `transport`, `registry`, `pstate`, `blockindex`, `pgroup`, `domain`, `debug`.

Deliverables:
- HTTP request/reply transport with configurable retry and timeout policies.
- WebSocket subscription client with full `subscribe` / `ack` / `nack` / `ping` / `reconnect-with-backoff` semantics, matching `PaladinWebSocketClientBase`.
- Fluent `TxBuilder` with deferred-error semantics and a `TransactionFuture` for receipt polling, mirroring the Go SDK.
- Domain helpers for Noto, Pente, and Zeto.
- A `paladin-sdk-testing` module: mock client, WireMock stubs, JUnit 5 Testcontainers extension.
- Maven Central artifacts plus a BOM for version alignment.
- Coverage gates: ≥ 80% line / ≥ 90% branch on core and client modules.

Out of scope: reimplementing domain logic (proof generation, endorsement, notarisation stays server-side); changing the Go/TS SDKs. A reactive (Project Reactor) bridge and an OkHttp adapter are proposed as *optional* modules, not core (see Open Questions).

## Module layout

Gradle multi-project under `sdk/java/`:

```
sdk/java/
├── paladin-sdk-bom/       # Bill of Materials — version alignment only
├── paladin-sdk-core/      # Data models, primitive types, QueryJSON DSL, exceptions
├── paladin-sdk-client/    # PaladinClient, HTTP transport, builders, WebSocket subscriptions
[...]

PR #1264, opened 2026-06-23
feat(java-sdk): HTTP JSON-RPC transport core

### LINKED ISSUE #1231: Enhancement: Java SDK - HTTP RPC transport

Implement the low-level HTTP JSON-RPC transport layer that all RPC module
clients will be built on top of.

### Reference implementations
- `sdk/go/pkg/rpcclient/` - RPC client core
- `sdk/go/pkg/retry/` -  retry policy
- `sdk/typescript/src/paladin.ts` - TS client

### Scope
- [x] `java.net.http`-based JSON-RPC client (no extra runtime dependency)
- [x] Configurable retry policy (max attempts, backoff)
- [x] Configurable timeout policy (connect, request)
- [x] All public methods return `CompletableFuture<T>`
- [x] `PaladinException` unchecked exception hierarchy
  - `PaladinRpcException` (non-2xx / JSON-RPC error response)
  - `PaladinTimeoutException`
  - `PaladinConnectionException`
- [x] Unit tests with a mock HTTP server (WireMock or `HttpServer`)

### Out of scope
WebSocket, RPC-namespace-specific methods, domain helpers.

### Acceptance criteria
- Client can make a JSON-RPC call and deserialize the response into a typed result
- Retry and timeout policies are exercised by unit tests
- No OkHttp or other third-party HTTP dependency in the default build

### Why is this needed?

Part of the Java SDK RFC #1224

divyam-jha123:
## What

Implements the low-level HTTP JSON-RPC transport layer for the Java SDK — the foundation every
RPC-namespace client (PTX, KeyManager, …) will be built on. Mirrors the Go reference
(`sdk/go/pkg/rpcclient`, `sdk/go/pkg/retry`) and the TypeScript client (`sdk/typescript/src/paladin.ts`),
using only `java.net.http` with no third-party HTTP dependency.

Enhancement #3 of the Java SDK RFC (#1224).

## How it's structured

The change is delivered as 5 small, independently-reviewable commits — each compiles and its tests
pass on their own:

- [x] 1. **PaladinException unchecked exception hierarchy** — base + `PaladinRpcException` /
      `PaladinTimeoutException` / `PaladinConnectionException`
- [x] 2. **JSON-RPC request/response envelope** — `JsonRpcRequest` / `JsonRpcResponse` /
      `JsonRpcError` + `JsonRpcErrorCode` constants
- [x] 3. **Retry & timeout policy config** — `RetryPolicy` (exponential backoff) + `RpcClientConfig`
      (URL, connect/request timeouts, headers)
- [x] 4. **`java.net.http` JSON-RPC client** — `RpcClient` interface (`CompletableFuture<T>`) +
      `HttpRpcClient` async implementation
- [x] 5. **Mock HTTP server unit tests** — `MockJsonRpcServer` (JDK `HttpServer`) + `HttpRpcClientTest`

## Design notes

- **No new runtime dependency.** Transport is pure `java.net.http`; the only added dependency is
  `jackson-databind` (already used by `:sdk:java:core`), pulled into `:sdk:java:client` for JSON binding.
- **Fully async.** Every public method returns `CompletableFuture<T>`; built on `HttpClient.sendAsync`.
  Failures surface as `PaladinException` subtypes (wrapped in `CompletionException` when the future is joined).
- **Retry semantics.** Connection failures, timeouts, and transient HTTP (5xx / 429) are retried with
  exponential backoff; JSON-RPC application errors and other 4xx are not.
- **Request IDs.** Monotonic `AtomicLong`, 9-digit zero-padded — mirrors Go's `%.9d`.

## Acceptance criteria (#1231)

- [x] `java.net.http`-based JSON-RPC client (no extra runtime dependency)
- [x] Configurable retry policy (max attempts, backoff)
- [x] Configurable timeout policy (connect, request)
- [x] All public methods return `CompletableFuture<T>`
- [x] `PaladinException` unchecked hierarchy (`PaladinRpcException` / `PaladinTimeoutException` /
      `PaladinConnectionException`)
- [x] Unit tests with a mock HTTP server (JDK `HttpServer`)
- [x] Client can make a JSON-RPC call and deserialize the response into a typed result
- [x] Retry and timeout policies are exercised by unit tests
- [x] No OkHttp or other third-party HTTP dependency in the default build

Out of scope (per the issue): WebSocket, RPC-namespace-specific methods, domain helpers.

## Testing

./gradlew :sdk:java:client:build

`HttpRpcClientTest` covers: typed-result deserialization, JSON-RPC error → `PaladinRpcException`,
non-2xx handling, retry-then-succeed, retry-exhausted, request timeout → `PaladinTimeoutException`,
and connection-refused → `PaladinConnectionException`.

Closes #1231


sdk/java/client/build.gradle
```diff
@@ -1,4 +1,5 @@
 // HTTP + WebSocket transport and RPC client
 dependencies {
     implementation project(':sdk:java:core')
+    implementation 'com.fasterxml.jackson.core:jackson-databind:2.18.6'
```
matthew1001, line 4: I notice that `2.18.9` is the latest `2.18.*` version. I think if we stick with `2.18` we should probably bump to the latest. Are there more recent version worth considering? It looks like `2.23` might be latest but I don't know what the main changes are

sdk/java/client/src/main/java/org/lfdt/paladin/sdk/client/rpc/JsonRpcErrorCode.java
```diff
@@ -0,0 +1,46 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.lfdt.paladin.sdk.client.rpc;
+
+/**
+ * JSON-RPC 2.0 error codes used by Paladin, mirroring Go's {@code rpcclient.RPCCode} constants.
```
matthew1001, line 18: I'd suggest we can probably drop the comments about "mirroring Go's implementation", just to keep things tidy

sdk/java/client/src/main/java/org/lfdt/paladin/sdk/client/rpc/JsonRpcRequest.java
```diff
@@ -0,0 +1,105 @@
...
+  /** The JSON-RPC protocol version string, always {@code "2.0"}. */
+  public static final String JSON_RPC_VERSION = "2.0";
+
+  private final String jsonrpc;
+  private final String id;
+  private final String method;
+  private final List<Object> params;
+
+  @JsonCreator
+  JsonRpcRequest(
```
matthew1001, line 43: For code cleanliness I'd like us to declare vars passed in to functions as `final` where they are not modified. Would you mind updating the build to use something like `Checkstyle` (I think its `FinalLocalVariable` rule is the one we want) so the build fails for non-reassigned local vars, and then fix up all the places where we need to add `final`?

sdk/java/client/src/main/java/org/lfdt/paladin/sdk/client/rpc/HttpRpcClient.java
```diff
@@ -0,0 +1,244 @@
...
+  private static Throwable unwrap(Throwable throwable) {
+    Throwable cause = throwable;
+    while ((cause instanceof CompletionException) && cause.getCause() != null) {
+      cause = cause.getCause();
+    }
+    return cause;
+  }
+
+  private static boolean isSuccess(int status) {
+    return status >= 200 && status < 300;
```
matthew1001, line 236: There were some recent changes to Paladin's behaviour in terms of JSON/RPC responses. The main change, which was to be compliant with the [JSON/RPC spec](https://www.jsonrpc.org/specification#error_object) was to return HTTP 200 for most error conditions. The PR is https://github.com/LFDT-Paladin/paladin/pull/1183.

This was a very recent change so not your fault since you probably don't have it in your branch. But we need to align the Java SDK with the changes. That PR ^^ has the relevant Go/Typescript SDK updates to guide the Java ones.

matthew1001 (commented):
Thanks @divyam-jha123 this is looking good. I've added a number of comments, one relating to a recent change in Paladin you probably wouldn't be aware of.

We'll definitely need per-client configurable mTLS configuration. It doesn't need to be included in this PR, but I just wanted to mention to make sure it's on the board as a future item/PR?

sdk/java/client/src/main/java/org/lfdt/paladin/sdk/client/rpc/JsonRpcErrorCode.java
divyam-jha123: ya sorry to mis that, I will take care of that in future prs

sdk/java/client/build.gradle
divyam-jha123: hey @matthew1001 , using the version 2.18.6 is not an independent pick for the SDK - it was chosen to match the rest of the Paladin repo.
Ex - In files `core/java/build.gradle`, `toolkit/java/build.gradle`, `domains/pente/build.gradle`, and here on this branch.

So the SDK deliberately aligned to the incumbent version rather than introducing a second Jackson version into the monorepo. That consistency is the reason — and it's also the constraint on any bump: ideally we should move all of them together, not just the SDK.

sdk/java/client/src/main/java/org/lfdt/paladin/sdk/client/rpc/JsonRpcRequest.java
divyam-jha123: Added a Checkstyle gate (`FinalLocalVariable` with `VARIABLE_DEF,PARAMETER_DEF`) in `subprojects {}` with `maxWarnings = 0`, so unmodified locals and params must be `final` or the build fails. Scoped the config to just that one rule so it doesn't overlap with Spotless. Fixed all violations across `core + client (main + test)`. GJF line-reflow from the added finals will land in the usual spotless-conform pass.

sdk/java/client/src/main/java/org/lfdt/paladin/sdk/client/rpc/HttpRpcClient.java
divyam-jha123: @matthew1001 , Turns out the Java HTTP transport was already written to the spec-compliant model: HttpRpcClient determines errors from the JSON-RPC error member before looking at the HTTP status, so a 200 carrying an error body is already surfaced as a PaladinRpcException and isn't retried — the same behaviour #1183 added to the TypeScript SDK (and Go already had). No production logic change was needed.

To align properly and lock it in, I've:
  - Added regression tests mirroring the #1001 scenario (idempotency clash PD012220 returned as HTTP 200) plus explicit 403 auth (not retried) and error-body-over-status cases.
  - Documented the 200/403/500 contract in the HttpRpcClient / JsonRpcErrorCode Javadoc.
 
One deliberate deferral: the new RPCCodeConflict = -32001 constant and the idempotency-key-clash recovery (look up the existing tx via getTransactionByIdempotencyKey, as Go's txbuilder.go does) belong with the TxBuilder work — there's no TxBuilder/PTX client in this transport PR — so I've left a TODO for it rather than adding a dangling constant here.

matthew1001: that's great, thanks for checking @divyam-jha123

sdk/java/client/build.gradle
matthew1001: that sounds reasonable to me, thanks for explaining

matthew1001 (approved):
Thanks @divyam-jha123 another great PR for the Java SDK 🙏

PR #1268, opened 2026-06-24
Take lock when writing or deleting pending private states

annamcallister:
It is currently possible for the DBTX that a domain event stream makes when checking for and writing pending private state data to interleave with a reliable message DBTX which is writing new private state data and deleting pending state data entries. This has the effect that neither is atomic and we can have orphaned private state data entries which will never be cleared and permanently block assembly of transactions for affected contracts.

It isn't possible to craft the SQL queries on either side to ensure each DB transaction is atomic, so this PR adds an advisory lock that enforces synchronisation between the two goroutines.


matthew1001 (approved):
This feels like a reasonable safety-first approach for the time being while we ensure the DB writes & deletes are safe. We can come back to re-assess if there's a more performant solution once we start pushing performance more

PR #1272, opened 2026-06-24
API changes required for UI features

gabriel-indik:
### Overview
This PR adds and extends query/filter APIs so the UI can search, filter, and sort Paladin data more flexibly. Most changes are new filter fields on existing query endpoints plus one new transport RPC.

### New API: `transport_peersWithQuery`
- New RPC endpoint that returns peers filtered/sorted via `QueryJSON`
- Supported filter fields: `name`, `activated`, `outboundTransport`, `outboundError`, and nested `stats.*` fields (createdAt, sent/received msgs/bytes, lastSend, lastReceive, reliableHighestSent).
- Exposed in:
  - Go SDK: `Transport.PeersWithQuery()`
  - TypeScript SDK: `transport.peersWithQuery()`
  - Docs: `doc-site/docs/reference/apis/transport.md`

### Extended query filters
| Area | Change |
|------|--------|
| **Group messages** | Added `node` filter |
| **Key manager** | Added `verifier` filter; `QueryKeys` joins `key_verifiers` |
| **Transport (reliable messages)** | Added `ack.time` filter |
| **Block indexer (transactions)** | Added `blockTimestamp` filter (via joined `Block.timestamp`) |
| **Uint256 fields** | `SupportsLIKE()` changed from `false` → `true` (enables partial/string matching on uint256 values) |

### Tests
- `TestQueryPeers` — peer filtering by `outboundTransport` and `name`
- RPC test for `PeersWithQuery`
- Reliable message query by `ack.time`
- Key query by `verifier`

### Other changes
- **Dockerfile**: Node.js version bumped from `20.17.0` to `20.19.0`
- **Copyright headers**: Updated from Kaleido to LFDT contributor format across touched files


gabriel-indik:
### Performance considerations
There are a number of changes within this PR that could potentially impact performance when it comes to running queries against the database. One of these changes is in `core/go/internal/keymanager/manager.go`:

```go
	q.Joins("LEFT OUTER JOIN key_mappings ON key_paths.path = key_mappings.identifier")
	q.Joins("LEFT OUTER JOIN key_verifiers ON key_paths.path = key_verifiers.identifier")
	q.Joins(`LEFT OUTER JOIN (SELECT parent AS "p" from key_paths AS p) AS k ON key_paths.path = k.p`)
	q.Where("key_paths.path != ''")
```

This change was added in order to make it possible to apply filters to the Keys panel:

image

Options that can be considered:
1. Remove the additional `JOIN` and also remove the `Filter by address` option in the UI
2. Refactor the code in the backend to make it more efficient.

gabriel-indik:
### Other performance considerations
In `core/go/internal/filters/field_uint256.go` support for `LIKE` was added:

```go
func (sf Uint256Field) SupportsLIKE() bool {
	return true
}
```

This makes it possible to apply filters where the user may not have the entire value, but for example knows the first or last n characters. In such cases there wouldn't be performance issues. However, if the user performs a search with `contains` where the expression would result in `LIKE %value%`, then the query would turn costly.

Options to consider:
1. Remove the "contains" JSON Query operator on this type.
2. Refactor the backend code to try and make the query as efficient as possible in this kind of cases.

gabriel-indik:
### Final performance considerations
There are a number of fields that had to be added to the filters to make it possible for the UI to filter/sort tables. Many of these fields are indexed, but some are not.

Possible options:
1. Remove the possibility to filter or sort by any field that is not currently indexed
2. Evaluate which fields we may want to filter/sort on and see if it is worth indexing them

PR #1276, opened 2026-06-25
API changes for UI

gabriel-indik:
The main change introduced in this PR is the modification to the JSON RPC API method `transport_peers`. It currently does not accept any arguments so there is no control of how many results it returns or which ones. In this PR the method is renamed to `transport_queryPeers` to follow the same naming convention used in other methods within Paladin. It now accepts one JSON Query argument.
Other changes in this PR: revert changes introduced in PR 1214. Specifically, in that PR `SupportsLIKE` was switched from `false` to `true` for certain fields in order to allow for queries where the user may know how a value starts, ends or what characters it may contain. This could have an impact on performance so the `SupportsLIKE` was restored to return `false`. Instead, the UI has been updated so that when applying a filter to a field that does not support `LIKE`, the comparisons `starts with`, `ends with`, `contains`, `does not contain`, `does not start with` and `does not end with` are removed.

### DIFF
--- a/core/go/internal/filters/field_bytes32.go
+++ b/core/go/internal/filters/field_bytes32.go
@@ -35,7 +35,7 @@ func (sf Bytes32Field) SQLColumn() string {
 }
 
 func (sf Bytes32Field) SupportsLIKE() bool {
-	return true
+	return false
 }
 
 func (sf Bytes32Field) SQLValue(ctx context.Context, jsonValue pldtypes.RawJSON) (driver.Value, error) {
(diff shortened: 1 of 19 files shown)


gabriel-indik:
This PR includes only backend changes, the UI changes that depend on them are in https://github.com/LFDT-Paladin/paladin/pull/1275

gabriel-indik:
This PR has been replaced by: https://github.com/LFDT-Paladin/paladin/pull/1277 and https://github.com/LFDT-Paladin/paladin/pull/1278

PR #1277, opened 2026-06-26
Restore `LIKE` restrictions for queries

gabriel-indik:
There is an explicit `LIKE` support flag for data types. Some of them restrict the ability to perform certain queries. In order to make it possible for the user to apply filters such as "Hash ends with xxx" or "ID starts with yyy", or "Address contains zzz" the `LIKE`support flag had to be changed, together with casting logic. While in principle this works fine, it could have a negative impact in performance. This PR restores the `LIKE` support flag for the data types for which it was changed. Note that as a consequence, queries like the ones listed here won't be possible.


matthew1001 (approved):
For now, given the discussions about possible impact on a production Paladin system of ineffcient DB queries, I think this is fine to go in.

We ought to bear in mind that we have a V1 release out now, so under normal circumstances we would want to issue a breaking-change notice for existing users to give them notice of a change like this. But since we are unlikely to have users already exploiting these wildcard options I think it's OK for this to just go into `main` as-is.

PR #1278, opened 2026-06-26
Add new JSON RPC method `transport_queryPeers`

gabriel-indik:
The existing API for listing peers is `transport_peers`. It does not accept any arguments. This presents a problem since the number of items returned can be arbitrarily large. This PR introduces a new method `transport_queryPeers` which takes as an argument a JSON Query. This makes sure that calls to the API have to explicitly indicate the limit for the number of entries returned. It also has the benefit that one can filter the result.

This PR introduces the new method alongside the old one, marking the old one as `deprecated`.


gabriel-indik:
Also included in this PR: introduced `transport_getPeer` while deprecating `transport_peerInfo`.

sdk/typescript/src/paladin.ts
```diff
@@ -1109,11 +1109,25 @@ export default class PaladinClient {
...
+    queryPeers: async (query: IQuery) => {
+      const res = await this.post<JsonRpcResult<any[]>>(
+        "transport_queryPeers",
+        [query]
+      );
+      return res.data.result;
+    },
+
+    /**
+   * @deprecated Use transport.queryPeers instead
```
matthew1001:
```suggestion
   * @deprecated Use transport.getPeer instead
```

core/go/internal/transportmgr/peer.go
```diff
@@ -111,6 +113,63 @@ func (tm *transportManager) listActivePeerInfo() []*pldapi.PeerInfo {
...
+	return p.vs
+}
+
+func peerInfoValueSet(p *pldapi.PeerInfo) filters.PassthroughValueSet {
+	return filters.PassthroughValueSet{
+		"name": p.Name,
+	}
+}
+
+func (tm *transportManager) QueryPeers(ctx context.Context, jq *query.QueryJSON) ([]*pldapi.PeerInfo, error) {
```
matthew1001: I don't think this needs exporting since it's not on the `TransportManager` interface. So can just be `queryPeers`.

matthew1001 (approved):
Approved with a couple of comments

core/go/internal/transportmgr/peer.go
gabriel-indik: Updated so it is not exported.

PR #1283, opened 2026-07-07
Split domain context into its separate read/write functions

annamcallister:
`DomainContext` is currently being used for two purposes:
- Staging private state writes ahead of being flushed to persistence
- Importing coordinator snapshots to give an ahead of chain view of states for assembly and endorsement

Since #1054 made the grapher the source of truth for tracking transaction dependencies and in memory state locks, there is no overlap between these two functions. The continue to shared key fields within the domain context, and terminology like `unflushed` is used inconsistently and inaccurately across the two functions. TX locks are still being accumulated in the domain context, and must be manually released, even though the domain context is no longer responsible for locks.

This PR separates the `DomainContext` interface and implementation into :
- `DomainStateWriter` : a buffer for staging private state data writes
- `DomainQueryContext`: A layer that sits in front of domain callbacks into `core`, holding information about the domain instance, allowing queries to be appropriately scope. A coordinator snapshot is an optional import where an in memory ahead of chain view of the domain instance needs to be layered on top of persistence.

The `ctx` and `info` parameters have been removed from the implementations. Context is passed in on every call in the interface, so that parent logging fields are passed through, allowing for clearer log correlation when debugging.

Making this split highlighted one place in code where there was a blurring between the two responsibilities. The long lived coordinator domain context used for staging state writes was also passed as the domain context in the `PrepareTransaction` call. This meant that the coordinators write buffer was available as a "snapshot", should a domain implementation use the `FindAvailableStates` callback back into `core` (the only callback which actually makes use of the snapshot). In reality a domain implementation is almost certainly never going to use this callback, since the transaction is already assembled at the point of dispatch, and all states known. However, I have left a TODO on what would be the technically correct way to populate this snapshot with the new implementation should it ever be needed.


core/go/internal/statemgr/domain_state_writer.go
```diff
@@ -0,0 +1,205 @@
+// Copyright © 2024 Kaleido, Inc.
```
matthew1001: Should probably use the new statements for new files

core/go/internal/statemgr/domain_state_writer.go
```diff
@@ -0,0 +1,205 @@
...
+		default:
+		}
+	}
+	if sw.unFlushed == nil {
+		sw.unFlushed = newPendingStateWrites(sw.ss)
+	}
+	return nil
+}
+
+func (sw *domainStateWriter) upsertStates(ctx context.Context, dbTX persistence.DBTX, holdingLock bool, stateUpserts ...*components.StateUpsert) ([]*pldapi.State, error) {
```
matthew1001, line 75: This is probably carry-over from the previous domain context but I've also thought that `upsert` is a term that really relates to persisting changes. Maybe that's an inaccurate/naive interpretation on my part, but given that we're only making in-memory changes here I wonder if we think about a different term.

core/go/internal/statemgr/domain_query_context.go
```diff
@@ -0,0 +1,426 @@
...
+		log.L(ctx).Tracef("mergeSnapshotStates: found %d matches", len(matches))
+		for _, m := range matches {
+			log.L(ctx).Tracef("Matched state: %s", m.ID)
+		}
+	}
+
+	return matches, nil
+}
+
+func (dqc *domainQueryContext) mergeInMemoryMatches(ctx context.Context, schema components.Schema, states []*pldapi.State, extras []*components.StateWithLabels, q *query.QueryJSON) (_ []*pldapi.State, err error) {
```
matthew1001: What do we merge from memory that isn't a snapshot? I think the fact I'm asking that shows that the term "in-memory" is a little unclear. I think of the query context as:

 - Take DB states I have
 - Overlay snapshot from coordinator
 - Use as the basis for state selection

We have `mergeSnapshotApplyLocks` which takes dbStates and merges snapshots into them. The last step (`mergeInMemoryMatches`) feels unclear what extra step is taking place. Maybe it's just unhelpfully named and should just be `merge()` as a helper function?

core/go/internal/statemgr/domain_query_context.go
```diff
@@ -0,0 +1,426 @@
...
+	ctx = log.WithLogField(ctx, "contract", contractAddress.String())
+	if schemaID != nil {
+		ctx = log.WithLogField(ctx, "schema", schemaID.String())
+	}
+	return ctx
+}
+
+// Short-lived, registered in the state manager. Always closed by the caller via defer dqc.Close(ctx).
+// May import a coordinator snapshot (creatingStates + txLocks) for FindAvailableStates queries.
+type domainQueryContext struct {
```
matthew1001, line 62: While we have the bonnet (hood?) open I do think it's worth having a naming discussion. The term "context" here hasn't always sat well with me. It's a valid term, but it's also a little overloaded in what it can mean. I use the term "overlay" a lot when we're discussing the way the coordinator provides the assembling node with optimistic states & locks. Maybe that term works well here?

I'm not going to stick firmly to "we need to change this", but now feels as good a time as any to discuss

matthew1001 (commented):
Thanks @annamcallister this really helps to clean the code up. I got a certain way through reviewing before needing to drop to other things, so. have added initial comments. Also once https://github.com/LFDT-Paladin/paladin/pull/1289#pullrequestreview-4673017592 merges the rest of the review should be a little easier

core/go/internal/statemgr/domain_query_context.go
annamcallister: Conclusion of in person discussion is full agreement with `context` being overloaded in golang but we don't have a good alternative right now.

core/go/internal/statemgr/domain_state_writer.go
annamcallister: renamed to `stageStateUpserts`

core/go/internal/statemgr/domain_query_context.go
annamcallister: I've done the following renames to make the function names more accurate

`mergeSnapshotStates` → `findSnapshotMatches`
`mergeInMemoryMatches` → `mergeAndSortStates`

core/go/internal/statemgr/domain_query_context.go
```diff
@@ -213,8 +213,8 @@ func (dqc *domainQueryContext) applyLocks(ctx context.Context, states []*pldapi.
 	return states
 }
 
-func (dqc *domainQueryContext) mergeSnapshotStates(ctx context.Context, schema components.Schema, dbStates []*pldapi.State, q *query.QueryJSON, excludeSpent, requireNullifier bool) (_ []*components.StateWithLabels, err error) {
-	matches := make([]*components.StateWithLabels, 0, len(dqc.creatingStates))
+func (dqc *domainQueryContext) findSnapshotMatches(ctx context.Context, schema components.Schema, dbStates []*pldapi.State, q *query.QueryJSON, excludeSpent, requireNullifier bool) (snapshotMatches []*components.StateWithLabels, err error) {
```
matthew1001, line 216: 👍

matthew1001 (approved):
Thanks Anna, agree with sticking with current terminology (although thanks for the state/merge/upsert changes which I think are much clearer

Issue #948, opened 2025-12-04
bug: Assess the use of both JSON and protobufs in the new sequencer


matthew1001:
### What happened?

This issue is intended to resolve comments in the new sequencer PR relating to the use of both JSON and protobuf within the sequencer code.

See comments:

 - https://github.com/LFDT-Paladin/paladin/pull/704#discussion_r2587065533
 - https://github.com/LFDT-Paladin/paladin/pull/704#discussion_r2587069826
 - https://github.com/LFDT-Paladin/paladin/pull/704#discussion_r2587085256

### What did you expect to happen?

N/A

### How can we reproduce it (as minimally and precisely as possible)?

N/A

### Anything else we need to know?

N/A

### OS version

N/A

annamcallister:
Fixed by #1284

PR #1284, opened 2026-07-07
Replace json parsing with more consistent use of protos

annamcallister:
Marshalling and unmarshalling JSON is substantially more expensive than marshalling and unmarshalling protos.

This PR addresses 3 places where we were marshalling JSON to include in a proto struct
- coordinator snapshots
- transaction preassembly
- transaction postassembly

Ensuring that the structs we send over the wire for these types are pure proto has driven a set of changes to make sure that we only include a minimal set of information in each message. Where fields are used to store values as the transaction progresses through its lifecycle but do not need to be sent to another node, they do not need to be defined in the proto; they instead sit alongside the the proto.

The transaction preassembly used to include details of block height, but these were actually captured at assembly time. I have moved this data to the transaction post assembly.

I have aimed to minimise transformation in and out of the proto where possible, and to refactor code such that the proto definitions aren't just bolted onto the edge at the transport interface. This lead to code analysis which show we had multiple type definitions representing the same data. This was especially true for states and locks. Some transformation is necessary because the type definitions we use for GORM and for the public API necessitate the use of types. I have aimed to keep this transformation to a minimum, and to remove unnecessary intermediate representations. This has, however, resulted in a lot of code changes as types have changed throughout many files. 

After making these changes, json marshalling and unmarshalling go from dominating CPU usage when running under moderate load, to not featuring in the top 80 nodes.

I've folded in one of the changes requested in https://github.com/LFDT-Paladin/paladin/issues/948 in this PR. The transport writer now accepts the full proto struct rather than accepting a flat list of parameters and building the proto struct itself. I have not addressed the other two comments.

- A golang list of message types rather than a proto enum

This message type is a string across the transport/messaging packages, and the sequencer is not the only package which provides values for it. Each message has a component value, which is a proto enum, for top level routing, and then a string message type which is specific to the component. This message type could be made to be an int32 so it could be used with proto enums but there is limited value when the value is entirely internal to one component communicating with other instances of itself. I.e. it is not an true external interface. Changing it would impact on identity resolver and reliable messages, with reliable messages in particular using the string as a value that gets persisted in its database.

- Protos being defined inside the sequencer and not in the prototk package

These protos are just ways that sequencers interact with each other. The protos in the toolkit represent component boundaries, specifically those which are pluggable. The PR comment referenced in the issue alluded to the sequencer moving in that direction, but in that case we would be defining the sequencer/transactionmanager interface as new protos. This proto interface is internal to this implementation of the sequencer.


domains/pente/src/main/java/io/kaleido/paladin/pente/domain/PenteDomain.java
```diff
@@ -222,7 +222,7 @@ protected CompletableFuture<InitContractResponse> initContract(InitContractReque
      @Override
      protected CompletableFuture<InitTransactionResponse> initTransaction(InitTransactionRequest request) {
          try {
-             var tx = new PenteTransaction(this, request.getTransaction());
+             var tx = new PenteTransaction(this, request.getTransaction(), null);
```
matthew1001: Could we switch to a Java `Optional` here?

core/go/pkg/proto/engine.proto
```diff
@@ -203,5 +214,43 @@ message PreDispatchRejection {
...
+// top-level state machine (common.CoordinatorState) — cast on each side rather than mirroring the enum.
+message CoordinatorSnapshot {
+    repeated SnapshotDispatchedTransaction dispatched_transactions = 1;
+    repeated SnapshotPooledTransaction pooled_transactions = 2;
+    repeated SnapshotConfirmedTransaction confirmed_transactions = 3;
+    repeated SnapshotRevertedTransaction reverted_transactions = 4;
+    int32 coordinator_state = 5;
+    uint64 block_height = 6;
+    repeated string endorser_candidates = 7;
+    io.kaleido.paladin.toolkit.StateSnapshot state_snapshot = 8; // handover only
```
matthew1001:
```suggestion
    io.kaleido.paladin.toolkit.StateSnapshot state_snapshot = 8; // for coordinator handover only
```

domains/pente/src/main/java/io/kaleido/paladin/pente/domain/PenteTransaction.java
```diff
@@ -110,13 +110,13 @@ static class ABIDefinitions {
      private final long baseBlockTimestamp;
      private Values values;
  
-     PenteTransaction(PenteDomain domain, TransactionSpecification tx) throws IOException, IllegalArgumentException {
+     PenteTransaction(PenteDomain domain, TransactionSpecification tx, BlockContext blockCtx) throws IOException, IllegalArgumentException {
```
matthew1001, line 117: See other comment about `Optional<BlockContext>` rather than just `BlockContext`

toolkit/proto/protos/private_tx.proto
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright © 2026 Kaleido, Inc.
```
matthew1001: nit: copyright wording

toolkit/proto/protos/to_domain.proto
```diff
@@ -354,9 +362,7 @@ message TransactionSpecification {
   string function_signature = 4; // The calculated signature of the function ABI the caller is attempting to invoke
   string function_abi_json = 5; // The function ABI (asserted by the caller) they wish to inovoke on the smart contract
   string function_params_json = 6; // The parameters supplied in the invocation in normalized JSON format, after validation against the ABI the caller supplied
-  int64 base_block = 7; // The block number on the base ledger, upon which this transaction is being assembled
-  Intent intent = 8; // The sender's intended outcome (a submitted transaction or only a call)
-  int64 base_block_timestamp = 9; // The timestamp for the provided block number on the base ledger, upon which this transaction is being assembled
+  Intent intent = 7; // The sender's intended outcome (a submitted transaction or only a call)
```
matthew1001, line 365: We should probably have the discussion about versioning protobuf, and possibly asserting some rules about changing protobuf field numbers since we have new domain implementations being developed. Is there a reason not to make this one of the first PRs we don't break protobuf backwards compatibility? If not, then we should probably set ourselves a deadline (i.e. V1.1 and beyond will commit to no breaking changes)

core/go/internal/statemgr/domain_query_context.go
```diff
@@ -138,79 +139,66 @@ func (dqc *domainQueryContext) Close(ctx context.Context) {
...
+		if l.GetType() == prototk.SnapshotStateLock_CREATE {
+			stateID := l.GetStateId()
+			if state, found := processedStates[stateID]; found {
 				dqc.creatingStates[state.ID.String()] = state
-			} else {
-				// The state distribution message may not have arrived yet, or we are not on the list.
-				// The domain can Park the transaction if it suspects the former.
-				log.L(ctx).Infof("ImportSnapshot: state %s not found in snapshot states", l.State)
 			}
+			// A snapshot can contain create locks for states which already have a corresponding
```
matthew1001, line 170: 👍

core/go/internal/sequencer/transport/transport_writer.go
```diff
@@ -47,26 +38,26 @@ var (
 type TransportWriter interface {
```
matthew1001, line 38: Love how clean this file becomes after these changes

core/go/internal/sequencer/transport_client.go
```diff
@@ -308,6 +286,7 @@ func (sMgr *sequencerManager) handleCoordinatorHeartbeatNotification(ctx context
 	heartbeatEvent.FromNode = from
 	heartbeatEvent.ContractAddress = contractAddress
 	heartbeatEvent.CoordinatorSnapshot = coordinatorSnapshot
+	heartbeatEvent.StateSnapshot = heartbeatNotification.GetCoordinatorSnapshot().GetStateSnapshot()
```
matthew1001: There's something a little unclear/asymmetrical about the way that in protobuf `StateSnapshot` is a child of `CoordinatorSnapshot`, but in the event fields we extract `StateSnapshot` from `CoordinatorSnapshot` and set it as a peer, `heartbeatEvent.StateSnapshot`. Could we just pass around `CoordinatorSnapshot` in the heartbeat event to mirror the protobuf structure and pull it out when needed? The comment mainly stems from the fact that there's a lot of "do something with the coordinator snapshot" going on in this function and it takes a while to work out why they are all needed

matthew1001 (commented):
Definitely cleans things up. A few comments but generally looks really good

core/go/internal/sequencer/transport_client.go
annamcallister: Changed so that the state snapshot stays embedded in the coordinator snapshot after processing- but as the proto struct.

toolkit/proto/protos/to_domain.proto
annamcallister: Discussed in person- added section to contributing doc page, but also noted this exception

toolkit/proto/protos/private_tx.proto
annamcallister: fixed

domains/pente/src/main/java/io/kaleido/paladin/pente/domain/PenteTransaction.java
annamcallister: Fixed with an overloaded constructor

matthew1001 (approved):

PR #1287, opened 2026-07-09
Release lock when doing DB work

annamcallister:
Holding the lock while persisting the dispatch blocks the main coordinator event loop if it is trying to update inflight transactions. This can result in critical actions like selecting for assembly being blocked.


matthew1001 (approved):
This looks to be safe, given we relock immediately after handling the event, and given that handling the event doesn't touch dispatch-related structures

PR #1288, opened 2026-07-09
More debug server options

annamcallister:
Allow more types of profile to be collected from the debug server- off by default


matthew1001 (approved):
LGTM

PR #1289, opened 2026-07-09
Migrate to zap logger

annamcallister:
Move away from logrus which has two key problems
* forces all logs through a single mutex
* uses excessive memory allocation formatting in Logf functions

I considered migrating to `slog` which is the go provided successor to `logrus` but it does not support formatting parameters and building them on top was inefficient. `zap` has a sugared formatter which supports formatting without allocating a new buffer every time.

I have attempted to keep the migration impact to a minimum across the code base, changing consumers only where they were directly interacting with `logrus` . The custom encoder for `zap` ensures that our logging format remains consistent.

Even with this change debug logging continues to carry a high CPU cost and the reality is that most of the time Paladin needs to be ran with debug logs to be able to debug any potential problems that might arise. I have added the ability to buffer logs before writing to help with this where there is a need to balance performance against observability (with the trade off that buffered logs will be lost on container crash) . I think there is also an exercise to critically review logs on hot paths, either removing the log, or trying to avoid logs which use formatted parameters.


common/go/pkg/log/log.go
```diff
@@ -26,12 +27,27 @@ import (
...
+// LevelTrace adds a level below zap's Debug(-1) to carry logrus's Trace level. The
+// code base uses Tracef/Trace but zap has no Trace level of its own; every other
+// level maps one-to-one onto zap (Debug/Info/Warn/Error, and the terminal Panic/Fatal).
+const LevelTrace = zapcore.Level(-2) // just below Debug(-1)
+
 var (
-	rootLogger = logrus.NewEntry(logrus.StandardLogger())
+	// atomLevel is the single global level control, shared by every core so
+	// SetLevel takes effect immediately across all loggers (matching logrus).
+	atomLevel = zap.NewAtomicLevelAt(zapcore.InfoLevel)
```
matthew1001, line 48: Does the caller need to know it's an atom, or could we just call it `globalLevel`?

common/go/pkg/pldmsgs/en_descriptions.go
```diff
@@ -406,6 +406,12 @@ var (
 	LogConfigUTC          = pdm("LogConfig.utc", "Sets log timestamps to the UTC timezone")
 	LogConfigFile         = pdm("LogConfig.file", "Configure file based logging")
 	LogConfigJSON         = pdm("LogConfig.json", "Configure json based logging")
+	LogConfigBuffer       = pdm("LogConfig.buffer", "Configure buffered log output")
+
+	// LogBufferConfig field descriptions
+	LogBufferConfigEnabled       = pdm("LogBufferConfig.enabled", "Enables buffered log output, batching lines in memory to reduce the number of write syscalls (default false)")
```
matthew1001, line 412: Is there a reason not to default `true`?

matthew1001 (approved):
LGTM, couple of small comments

common/go/pkg/log/log.go
annamcallister: It's not actually exposed to the caller outside this package.

common/go/pkg/pldmsgs/en_descriptions.go
annamcallister: yes - there's a risk of crucial losing logs at the point of a crash if this is enabled. It's a useful option to have in some situations- for example early stage performance work where the debug logs are giving us key information, but can really hammer CPU - but I think it should always be a choice to accept this trade off.

PR #1292, opened 2026-07-14
Avoid anti join when polling for pending transactions

annamcallister:
The public transaction manager has been polling for transactions by doing a join over the entire transactions table to find the transactions that do not have a completion. This gets more and more costly as the system ages and there isn't an index on the existing tables that can be added to speed up an anti-join.

This PR adds a new completed column plus a partial index on it to the transactions table that can be used instead of the join. There is a small cost to the DB of maintaining the partial index over just the pending transactions, but as it stays small compared to the total number of completed transactions it is a lot more efficient than the previous query.

It also adds a script that does `go mod download` and `go mod tidy` on every go package in the repo plus the changes that came from running all these commands and fully tidying up the `go.mod` and `go.sum` files.


matthew1001 (approved):
LGTM

PR #1293, opened 2026-07-14
Public tx manager lock contention

annamcallister:
This PR makes two changes

* limits the time that the inflight tx lock is held by having the work to update it operate on a copy and only hold the lock to update the slice point when the work is finished (the copy is a copy of a slice of pointers which keeps the size of the copy low). This massively reduces how often the goroutine handling transaction completions is blocked waiting to read this slice. The analysis from Claude on why this design is safe is included below.
* adds a lock to the the orchestrator state and state entry time which was previously unguarded but accessed from different goroutines

---

# Public-tx poll lock: design and safety

This documents the concurrency design of the per-signing-address orchestrator's poll cycle
(`pollAndProcess` in `core/go/internal/publictxmgr/transaction_orchestrator.go`) and why it is
safe for confirmations to run concurrently with a poll.

## The players

Three goroutines touch an orchestrator's in-flight transaction set:

- **The poll loop** (`orchestratorLoop` → `pollAndProcess`). One per signing address. It is the
  sole writer of `oc.inFlightTxs`: it decides which transactions to keep, reads the database for
  new ones, and processes each transaction.
- **The confirm path** (`NotifyConfirmPersisted` → `dispatchAction`). Runs on the block-indexer's
  goroutine when a transaction is confirmed on-chain. It finds the matching in-flight transaction
  by nonce and records the confirmation against it.
- **The engine loop** (`pubTxManager.poll` → `flushStaleOrchestratorsGetCount`). Reads each
  orchestrator's `state`/`stateEntryTime` to decide lifecycle actions.

Two locks are relevant:

- `oc.inFlightTxsMux` — guards the `oc.inFlightTxs` slice header.
- `it.transactionMux` — per in-flight transaction; guards that transaction's own mutable state
  (including its pending `newStatus`).

## The design: build a private list, publish it once

`pollAndProcess` builds its next in-flight list in a **local** variable, `newInFlight`, and never
mutates the shared `oc.inFlightTxs` while building it:

1. Read `oc.inFlightTxs` (the poll loop is its only writer, so this needs no lock) and copy the
   survivors — transactions not yet removable — into `newInFlight`.
2. Read the database for new transactions to fill spare slots, allocate their nonces, and append
   the resulting controllers to `newInFlight`.
3. **Publish** in a single locked assignment — the only hot-path critical section:
   ```go
   oc.inFlightTxsMux.Lock()
   oc.inFlightTxs = newInFlight
   oc.inFlightTxsMux.Unlock()
   ```
4. Process the transactions by passing `newInFlight` to `ProcessInFlightTransactions`, which runs
   outside the lock.

The lock is therefore held only for one slice-header assignment, not across the database read or
the processing.

## Why concurrent confirmations are safe

The confirm path (`oc.dispatchAction`) takes `oc.inFlightTxsMux`, scans `oc.inFlightTxs` for the
matching nonce, and calls `NotifyStatusUpdate` on the transaction it finds. Safety rests on two
independent layers.

### Layer 1 — the list is published atomically

`pollAndProcess` mutates only its local `newInFlight`; the shared `oc.inFlightTxs` changes exactly
once, under the lock, via a single header assignment. The confirm path reads `oc.inFlightTxs` only
while holding that same lock. Therefore a confirmation always observes a **complete** list — either
the entire previous list or the entire new one — and never a partially rebuilt one. The publish and
the read cannot interleave, because both hold `oc.inFlightTxsMux`.

### Layer 2 — the transactions are shared and individually locked

The list holds **pointers** to `inFlightTransactionStageController`. A transaction that survives a
poll cycle is the *same object* in both the previous and the new list, so it does not matter which
version of the list a confirmation reaches it through.

Mutations to a transaction are serialized by its own `transactionMux`:

- The confirm path sets the pending status under `transactionMux` (`NotifyStatusUpdate`).
- The processing path reads and applies that pending status under `transactionMux`
  (`ProduceLatestInFlightStageContext`).

Because processing acquires `transactionMux` per transaction, it is mutually exclusive with a
confirmation for that same transaction even though it runs outside `oc.inFlightTxsMux`.

### The edge case — a removable transaction dropped from the new list

A transaction is copied into `newInFlight` only while it is not yet removable; once terminal
(`CanBeRemoved`) it is left out. A confirmation for such a transaction is harmless either way:

- Arriving **before** the publish, it finds the transaction in the current list and records the
  status. The update is idempotent — an `IsReadyToExit` guard makes a late update a no-op.
- Arriving **after** the publish, it does not find the transaction and does nothing.

Either outcome is correct, because a transaction is dropped only once it is already terminal, so a
confirmation arriving afterwards is redundant. No meaningful confirmation is lost.

## Scaling

The per-cycle copy is a slice of 8-byte pointers, not of transactions. At ~500 in-flight
transactions the list is ~4 KB, built with O(n) pointer appends outside the lock — negligible next
to the database read and processing it accompanies. The publish is O(1): a single header
assignment. The lock-hold time is therefore constant and independent of the number of in-flight
transactions.

## Measured impact

Captured from profiling run `profiles-20260713-132229` (~145s steady window), compared against the
pre-change baseline `profiles-20260709-165219`. Both are mutex-contention diffs (`mutex_A` →
`mutex_B`) over the same workload.

### Lock contention removed

`inFlightTxsMux` is no longer a meaningful contention point:

| contended stack | before | after |
|---|---|---|
| `orchestratorLoop → pollAndProcess` | 8.09s (26.6%) | 0.46s (2.2%) — **−94%** |
| `NotifyConfirmPersisted → dispatchAction` | 3.12s (10.3%) | below the 0.11s reporting threshold — **negligible** |
| total contended time (all locks) | 35.93s | 21.07s — **−41%** |

Holding the lock only for the one-line publish, instead of across the DB read and the whole
processing pass, is what removes the poll path from the profile. The confirm path benefits as a
knock-on: `dispatchAction` no longer queues behind a poll cycle holding the lock, so its own
contention falls away even though its code is unchanged.


matthew1001 (approved):
The changes look safe, and reduce the previous coarse-grained locking of the whole inFlightTxs mutex for the duration of the poll by using locks just to set and get states.

One minor nit about `setState()` which doesn't follow the `defer(unlock)` pattern of the other new functions. But not a safety issue, just a code convention comment.

PR #1294, opened 2026-07-14
Batch persist dispatches

annamcallister:
Breaks dispatch into two phases:
- Preparing the transaction 
- Persisting the dispatch 

This allows for two performance improvements
* Prevents the transaction from being locked for the full time that the DB write is taking place, causing lock contention when the main coordinator event loop needs to do a locked read on that transaction.
* Allows transactions to be persisted in batches, stopping DB IO from becoming a limiting factor for dispatch throughput.

Also includes a change to make transactions track their own addition and removal from the inflight tx count, removing the need to call `getTransactionsInStates` across all transactions to recompute the full list every time a transaction leaves dispatched state.

I've left a TODO about handling dispatched events in parallel, since it is possible but with batching just persistence, dispatch throughput stops being a bottleneck within the sequencer.


common/go/pkg/pldmsgs/en_descriptions.go
```diff
@@ -748,6 +748,7 @@ var (
 	SequencerConfigInactiveGracePeriod               = pdm("SequencerConfig.inactiveGracePeriod", "Number of heartbeat intervals without activity before a node is considered inactive")
 	SequencerConfigMaxInflightTransactions           = pdm("SequencerConfig.maxInflightTransactions", "Maximum number of inflight transactions")
 	SequencerConfigMaxDispatchAhead                  = pdm("SequencerConfig.maxDispatchAhead", "Maximum number of transactions to dispatch ahead")
+	SequencerConfigMaxDispatchBatchSize              = pdm("SequencerConfig.maxDispatchBatchSize", "Maximum number of transactions prepared and committed in a single dispatch batch")
```
matthew1001: Should there be an accompanying dispatch batch timeout config option?

core/go/internal/sequencer/coordinator/transaction/dispatch.go
```diff
@@ -25,20 +25,25 @@ import (
...
 	"github.com/LFDT-Paladin/paladin/sdk/go/pkg/pldtypes"
 	"github.com/LFDT-Paladin/paladin/toolkit/pkg/prototk"
 	"github.com/google/uuid"
 )
 
-// action_Dispatch runs the full dispatch flow when handling Event_Dispatched in State_Ready_For_Dispatch.
-func action_Dispatch(ctx context.Context, t *coordinatorTransaction, _ common.Event) error {
-	return t.dispatch(ctx)
+// action_DispatchPrepare handles Event_Dispatched in State_Ready_For_Dispatch by building the transaction's
+// dispatch, which the dispatch loop then detaches and persists off-lock.
```
matthew1001: I could do with some extra clarity in the comment around what "detaching and persisting off-lock" means. I guess we are getting ready to dispatch but won't persist anything here, and will then persist the dispatch within the new batch mechanism? The comment below for `dispatchPrepare` is clearer - so maybe it's just the phrasing here I think could be a little clearer

core/go/internal/sequencer/coordinator/transaction/dispatched.go
```diff
@@ -33,6 +34,35 @@ func action_NotifyDispatched(ctx context.Context, t *coordinatorTransaction, _ c
 	return t.transportWriter.SendDispatched(ctx, t.originatorNode, msg)
 }
 
+// guard_HasDispatchedPublicTransaction reports whether dispatching this transaction will send a public
```
matthew1001: I misread `guard_HasDispatchedPublicTransaction` at first - it reads like a past-tense check ("has dispatched"). From the comment it's really `guard_WillDispatch...` but I know why you're using `Has` as the term for the guard. Maybe `guard_HasADisapatchedPublicTransction` or something else that reads more accurately as to what it's guarding?

core/go/internal/sequencer/coordinator/transaction/dispatched.go
```diff
@@ -33,6 +34,35 @@ func action_NotifyDispatched(ctx context.Context, t *coordinatorTransaction, _ c
...
+// transaction (rather than producing new private/prepared transactions), gating whether it counts
+// towards the coordinator's dispatch-ahead limit.
+func guard_HasDispatchedPublicTransaction(_ context.Context, t *coordinatorTransaction) bool {
+	return t.pt.PreparedPublicTransaction != nil &&
+		t.pt.PreAssembly.TransactionSpecification.Intent == prototk.TransactionSpecification_SEND_TRANSACTION
```
matthew1001, line 42: Do you think we should consider renaming these? Not necessarily in this PR, but reading this logic feels weird how `SEND_TRANSACTION` means "has a public transaction".

core/go/internal/sequencer/coordinator/dispatch_loop.go
```diff
@@ -21,75 +21,151 @@ import (
...
-			log.L(ctx).Debugf("submitting transaction %s for dispatch", tx.GetID().String())
-			err := tx.HandleEvent(ctx, &transaction.DispatchedEvent{
-				BaseCoordinatorEvent: transaction.BaseCoordinatorEvent{
-					TransactionID: tx.GetID(),
-				},
-			})
-			if err != nil {
-				log.L(ctx).Errorf("error dispatching transaction %s: %v", tx.GetID().String(), err)
-				continue
+			if capacity > c.maxDispatchBatchSize {
```
matthew1001: Yeah this feels like there should be a timeout as per my previous comment, where we dispatch regardless

config/pkg/pldconf/sequencermgr.go
```diff
@@ -81,6 +83,7 @@ var SequencerDefaults = SequencerConfig{
 	IdleSequencerCleanupInterval:      confutil.P("1m"),
 	InactiveGracePeriod:               confutil.P(2),
 	MaxDispatchAhead:                  confutil.P(50),
+	MaxDispatchBatchSize:              confutil.P(20),
```
matthew1001: Maybe bump this to 100?

common/go/pkg/pldmsgs/en_descriptions.go
annamcallister: Renamed to dispatchMaxBatchSize following discussion to make it clearer what this controls. The dispatch loop runs continuously to maximise throughput against the maximum number of transactions we are able dispatch optimistically to the chain without confirming dependencies (`maxDispatchAhead`). This config variable caps the maximum number of transactions that the loop will persist in a single batch.

I intentionally haven't included a timeout variable at this stage, preferring to process and dispatch available work as soon as it becomes available. We could add this in later when we've reached a more stable max TPS and measure whether it has an impact.

core/go/internal/sequencer/coordinator/transaction/dispatch.go
annamcallister: comment updated

core/go/internal/sequencer/coordinator/transaction/dispatched.go
annamcallister: Changed to `guard_WillDispatchPublicTransaction`

core/go/internal/sequencer/coordinator/transaction/dispatched.go
annamcallister: `SEND_TRANSACTION` applies to both public and private - the intent being to ultimately send a transaction to the base ledger.

matthew1001: Thanks for the reminder - so SEND is basically "will at some point produce something for the base ledger"

matthew1001 (approved):
Thanks for the in-person chat & for addressing the comments, LGTM

PR #1295, opened 2026-07-14
Track assembling transaction

annamcallister:
Avoids a getTransactionInStates call on every state transition, which required the transaction lock to be taken sequentially for every transaction being coordinated.

Aggravated by #1294 which was holding a transaction lock for far longer than necessary


core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -777,9 +777,18 @@ var stateDefinitionsMap = StateDefinitions{
 					Actions: []ActionRule{{If: guard_HasTransactionAssembling, Action: action_cancelCurrentlyAssemblingTransaction}},
 				}, {
 					Validator: validator_TransactionStateTransitionTo(transaction.State_Pooled),
-					Actions:   []ActionRule{{Action: action_PoolTransaction}},
+					Actions: []ActionRule{
+						{Action: action_PoolTransaction},
+						{If: statemachine.GuardNot(guard_HasTransactionAssembling), Action: action_SelectTransaction},
```
matthew1001: Just checking for my understanding, was this just condensing 2 actions into one action rule, where previously there were 2 actions that both ran? I.e. there's no actual logic change here?

core/go/internal/sequencer/coordinator/state_machine.go
```diff
@@ -777,9 +777,18 @@ var stateDefinitionsMap = StateDefinitions{
...
+					},
 				}, {
-					Actions: []ActionRule{{If: statemachine.GuardNot(guard_HasTransactionAssembling), Action: action_SelectTransaction}},
+					// The single assembly slot has been freed; clear the flag and select the next pooled transaction.
+					// Kept after the Pooled handler so the repool case (Assembling->Pooled) selects exactly once.
+					Validator: validator_TransactionStateTransitionFrom(transaction.State_Assembling),
```
matthew1001, line 782: Actually on reading this change I think I'm keen to understand in more detail. So previously we had one action that caused us to select transaction. Now it looks like we have 2:
 - Transition to pooled (select transaction)
 - Transition from assembling (select transaction)
Are we going to do 2 select transaction calls if this state change is **from** assembling **to** pooled? Or is that path not a valid path for a transaction?

matthew1001 (commented):
Agree with the change, had comments to understand the state machine changes fully.

core/go/internal/sequencer/coordinator/state_machine.go
annamcallister: Previously on every state transition we ran a handler with no validators, which checked to see if we had an assembling transaction and if not selected one for assembly.

This PR makes two changes to this logic
1. stops checking the state of every single transaction in order to determine whether we have a transaction assembling and instead tracks with a single boolean - this makes the big performance difference
1. applies validators for the cases where we might select for assembly to avoid redundant checks- this probably makes very little different given the first change

Assembling to pooled is tested to not result in two selects; however, I agree that its really not obvious from the code. I've reverted the second part of the change to return to doing a guarded select on every state transition, which should be 
a. cheap- it's just a boolean check now
b. more defensive against future changes

matthew1001: thanks for the explanation and tweaks, sounds good

matthew1001 (approved):

PR #1296, opened 2026-07-14
Stop using fmt to stringify hex

annamcallister:
The old functions resulted in ~5-6 allocs per call, which is driving a lot of GC


matthew1001 (approved):
LGTM

PR #1297, opened 2026-07-14
Format signer nonce once

annamcallister:
Formatting this string in a log line which is hit repeatedly in public tx processing was driving high allocations and GC. There's a bigger piece of work to remove our whole code base for this, but this picks off a log line that is individually driving a lot of allocations.


matthew1001 (approved):
LGTM

PR #1298, opened 2026-07-14
Move verifier resolution and signing into their own states

annamcallister:
Only one transaction is allowed to be in assembling state at a time. This means we want to aim to reduce the amount of work that takes place in that state, to increase throughput of transaction assembly.

Assembly currently begins with verifier resolution. This can be very quick if the verifiers are already in the cache, but if not it involves DB reads. Verifiers can be resolved once per tx, and survive being repooled for repeat assembly. Verifier resolution is now a separate state on the originator transactions, performed in its own goroutine, before transactions are delegated.

Signing is always performed by the originator after assembly. Transactions input and output states are known once assembly is completes, which means that it is safe to select a new transaction for assembly as soon as the dependency graph has been updated with this information. I have modified the originator to send an assembly response as soon as it is available, it then transitions into a new state to perform signing and sends a separate signing response. On the coordinator transaction there is a transition into signing state once the assemble response is received. No action is taken in this signing state other than to wait for the sign response. There is a small possibility if signing is fast (performed with the embedded signer) that the sign response may arrive/be handled ahead of the assemble response. To handle this, the assemble response is also sent in the sign response, allowing the coordinator transaction to "fast forward".


core/go/internal/sequencer/coordinator/transaction/dependency_reset_handlers.go
```diff
@@ -0,0 +1,67 @@
...
+
+// dependencyResetHandler handles Event_DependencyReset for a post-assembly state.
+var dependencyResetHandler = dependencyResetToPreAssemblyOrPool()
+
+// dependencyRevertedHandler handles Event_DependencyConfirmedReverted for a post-assembly state.
+var dependencyRevertedHandler = dependencyResetToPreAssemblyOrPool()
+
+// chainedDependencyFailedHandler handles Event_ChainedDependencyFailed for a post-assembly state.
+var chainedDependencyFailedHandler = EventHandlers{
+	Match: statemachine.MatchFirst,
```
matthew1001, line 37: Is this the first time we've moved state machine event handlers outside of a `state_machine.go` file? I sort of feel resigned to the fact that we probably need to structure things in a different way to "everything in a single file". But thought it worth raising. Does the state machine docs handler cope with this?

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -315,6 +345,26 @@ var stateDefinitionsMap = StateDefinitions{
...
+			// against the shared assemble retry budget, mirroring the State_Signing Event_SignError handler.
+			Event_SignError: {
+				Match: statemachine.MatchFirst,
+				Handlers: []EventHandler{{
+					Validator: validator_MatchesPendingAssembleRequest,
+					Actions:   []ActionRule{{Action: action_SignError}},
+					Transitions: []Transition{
+						{
+							If:      guard_CanRetryErroredAssemble,
+							To:      State_Pooled,
```
matthew1001, line 358: I'm wondering what the real world cases for a sign error are, and whether they are caused by the assembly being bad or something else transient. We might in the fullness of time want a separate retry count for signing - it feels odd now we've separated them out, to re-use assmble-retry guards.

operator/config/samples/core_v1alpha1_besugenesis.yaml
```diff
@@ -9,8 +9,8 @@ spec:
   chainID: 1337
   gasLimit: 700000000
   consensus: qbft
-  blockPeriod: 100ms
-  emptyBlockPeriod: 10s
+  blockPeriod: 2s
```
matthew1001: Think these slipped in again :) ?

core/go/internal/sequencer/coordinator/transaction/state_machine.go
annamcallister: They're going to be in the same 'unexpected' category as assemble errors, so either a code bug, likely in the signing module implementation, or the signing service, if not local to Paladin, being unavailable.

I don't think, at least for the domains we have currently, that a bad assemble could cause a signing error.

core/go/internal/sequencer/coordinator/transaction/dependency_reset_handlers.go
annamcallister: Yes it is, and I'm not overly sold that this is the best way forwards, but it felt better than this handlers all inlined in the ~5 places they get used, and the docs generation handles this.

operator/config/samples/core_v1alpha1_besugenesis.yaml
annamcallister: and removed again!

core/go/internal/sequencer/coordinator/transaction/state_machine.go
annamcallister: I've added a separate retry count for signing

core/go/internal/sequencer/coordinator/transaction/dependency_reset_handlers.go
matthew1001: OK thanks. It kind of felt like it was coming given the size and complexity of the state machines

matthew1001 (approved):
LGTM 👍

PR #1299, opened 2026-07-14
Reduce delegations

annamcallister:
Reduce the number of delegation requests the coordinator needs to handle by
* sending delegations at a fixed interval, rather than whenever a delegation arrives- because each delegation request includes transactions already delegated they can get very large and cause a lot of repeat processing
* when delegating to the same coordinator and when there are no missing transactions, only send transactions in pending and delegated state, since these are the only ones which matter for enforcing FIFO until first assemble order


core/go/internal/sequencer/originator/observing.go
```diff
@@ -141,10 +141,35 @@ func (o *originator) processRevertedTransactions(ctx context.Context, event *com
...
+	for _, txn := range o.getTransactionsNotInStates(excludedStates) {
+		// A freshly-delegated transaction races the coordinator's snapshot: the snapshot we are checking
+		// against may have been generated before the delegation reached the coordinator, so its absence is
+		// expected rather than a drop. Only consider a transaction dropped once at least one full heartbeat
+		// interval has elapsed since it was first delegated to the current coordinator, by which point that
+		// coordinator has had the chance to include it in a snapshot. The first-delegation time is reset if
+		// the transaction is redirected to a different coordinator, so the grace restarts on each handover
+		// but is not extended by the partial FIFO resend to the same coordinator.
+		firstDelegated := txn.GetFirstDelegatedTime()
+		if firstDelegated == nil || o.clock.Now().Sub(*firstDelegated) < o.heartbeatInterval {
```
matthew1001: Does this need any extra allowance - 10% extra or something - for any delays (such as network IO) in the case where the delegation arrives at the coordinator the instant after a heartbeat has just been sent?

core/go/internal/sequencer/originator/originating.go
```diff
@@ -102,10 +102,21 @@ func (o *originator) addToTransactions(
 	return nil
 }
 
-func sendDelegationRequest(ctx context.Context, o *originator) error {
+// sendDelegationRequest delegates transactions to the current active coordinator in the order they
+// were created on the originating node.
+//
+// When full is true, every resolved transaction in the contiguous prefix is (re)delegated. This is
```
matthew1001: Could you clarify the meaning of "contiguous prefix" here. I note we also have it in the pre-existing comment below and I'm not sure what we mean by it.

core/go/internal/sequencer/originator/state_machine.go
```diff
@@ -41,6 +41,7 @@ const (
 	Event_OriginatorCreated         EventType = iota + 300 // fired once by Start to drive the initial coordinator selection
 	Event_TransactionCreated                               // a new transaction has been created and is ready to be sent to the coordinator TODO maybe name something like Intent created?
 	Event_DelegationRequestRejected                        // pushed by transport_client when a DelegationResponse arrives with Accepted == false
+	Event_DelegateFlush                                    // fired by the delegation batching goroutine when the batch timer coalesces one or more delegation requests
```
matthew1001: Can I make a case for `Event_DelegateSendBatch`, `Event_DelegateBatchTimerFired` or something a little more descriptive about the action this is going to cause?

core/go/internal/sequencer/originator/originator.go
```diff
@@ -148,6 +161,104 @@ func (o *originator) WaitForDone(ctx context.Context) {
 	o.stateMachineEventLoop.WaitForDone(ctx)
 }
 
+// startDelegationLoop creates the signal channels and starts the batching goroutine. Called from the
+// State_Sending entry hook on the event-loop goroutine. No-op if the originator has not started yet
+// or the loop is already running (nil-guarded like the coordinator dispatch loop).
+func (o *originator) startDelegationLoop() {
```
matthew1001: We've added a lot of delegation-specific logic into `originator.go` - a number of new channels and functions all very focused on delegation behaviour. I'd make a case for having all of this logic in another file in the package, but if you'd prefer to keep it here I think functions like `chooseFlush` and `drainSignal` need to be renamed to be a little clearer what they relate to.

Related to my other comment on event naming, I wonder if terms like `chooseFlush` would be clearer if they called were something like `chooseDelegateBatchType`. I think for a new reader of this area of code, knowing that it relates to:

 - delegations
 - the batching of those delegations
 - what content of those batches of delegations

is the kind of thing I'd like us to bear in mind with any of the functions/state machine/event names.

config/pkg/pldconf/sequencermgr.go
```diff
@@ -77,6 +79,7 @@ var SequencerDefaults = SequencerConfig{
 	ClosingGracePeriod:                confutil.P(2),
 	CoordinatorEventQueueSize:         confutil.P(100),
 	CoordinatorPriorityEventQueueSize: confutil.P(500),
+	DelegationBatchInterval:           confutil.P("100ms"),
```
matthew1001: I think this is implying a minimum average per-TX latency of 50ms? Mainly commenting to check I've understood, not that I necessarily think we should change it, but also now seems a good point to discuss.

core/go/internal/sequencer/originator/observing.go
annamcallister: added

core/go/internal/sequencer/originator/originating.go
annamcallister: I thought I had cleaned this claude provided wording up but apparently not completely :)

core/go/internal/sequencer/originator/state_machine.go
annamcallister: renamed to Event_DelegateSendBatch

core/go/internal/sequencer/originator/originator.go
annamcallister: I've moved delegation code to a new file, removed a number of functions where I felt that actually having the code inline made it easier to understand how the channels are being used, and switched to talking about notifications rather than signals as that's the pattern used across the rest of the code base

config/pkg/pldconf/sequencermgr.go
annamcallister: Dropped to 50ms as the default after discussion.

core/go/internal/sequencer/originator/delegation.go
```diff
@@ -0,0 +1,241 @@
+/*
+ * Copyright © 2025 Kaleido, Inc.
```
matthew1001: nit: copyright format

core/go/internal/sequencer/originator/state_machine.go
```diff
@@ -192,7 +192,7 @@ var stateDefinitionsMap = StateDefinitions{
 			// coordinator. If the coordinator is still in Elect or Prepared it will accept the
 			// delegation and manage the handover itself.
 			{Action: action_StartDelegationLoop},
-			{Action: action_SignalDelegateAll},
+			{Action: action_NotifyFullDelegation},
```
matthew1001, line 195: 👍

matthew1001 (approved):
Thanks for the discussion & changes. LGTM. One minor comment re. copyright in the new file

PR #1300, opened 2026-07-14
Don't reallocate notx

annamcallister:
```
type provider struct {
	p    SQLDBProvider
	gdb  *gorm.DB
	db   *sql.DB
	conf *pldconf.SQLDBConfig
}
```

Is our GORM provider - `gdb  *gorm.DB` is returned wrapped in `newNOTX` on every call to `NOTX` - this results in a new wrapper being allocated every time, which is a key driver of GC and CPU usage. 

Since it's the same underlying `*gorm.DB` we can have a static notx which is returned in every call, reducing allocations.


matthew1001 (approved):
LGTM

PR #1303, opened 2026-07-22
Reduce domain state writer lock contention

annamcallister:
- **Reduce domain state writer lock contention.** The writer used to hold `stateLock` across the entire flush, including the DB write. This caused delays staging state writes post assembly and knock on delays selecting the next transaction for assembly. The lock is now held only for the in-memory buffer swap (`rotateForFlush`); the batch DB write (`exec`) runs lock-free.

- **Nullifier staging made explicit and atomic.** Moving the write out of the lock exposed that nullifier upserts scanned both the `unFlushed` *and* `flushing` buffers to find their creating state. That was technically safe without the lock, but (a) fragile — it happened to touch a field that didn't affect the DB write — and (b) rested on the  assumption that the state would be in one of the two buffers at dispatch. That assumption does not hold under load: states are staged at assembly time but nullifiers are not built until dispatch. If another transaction dispatches in between the states will be flushed to the DB. Nullifiers are now validated against the states supplied in the *same* call and staged atomically with them, rather than looked up across the live buffers.

- **State staging moved from assembly to dispatch.** Post-assembly is too early to write potential states — confidence that a transaction will reach the chain only exists after endorsement and pre-dispatch confirmation. Staging at dispatch lets a transaction's states and nullifiers be staged as a single atomic write. There is no advantage to having states written earlier than dispatch. Nothing reads them from the DB until post dispatch, and flushing them earlier increases the likelihood of states which never have a corresponding create record being persisted (acceptable but wasteful).

- **Label computation left on the assembly path (deliberately).** Label computation could also be deferred to dispatch, taking it off the sequential assembly path. There is likely work coming to export labels — not full state data — in the assemble snapshot,  so it remains part of the post-assembly work for now.

- **Supporting cleanup.** 
   - Retired `StateUpsert` (it added nothing over `prototk.EndorsableState`); the write-path validator now consumes `EndorsableState` directly. Removed the vestigial grapher create-lock gate and the `MapPotentialStates` chain that only fed it.
   - Naming aligned to intent: assembly-time work *resolves* (`ResolveStatesForTransaction` → `ResolvePotentialStates` → `dsw.ResolveStates`), and persistence *stages/writes* at dispatch (`StageWrites`).


core/go/internal/components/statemgr.go
```diff
@@ -94,12 +94,15 @@ type StateQueryOptions struct {
 
 // DomainStateWriter is a long-lived write buffer used for flushing domain states and nullifiers to the DB.
 type DomainStateWriter interface {
-	// StageStateUpserts creates or updates states in the in-memory write buffer.
-	StageStateUpserts(ctx context.Context, dbTX persistence.DBTX, states ...*StateUpsert) (s []*pldapi.State, err error)
+	// ResolveStates validates proto states at the trust boundary and resolves them to states
```
matthew1001: Could you clarify the meaning behind this comment a little for me? Which trust boundary is it referring to?

core/go/internal/components/transaction.go
```diff
@@ -60,6 +60,10 @@ type TransactionPostAssembly struct {
 	OutputStates []*prototk.EndorsableState
 	InfoStates   []*prototk.EndorsableState
 
+	// StatesToStage carries the output+info states resolved at assembly, held in memory so they can be
```
matthew1001: Could you clarify this comment a little for me. It says that it holds output + info states, but we've got in-memory output and info states in the lines above. Why aren't we using those instead of duplicating them in a new slice?

I suspect the code/reason is correct, but the comment doesn't help explain why we have the duplication

core/go/internal/domainmgr/private_smart_contract.go
```diff
@@ -284,26 +284,27 @@ func (dc *domainContract) AssembleTransaction(ctx context.Context, dqc component
...
+	// Resolve the OutputStatesPotential+InfoStatesPotential arrays into concrete states. This runs on
+	// the coordinator once the transaction has been assembled:
+	// 1) Computing their identifiers and storing them into the OutputStates/InfoStates lists
+	// 2) Holding the resolved states in StatesToStage for the write buffer to persist at dispatch
 	postAssembly := tx.PostAssembly
 	outputStatesPotential := postAssembly.AssembleResponse.GetOutputStatesPotential()
 	infoStatesPotential := postAssembly.AssembleResponse.GetInfoStatesPotential()
-	log.L(ctx).Debugf("WritePotentialStates: Writing %+v output states", len(outputStatesPotential))
-	postAssembly.OutputStates, err = dc.upsertPotentialStates(ctx, dsw, readTX, tx, outputStatesPotential, true)
+	log.L(ctx).Debugf("ResolvePotentialStates: Resolving %+v output states", len(outputStatesPotential))
```
matthew1001: Did we decide that this count was useful in the logs? Happy for you to remove it if not (I have a feeling it was me who added it).

core/go/internal/domainmgr/private_smart_contract.go
```diff
@@ -284,26 +284,27 @@ func (dc *domainContract) AssembleTransaction(ctx context.Context, dqc component
...
 	}
 
-	// Now we're confident enough about this transaction to (on the sequencer) to have allocated
-	// it to a sequence, and we want to write the OutputStatesPotential+InfoStatesPotential arrays:
-	// 1) Writing them to the DB (unflushed at this point)
-	// 2) Storing their identifiers into the OutputStatesFull list
-	//
-	// Note: This only happens on the sequencer node - any endorsing nodes just take the Full states
-	//       and write them directly to the sequence prior to endorsement
+	// Resolve the OutputStatesPotential+InfoStatesPotential arrays into concrete states. This runs on
```
matthew1001: Could we bump this comment to above the function rather than here since the current comment `Happens only on the sequencing node` isn't super helpful, and since the interface doesn't have comments for the function declarations

core/go/internal/components/statemgr.go
```diff
@@ -94,12 +94,15 @@ type StateQueryOptions struct {
...
+	// (computing IDs and label values) without staging them in the write buffer. Callers stage
+	// the results later via StageWrites.
+	ResolveStates(ctx context.Context, dbTX persistence.DBTX, states ...*prototk.EndorsableState) (s []*StateWithLabels, err error)
```
matthew1001: Having read where this is used in `resolvePotentialStates` I'm struggling a but with an interface called `DomainStateWriter` having what I believe is a read-only function `ResolveStates`. In the calling site, `dsw.ResolveStates(...)` just reads slightly weirdly when you know that `dsw` is a writer.

core/go/internal/statemgr/state.go
```diff
@@ -487,21 +482,18 @@ func (ss *stateManager) processStateForSet(ctx context.Context, domainName strin
...
 
-// validateStateUpserts validates state upserts against their schemas (write-buffer path).
-func (ss *stateManager) validateStateUpserts(ctx context.Context, domainName string, contractAddress pldtypes.EthAddress, customHashFunction bool, dbTX persistence.DBTX, stateUpserts ...*components.StateUpsert) (*validatedStateSet, error) {
-	vss := &validatedStateSet{
-		states:     make([]*pldapi.State, len(stateUpserts)),
-		withValues: make([]*components.StateWithLabels, len(stateUpserts)),
-	}
-	for i, ns := range stateUpserts {
-		vs, err := ss.processStateForSet(ctx, domainName, contractAddress, customHashFunction, dbTX, ns.ID, ns.Schema, ns.Data, true)
+// validateStates validates proto-native states against their schemas at the write-buffer / mint trust
```
matthew1001: Struggling again a little with the "trust boundary" wording here

core/go/internal/domainmgr/private_smart_contract.go
annamcallister: No I don't think it's that useful as we can establish the number from the arguably more useful logs that include the state IDs elsewherein the code

core/go/internal/components/statemgr.go
annamcallister: Rewritten- it was referring to cross node trust boundaries- but I don't actually think that's a particularly pertienent point to make here.

core/go/internal/components/statemgr.go
annamcallister: Refactored to not have this function anymore

core/go/internal/components/transaction.go
annamcallister: it should be clearer now

core/go/internal/domainmgr/private_smart_contract.go
annamcallister: moved up

core/go/internal/domainmgr/private_smart_contract.go
annamcallister: I've removed the info level comment as well, since it seems particularly out of keeping with what else we log at info level

matthew1001 (approved):
Thanks for all the updates, really like how much cleaner things like the `DomainStateWriter` are now

PR #1304, opened 2026-07-22
Plugin log level + lazy log field binding

annamcallister:
Two changes to improve performance when logging is set to `error`:

- Pass log level through to plugins
- Only encode fields from `WithLogField` and `WithComponent` when actually needed


matthew1001 (approved):
Wasn't quite sure about the term `materialize` in the logging package, but if it's the right term then I'm fine with it.

PR #1306, opened 2026-07-22
Pass resolved local tx to originator

annamcallister:
The transaction was being resolved ahead of every assembly, but this duplicates work already done when handling the new tx ahead of passing it to the originator. 

The resolved transaction is now passed into the originator on the created event and passed through to assembly.


matthew1001 (approved):
LGTM

PR #1308, opened 2026-07-29
[sdk] add catch for privacy group creation

shorsher:


sdk/typescript/src/domains/pente.ts
```diff
@@ -65,7 +65,9 @@ export class PentePrivacyGroupFuture {
       group.genesisTransaction
         ? new TransactionFuture(paladin, group.genesisTransaction)
         : undefined
-    );
+    ).catch((err: any) => {
```
annamcallister, line 68: It looks to me like this is swallowing the error and replacing it with a log?

It's not clear to me why this would be desirable - if this is the intended behaviour please could you give the rational in the PR description.

annamcallister (changes requested):

PR #1309, opened 2026-07-30
Noto nullifiers scheme enhancement to bind coin ownership into the nullifier

jimthematrix:
## Summary

Fixes a potentially exploitable flaw in the `NotoNullifiers` variant where a coin's nullifier was derived from `(amount, salt)` only, so two different coins could share one nullifier. Adds layered guards against the same class of mistake at the notary and on the base ledger.

**Severity**. Exploitation requires an attacker running modified Paladin domain code. Our working assumption is that this is outside the threat model for enterprise deployments, which is why this is "potentially" rather than actively exploitable — but the fix is worth making regardless.

**Breaking:** the nullifier derivation changes, so existing `NotoNullifiers` (variant `0x0102`) deployments must be redeployed. See [Upgrade](#upgrade-notes).

## The vulnerability

`calculateNullifier` hashed `abi.encode(uint256 amount, bytes32 salt)` — no owner, no domain separation. The commitment (state ID) is an EIP-712 hash over the full coin *including* owner, so two coins differing only by owner had **distinct commitments and an identical nullifier**.

Assembly runs on the originating (sender's) node, so the sender chooses output salts. A patched and malicious sender could build a transfer with two outputs sharing a salt and amount — one to the recipient, one to itself — then spend its own copy. The recipient's coin is then permanently unspendable: `_processNullifiers` has recorded that nullifier as used, and there is no recovery because the commitment lives in an append-only tree. Worse, it is silent — spend records are indexed by nullifier, so the recipient's node marks their coin spent and the balance just drops.

Nothing in the pipeline could detect it. The notary validated only that amounts net to zero, that the sender owned the inputs, and the sender's signature; the base ledger only ever sees commitments and nullifier freshness.

Variants of the same root cause:

- **Dead-on-arrival payment** — pay someone a coin whose `(amount, salt)` matches any coin whose nullifier is already burned. Endorsed, accepted, receipt says success, coin unusable on arrival.

## What changed

**1. The nullifier binds the whole coin** (`states.go`)

```javascript
    tag = keccak256("noto:nullifier:coin")
    nullifier = keccak256(abi.encode(tag, salt, owner, amount))
```

Because the preimage now covers exactly the fields that determine the commitment, a collision requires a *duplicate coin* — which the base ledger already rejects (it will not re-add an existing commitment to the tree) and the state store already rejects (same state ID). That closes all three variants above. The derivation stays key-free by necessity: the notary has to recompute it from the unmasked coin data in order to endorse a spend.

**2. One place decides how a state is identified on-chain** (`stateNullifier` / `endorsableStateID`)

Unlocked coins are identified by nullifier; everything else — locked coins and lock info states — by state ID, regardless of what the caller passes. That matches the base ledger (locked states are consumed by ID via `_locked[]`) and local bookkeeping (locked coins are queried without nullifiers, so their spend records are keyed by ID). Previously each call site chose with a boolean and several chose wrongly. Nullifier payload decoding is also strict, so a locked coin can never be nullified as though it were an unlocked one with its `lockId` silently dropped.

**3. Notary-side guard** — `EndorseTransaction` → `validateDistinctNullifiers` (PD200045)

Rejects any transaction whose inputs and outputs contain two states deriving the same nullifier. Placed at the single endorsement entry point rather than per-handler, so no transaction type can be missed. Strictly belt-and-braces after (1), but it turns a base ledger revert into a clear endorsement error and catches any future regression in the derivation.

**4. On-chain guard** — `NotoNullifiers._processNullifiers`, new error `NotoNullifierIsCommitment`

Rejects an input that is itself a commitment in the tree. The contract cannot verify that a nullifier corresponds to an unspent leaf — that is the point of a nullifier, and proving it needs a membership proof this variant does not carry — but commitments *are* public, so it can cheaply catch a caller that spends by commitment ID. Without this, that mistake is accepted silently, leaving the real nullifier unburned and the coin spendable again. A legitimate nullifier can
never trip it short of a hash collision.

## Fixes uncovered by adding the missing coverage

No test were ran on a lock/unlock flow against the nullifier variant (`testNotoLock` only deployed the plain implementation, and `TestNotoV1Nullifiers` covers mint/transfer/burn flows only).

## Upgrade notes

Nullifiers already burned on-chain and `state_nullifiers` rows written by existing nodes use the old `keccak256(amount, salt)` formula and will not match. **Existing nullifier-variant deployments must be redeployed; there is no migration.**

The golden vector in `states_test.go` carries a comment so that a future accidental change to the derivation trips a test with an explanation rather than silently breaking deployed contracts.


domains/noto/internal/noto/states.go
```diff
@@ -585,29 +624,67 @@ func encodedStateIDs(states []*pldapi.StateEncoded) []string {
...
+	if n.coinSchema == nil || state.SchemaId != n.coinSchema.Id {
+		return "", false, nil
+	}
+	coin, err := n.unmarshalCoinStrict(state.StateDataJson)
+	if err != nil {
+		return "", true, err
+	}
+	hash, err := calculateNullifier(ctx, contract, coin)
+	if err != nil {
+		return "", true, err
```
annamcallister: While all consumers of this function disregard the return values for `nullifier` and `hasNullifier` if an error is returned, this function would read more easily if `hasNullifier` was returned as `false`, since this is the default go boolean value. `true` suggests intention and I spent a while trying to establish what the semantics of `true` AND `err` meant.

domains/noto/internal/noto/handlers.go
```diff
@@ -144,6 +144,73 @@ func (n *Noto) validateUnlockAmounts(ctx context.Context, tx *types.ParsedTransa
...
 
+// Check that no two coins in the transaction derive the same nullifier.
+//
+// The nullifier derivation covers every field of a coin, so a collision means a duplicate
+// coin - which is already rejected by the base ledger and the state store. This check is
+// belt and braces: it catches any regression in the derivation, and turns what would be a
+// base ledger revert (or worse, an unspendable coin) into a clear endorsement failure.
+//
+// Both inputs and outputs are checked as one set, because an output that collides with an
+// input is nullified by the very transaction that creates it.
```
annamcallister, line 155: 👍 to moving failure checking as early as possible, and not potentially spending gas to get a base ledger revert when we can recognise the condition ahead of submission. 

However, also noting that endorsement validation doesn't interact completely cleanly with the sequencing algorithm as-is. Transaction which fails endorsement are currently reset back to `Pooled` state- to retry assembly and endorsement indefinitely https://github.com/LFDT-Paladin/paladin/blob/2a4fbc98cad7dfaa74e8199bfad42289d8b7dda8/core/go/internal/sequencer/coordinator/transaction/state_machine.go#L632. I think this is a fairly serious bug in the sequencer logic- it allows a node to delegate transactions that indefinitely occupy an inflight transaction slot on the notary/coordinator, taking up valuable sequential assembly time, and eventually potentially filling all the inflight slots and blocking the whole domain instance. A notary choosing not to endorse a transaction should be a final outcome for that transaction.

domains/noto/internal/noto/noto.go
```diff
@@ -777,14 +777,36 @@ func (n *Noto) AssembleTransaction(ctx context.Context, req *prototk.AssembleTra
...
 
 func (n *Noto) EndorseTransaction(ctx context.Context, req *prototk.EndorseTransactionRequest) (*prototk.EndorseTransactionResponse, error) {
 	ctx, tx, handler, err := n.validateTransactionAndGetLogContext(ctx, req.Transaction)
 	if err != nil {
 		return nil, err
 	}
+	// Defense in depth for the nullifier variants: the notary is the only party that can
+	// detect a nullifier collision, as the base ledger never sees the coins behind the
+	// nullifiers. Applied to every handler here rather than per-handler, so no transaction
+	// type can be missed.
```
annamcallister: This comment doesn't feel completely accurate.

This check uses the notary's full visibility of the coins in _this_ transaction to preempt what would be a one cause of a base ledger revert, but it can't recognise all potential duplicates (e.g. when a the nullifier for a new output state clashes with a nullifier for an output state from a previous transaction)

The full fix being delivered in this PR isn't based on the notary's visibility of all coins, it's tying the uniqueness of nullifiers to the uniqueness of state IDs, which the base ledger does have visibility of, and ultimately enforces.

This is well described elsewhere- it's just this comment that feels misleading.

annamcallister (commented):
I think there might still be an exploitable gap around duplicate nullifiers which this PR doesn't cover. 

The nullifier is computed over a parsed address for the owner, so its always in canonical form. There is nothing to prevent the owner that is hashed into the state ID from being in another form, e.g. checksummed, so 

Coin A- state ID: `0xc5ec8cc3aa6dcdff5dd10e298d411f018bdcc4ce3a7adffc53e9a086c927815b`

```json
{
  "amount":"100",
  "owner":"0xbB2b99DDe4cA2D4c99f149D13cd55a9EDaDA69Eb",
  "salt":"0x3f2b1d9c8a7e6f5d4c3b2a19087f6e5d4c3b2a19087f6e5d4c3b2a1908f7e6d5"
}
```
and
Coin B- state ID: `0xc3787eaf1a8f127bacc88a14b02b0e2a48f6cc973694ea33e416a0cbd0e45ea2`
```json
{
  "amount":"100",
  "owner":"0xbb2b99dde4ca2d4c99f149d13cd55a9edada69eb",
  "salt":"0x3f2b1d9c8a7e6f5d4c3b2a19087f6e5d4c3b2a19087f6e5d4c3b2a1908f7e6d5"
}
```

both derive nullifier: `3bf497ab754c649ef6f4019826e35e7a0924186d8f1273edc08f25fcbbd53e44`

If coin A is minted first, when the tx is indexed by the recipient, the nullifier is associated in the DB with the private state data for coin A. When coin B is minted, the duplicate nullifier insert hits the do nothing on conflict. Coin A is unselectable by the recipient for future transactions because on an unmodified node, availability queries use the canonical eth address as the label value for owner. Coin B is unselectable as it is not associated with a nullifier.

annamcallister:
There's some missing detail in the PR description under
> Fixes uncovered by adding the missing coverage

I think the fixes amount to:
- creating nullifier specs for the unlocked change from lock operations
- creating nullifier specs for unlocked coins
- adding these nullifiers to the storage tree when indexing the transactions which created the corresponding output states

Does my understanding sound correct.

jimthematrix:
The fact that two Noto coins sharing the same salt, amount and the owner but have different spelling (casing) of the same owner's address are being allowed to be minted is itself a problem that should be addressed.

Because if two coins are minted with only the address spelling difference, only one of them (all lower cases address string) is currently selectable for spending, because that's the format resolved identifiers use.

I think this also points to a code problem where the "owner" was treated as a string rather than an address, such that the hash for the address was not done over the address' internal bytes, but on the string, thus resulting in different hashes for the same address (but different spelling). IF the owner field needs to be a string, then we need to make sure its consistently handled in terms of spelling.

domains/noto/internal/noto/handlers.go
jimthematrix: > Transaction which fails endorsement are currently reset back to Pooled state- to retry assembly and endorsement indefinitely

This does sound problematic. Is this by design or a point in time state?

jimthematrix:
on the question for the details of the fix for the lock related transactions with the NotoNullifier token, beside what you listed above, it also fixed the transaction assembly to include the SMT root in the `proof` parameter, along with the notary signature.

domains/noto/internal/noto/noto.go
```diff
@@ -798,9 +798,8 @@ func (n *Noto) EndorseTransaction(ctx context.Context, req *prototk.EndorseTrans
 	if err != nil {
 		return nil, err
 	}
-	// Defense in depth for the nullifier variants: the notary is the only party that can
-	// detect a nullifier collision, as the base ledger never sees the coins behind the
-	// nullifiers. Applied to every handler here rather than per-handler, so no transaction
+	// Defense in depth for the nullifier variants: catches invald transactions that includes inputs/outputs states
```
annamcallister:
```suggestion
	// Defense in depth for the nullifier variants: catches invalid transactions that includes inputs/outputs states
```

annamcallister:
Thanks @jimthematrix - I spotted a comment typo but otherwise looks good.

Should probably note that while this will be a completely breaking change for Noto with nullifiers landing outside of a major version upgrade, it is acceptable since nullifiers were only added in the last release, and, as the fixes in this PR demonstrate, a `v1.0` Noto with nullifiers contract will never work properly with locks.

annamcallister (approved):

Issue #1240, opened 2026-06-16
Enhancement: Java SDK - WebSocket subscription client


divyam-jha123:
### What would you like to be added?

Implement the WebSocket subscription client so callers can receive push notifications
(e.g. transaction receipts, block / event subscriptions) rather than polling.

### Reference implementations
- `sdk/go/pkg/wsclient/wsclient.go` — Go WebSocket client
- `sdk/typescript/src/websocket.ts` — TS WebSocket client
- `sdk/go/pkg/wsclient/wstestserver.go` — useful pattern for a test server

### Scope
- [ ] WebSocket JSON-RPC client with subscribe / unsubscribe
- [ ] Pub/sub dispatch of subscription notifications to typed handlers
- [ ] Automatic reconnect with configurable backoff; re-establish active subscriptions on reconnect
- [ ] Connection lifecycle (connect, close) returning `CompletableFuture<T>` where appropriate
- [ ] Surface connection / subscription failures through the `PaladinException` hierarchy
- [ ] Unit tests against a mock WebSocket server (covering subscribe, message dispatch, reconnect)

### Out of scope
HTTP RPC transport (already delivered), RPC-namespace method coverage, domain helpers.

### Acceptance criteria
- Client can establish a subscription and dispatch notifications to a typed handler
- Reconnect re-establishes active subscriptions and is exercised by unit tests
- No third-party WebSocket dependency required in the default build (or dependency justified)

### Why is this needed?

Part of the Java SDK RFC #1224

PR #1310, opened 2026-07-31
feat(java-sdk): fluent TxBuilder with deferred validation and receipt polling

### LINKED ISSUE #1233: Enhancement: Java SDK - TxBuilder

Implement the fluent transaction builder that provides a high-level API for
constructing and submitting Paladin transactions, with receipt polling.

### Reference implementations
- `sdk/typescript/src/transaction.ts` - TS fluent builder
- `sdk/go/pkg/pldclient/txbuilder.go` - Go builder

### Scope
- [x] Fluent `TxBuilder` API (method chaining, deferred validation)
- [x] Errors deferred until `send()` / `build()` - not thrown during chaining
- [x] `send()` submits the transaction and returns a `CompletableFuture<TransactionReceipt>`
- [x] Built-in receipt polling with configurable interval and timeout
- [x] Unit tests covering builder construction, validation errors, and receipt polling

### Acceptance criteria
- Builder API is ergonomic for Java callers (fluent, no required intermediate objects)
- Deferred-error behavior is verified by tests (malformed tx does not throw until `send()`)
- Receipt polling respects the configured timeout and throws `PaladinTimeoutException`

### Why is this needed?

Part of the Java SDK RFC #1224

divyam-jha123:
Closes #1233. Part of the Java SDK RFC #1224.

Adds the fluent transaction builder in a new `org.lfdt.paladin.sdk.client.tx` package, layered over `PtxClient`. Reference implementations: `sdk/go/pkg/pldclient/txbuilder.go` and `sdk/typescript/src/transaction.ts`.

```java
TransactionReceipt receipt =
    TxBuilder.on(rpc)
        .publicTx()
        .from("alice")
        .to("0x0102030405060708090a0b0c0d0e0f1011121314")
        .function("transfer")
        .abiJson(abiJson)
        .inputs(Map.of("to", "0x...", "amount", 100))
        .receiptTimeout(Duration.ofMinutes(1))
        .send()
        .join();
```

## What's here

**Fluent API** covering the full `TransactionInput` surface — `publicTx()`/`privateTx()`, `from`, `to`, `function`, `constructor()`, `domain`, `abi`/`abiEntry`/`abiJson`, `inputs`/`inputsJson`, `idempotencyKey`, `dependsOn`, `gas`/`value`/EIP-1559 fees, `bytecode` — with convenience overloads that parse hex addresses, bytecode, ABI JSON and arbitrary input values.

**Deferred errors.** No chaining call throws. Setters that can fail record a `PaladinTransactionException` and return `this`, so a chain reads uninterrupted and a malformed transaction surfaces at exactly one place. The first error wins and is replayed from `build()` (thrown) and from `submit()`/`send()` (as a failed future).

**Validation** mirrors Go's `txBuilder.validateForSend` — type required, domain required for private, the function↔`to` pairing, and the public-needs-bytecode / private-rejects-bytecode deploy rules — plus the signing-identity check Go applies on send.

**Receipt polling.** `send()` submits and then polls `ptx_getTransactionReceipt` every `pollingInterval` (default 1s) until a receipt lands or `receiptTimeout` (default 30s) elapses, at which point the future fails with `PaladinTimeoutException`. Polling uses `CompletableFuture.delayedExecutor`, so it holds no thread while waiting, and clamps each sleep to the time remaining — a polling interval coarser than the timeout still fails on schedule rather than overshooting. `submit()` sends without waiting and returns the transaction id.

Also adds `PaladinInvalidTransactionException` for client-side rejections, listed among the subtypes documented on `PaladinException`.

## One deliberate divergence from Go — worth a look

A reverted transaction completes `send()` **normally**, with a receipt whose `success()` is `false`. Go's builder defers an error in that case instead. The reasoning: callers can reach `failureMessage()`/`revertData()` directly rather than unwrapping a `CompletionException`, and it matches how receipts are already consumed elsewhere in the Java SDK. Happy to flip to Go parity if reviewers prefer consistency across SDKs.

## Testing

33 tests built on the existing `MockJsonRpcServer` harness — no new test dependencies. They cover full-field assembly, every validation rule, every deferred-error path (including that a malformed transaction does not throw until the terminal call, and that nothing reaches the node), submit-without-polling, poll-until-receipt, reverted receipts, timeout timing in both directions, and a transport failure mid-poll.

```
./gradlew :sdk:java:core:check :sdk:java:client:check :sdk:java:domains:check :sdk:java:testing:check
```

Green across tests, Checkstyle, Spotless, Javadoc `-Werror` doclint and JaCoCo. `TxBuilder` is at 100% instruction / 98% branch; the client module sits at 97% against the 0.78 gate.


sdk/java/client/src/main/java/org/lfdt/paladin/sdk/client/tx/TxBuilder.java
```diff
@@ -0,0 +1,641 @@
...
+ * <em>normally</em> with {@code success() == false}; inspect {@link TransactionReceipt#success()}
+ * rather than relying on the future failing. Use {@link #submit()} to submit without waiting.
+ *
+ * <p>Instances are mutable and not thread-safe; build and send from a single thread. The returned
+ * futures may be composed freely from any thread.
+ */
+public final class TxBuilder {
+
+  /** Default delay between receipt polls, used unless {@link #pollingInterval(Duration)} is set. */
+  public static final Duration DEFAULT_POLLING_INTERVAL = Duration.ofSeconds(1);
```
matthew1001, line 89: I think this is somewhat higher than the Typescript default (100ms I believe) although Golang does use `1s`. I think we ought to come back to this once we have WebSocket support in the SDK, so perhaps we leave it as `1s` for now. But it does seem a little slow

sdk/java/client/src/main/java/org/lfdt/paladin/sdk/client/exception/PaladinInvalidTransactionException.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * <p>This is the deferred-error carrier for the fluent transaction builder: a chaining call that
+ * cannot be honoured (an unparseable address, malformed ABI JSON) records one of these instead of
+ * throwing, and the builder replays it — along with any structural validation failure, such as a
+ * missing transaction type — from {@code build()}, {@code submit()} and {@code send()}.
+ *
+ * <p>Unlike the transport-level failures in this package it never indicates a problem with the node
+ * or the network; the transaction as described could not be assembled at all.
+ */
+public class PaladinTransactionException extends PaladinException {
```
matthew1001: Maybe it's better to have a more specific exception name - `PaladinInvalidTransactionException` perhaps?

matthew1001 (commented):
Looking good @divyam-jha123, one main comment to address

sdk/java/client/src/main/java/org/lfdt/paladin/sdk/client/exception/PaladinInvalidTransactionException.java
divyam-jha123: yes, thats sounds specific

sdk/java/client/src/main/java/org/lfdt/paladin/sdk/client/tx/TxBuilder.java
divyam-jha123: ya we will probabily get back to it in https://github.com/LFDT-Paladin/paladin/issues/1240 - the websocket issue.

I took Go as the reference deliberately, since this builder is modelled on `pldclient.txBuilder` throughout — deferred first-error-wins validation, the `validateForSend` checks, the same clamp so the interval never overshoots the receipt deadline. Matching its polling default keeps that parity rather than mixing conventions from two SDKs.

matthew1001:
@divyam-jha123 re this part of the PR description:

```
A reverted transaction completes send() normally, with a receipt whose success() is false. Go's builder defers an error in that case instead. The reasoning: callers can reach failureMessage()/revertData() directly rather than unwrapping a CompletionException, and it matches how receipts are already consumed elsewhere in the Java SDK. Happy to flip to Go parity if reviewers prefer consistency across SDKs.
```

Is there not a more basic divergence between the Java and Go SDKs here, which is that your `send()` call does `pollForReceipt()` where the Go SDK expects you to do `Send().Wait()` to get the equivalent behaviour?

I think that decision is separate to whether or not you have an exception on revert or treat that as a success followed by calling.

So I think you're suggesting that a revert isn't a throwable, which is the same way that the Typescript SDK behavesIIRC, but like you say, the Go SDK considers that an "error", not "inspect the fields to find out". I'm happy with revert not being a throwable. So the remaining question might be about `Send()` and/or `Wait()`.

In both the Go and Typescript SDKs, `Send()` doesn't wait for TX completion. I think it would be good for the Java SDK to retain that behaviour - you've got `submit()` as the function that submits and doesn't wait - I think this should probably be called `send()` for consistency with the other SDKs. Then maybe you have `wait()` or `waitForReceipt()` which calls `private pollForReceipt()` so a caller might look like:

```
CompletableFuture<TransactionReceipt> future = client.newTx()
      .type(TransactionType.PRIVATE)
      .domain("noto")
      ...
      .send()
      .waitForReceipt();

  future.whenComplete((receipt, error) -> {
      ...
  }
```

rather than

```
CompletableFuture<TransactionReceipt> future = client.newTx()
      .type(TransactionType.PRIVATE)
      .domain("noto")
      ...
      .send();

  future.whenComplete((receipt, error) -> {
     ...
  }
```

divyam-jha123:
@matthew1001 , `send()` now submits and returns immediately, and waiting is a separate explicit step, matching Go's `Send()` and TS's `sendTransaction()`:                                       

```                                                                                                            
   CompletableFuture<TransactionReceipt> future =                                                                                                                              
       client.newTx()                                                                                                                                                          
           .privateTx()                                                                                                                                                        
           .domain("noto")                                                                                                                                                     
           ...                                                                                                                                                                 
           .send()                                                                                                                                                             
           .waitForReceipt();        
                                                                                                                                          
```                                                                                                                                                                          
   Concretely:                                                                                                                                                                 
   - The old submit() is gone; its behaviour is now send().                                                                                                                    
   - send() returns a new SentTransaction handle carrying the polling settings in force at send time. On it: `waitForReceipt()`, waitForReceipt(Duration) for a per-call timeout 
   override, id() for the transaction id without waiting, plus getReceipt()/getTransaction() for one-shot lookups.                                                             
   - `pollForReceipt()` moved out of TxBuilder into SentTransaction as the private polling loop, unchanged — same deadline clamp, same async no-thread-held waiting.             
   - Added `PtxClient.newTx()` as the entry point, so the chain reads as `client.newTx()...` like your snippet.                                                                    
                                                                                                                                                                               
Deferred validation is unchanged: a chain that recorded an error still returns a handle whose futures all fail with the same `PaladinInvalidTransactionException`, so nothing throws mid-chain.                                                                                                                                                           
                                                                                                                                                                               
And agreed on revert staying non-throwable — leaving that as-is (TS behaviour), with `success()/failureMessage()/revertData()` on the receipt.                                
                                                                                                                                                                               
Polling default stays at `1s` per the other thread; worth revisiting under #1240 with WebSocket support. `:sdk:java:client:check` is green local

matthew1001 (approved):
LGTM @divyam-jha123 👍

PR #1313, opened 2026-08-05
Remove double WithContext clone

annamcallister:
Modifies the `DB` call that lives on `Transaction`, `NOTX` and `Persistence` to take a ctx parameter, removing the need for a `WithContext` call.

For `Transaction`, which is created with a context, this was resulting in a redundant clone of the GORM statement, which was driving ~21% of allocations, and creating unnecessary garbage collection pressure.

The fix couldn't be as simple as stop calling `WithContext`, since it is necessary for `NOTX` and `Persistence` calls. Where we support both a transaction or `NOTX`, there isn't a way to simply decide whether setting the context is needed or not. There is also the risk when we use `WithContext` in some places, that it can be copied inappropriately in new code, reintroducing the redundant clones.

Making it a parameter for `DB()` allows each implementation to decide whether they will use the parameter or not, and removing all uses of `WithContext` from the code base makes it less likely that it will be inadvertently reintroduced. (We can't guarantee it though since its part of GORM and not our libraries).


matthew1001 (approved):
While it clearly touches a lot of call sites, the approach of "always pass, decide whether to use" is a nice concise way of eliminating unnecessary calls while retaining those that are still needed. So in short, LGTM :)

PR #1314, opened 2026-08-05
Distributed sequencer metrics

annamcallister:
Adds more metrics to the distributed sequencer. I owe a guide to using these in a follow up PR.


core/go/internal/sequencer/coordinator/dispatch_loop.go
```diff
@@ -25,12 +26,20 @@ import (
 	"github.com/LFDT-Paladin/paladin/core/pkg/persistence"
 )
 
+// queuedDispatch carries a transaction onto the dispatch queue along with the time it was enqueued,
+// so the dispatch loop can observe how long it waited before being dequeued.
```
matthew1001:
```suggestion
// so the dispatch loop can record a metric for how long it waited before being dequeued.
```

core/go/internal/sequencer/metrics/metrics.go
```diff
@@ -29,26 +29,73 @@ type DistributedSequencerMetrics interface {
 	IncDispatchedTransactions()
 	IncConfirmedTransactions()
 	IncRevertedTransactions()
-	ObserveSequencerTXStateChange(state string, duration time.Duration)
+	ObserveSequencerTXStateChange(role, state string, duration time.Duration)
```
matthew1001:
```suggestion
	ObserveSequencerTXStateChange(role, fromState string, duration time.Duration)
```

I think this is how we're using this now - also a `from` state?

core/go/internal/sequencer/metrics/metrics.go
```diff
@@ -66,24 +113,51 @@ func InitMetrics(ctx context.Context, registry *prometheus.Registry) *distribute
...
 	metrics.activeCoordinators = prometheus.NewGauge(prometheus.GaugeOpts{Name: "active_coordinators",
 		Help: "Distributed sequencer active coordinators", Subsystem: METRICS_SUBSYSTEM})
 	metrics.activeSequencers = prometheus.NewGauge(prometheus.GaugeOpts{Name: "active_sequencers",
 		Help: "Distributed sequencer active sequencers", Subsystem: METRICS_SUBSYSTEM})
 	metrics.coordinatingTransactions = prometheus.NewGauge(prometheus.GaugeOpts{Name: "coordinating_txns",
 		Help: "Distributed sequencer coordinating transactions", Subsystem: METRICS_SUBSYSTEM})
+	metrics.domainCall = prometheus.NewHistogramVec(prometheus.HistogramOpts{Name: "domain_call_duration_seconds",
+		Help: "Wall time of one domain-plugin RequestReply round-trip", Subsystem: METRICS_SUBSYSTEM, Buckets: operationBuckets}, []string{"domain", "method"})
+	metrics.assembleResponseApply = prometheus.NewHistogram(prometheus.HistogramOpts{Name: "assemble_response_apply_duration_seconds",
+		Help: "Coordinator response-queue wait plus applyPostAssembly", Subsystem: METRICS_SUBSYSTEM, Buckets: operationBuckets})
```
matthew1001: It looked like this metric was purely the duration of the `applyPostAssembly()` call - does the response-queue wait time also get included in the time?

core/go/internal/sequencer/coordinator/dispatch_loop.go
```diff
@@ -136,6 +148,18 @@ func (c *coordinator) dispatchBatch(ctx context.Context, batch []transaction.Coo
 		return
 	}
 
+	// Record the composition of the batch about to be persisted; the low end of the histogram
+	// reveals batches collapsing to a single entry.
+	var public, private, prepared int
+	for _, pd := range dispatchBatch.Dispatches() {
+		public += len(pd.Dispatch.PublicDispatches)
```
matthew1001, line 155: Not specific to this PR, but do we ever have a case where these slices have > 1 entry in them? No need to address here, but the `loop + len()` made me dig into whether we have > 1 entry in any of these

matthew1001 (approved):
LGTM, a few minor comments/questions

core/go/internal/sequencer/coordinator/dispatch_loop.go
annamcallister: No we don't. This has been a feature of the code structure for some time but no existing domain uses it and while there's an argument for leaving flexibility for future domains/enhancements, its hard to see this ever being used, since more than one dispatch of any type removes atomicity and the transactionality of the transaction...

There is a challenge to changing it, as it would require a proto change on the domain interface.

core/go/internal/sequencer/metrics/metrics.go
annamcallister: No, but intentionally; applying the post assembly is a critical piece of work that needs to be kept short.

I've added a separate metric for tracking how long an event stays on each queue (normal priority). I haven't distinguished this by event type as it's a function of everything else that's going on the system and its the overall wait time we'd be aiming to reduce.

PR #1315, opened 2026-08-05
Query in memory states on the coordinator

### LINKED ISSUE #1248: Request missing private state data on demand rather than including in every assemble/endorsement request

The ability for a node which is assembling/endorsing to selectively request private state data it is entitled to from the active coordinator, rather than having it all included in the assemble/endorsement request.

### Why is this needed?

In the current model the active coordinator includes a full snapshot of locks and private state data representing its ahead of chain view of the world, as well as the locks and private state data for transactions confirmed in the last `N` blocks, where `N` is the configured block height tolerance. Under load this can result in very large snapshots.

Private state data can be distributed as soon as immediately after the assembly which proposes it, meaning that most assemblers/endorsers likely already hold the private state data they receive in the snapshot.

annamcallister:
For private transactions to be assembled optimistically, using the outputs of unconfirmed transactions as potential inputs, the transaction originator doing the assembly requires visibility the ahead of chain view of both available and spent states maintained by the domain instance's coordinator's grapher.

Currently this visibility is provided by including a "snapshot" of the grapher's current state, alongside available private state data, in every assemble request. This snapshot includes the locks and available private state data forming the unconfirmed graph, as well as the locks and available private state data from the previous N confirmed blocks, where N is the configured block height tolerance. The originator imports this snapshot into its domain query context, and superimposes it over its own persisted records of states confirmed/spent on chain. This approach has worked well when the ahead of chain view is small, but under heavy sustained load, where many transactions are being mined into each block, the snapshot can grow very large. This results in large amounts of data being sent cross node, but also drives a non trivial amount amount of allocationss and subsequent garbage collection from marshalling the request.

This PR replaces the imported snapshot with a `StateView` component. An originator passes a `StateView` client, initialised with the assemble request ID, into each domain query context it creates for assembly. When the assembly needs to query for available private state data, the query is passed to the client, which forwards it to the coordinator. A coordinator maintains a `StateView` server, and each time it sends an assemble request, it uses the server to create a new session specifically for that assemble request. The session is preloaded with the same snapshot (i.e. scoped to the assembling node's permitted visibility) that would have previously been sent to the originator, meaning the query can run on the coordinator and only the matching available private states are sent back to the originator. Each session is served by its own goroutine, which the transport client queues requests to directly, avoiding the additional latency of going via an event on the coordinator event loop. A request is only accepted if the sending node matches the assembling node recorded on the session. Node identity authentication is performed at the transport layer so is not built into any of the sequencer components.

After making this change, performance profiling still showed that building the snapshot on the coordinator was the largest drivers of CPU (direct and indirect from garbage collection via allocations). This lead to the following changes:
* Adding a copy of the labels that are already computed when processing a transaction's assembly result into the state visibility store, instead of computing the labels each time a snapshot is built.
* Maintaining an incremental view of the current snapshot for each inflight originator, tracking changes to a small number of states with each new block/assembly result , rather than building a large snapshot from scratch with each assembly request.
* Keeping a cache of validated states


matthew1001 (commented):
This feels like quite a fundamental change in architecture that I'm cautious of us dropping in casually to a performance PR.

Here are a few high-level thoughts:
1. This ought to come with some updated docs/diagrams that clearly describe this new relationship between originator and coordinator. When will remote requests be used, who will satisfy them, why is it safe to offload DB queries to some other node etc.
2. There needs to be a clear outline of how this is safe and is not going to expose state data to the wrong nodes. What if a node guesses someone else's assemble ID? Can they just ask for states relating to someone else?

I'll carry on with the rest of the review at more of a code level, but I suspect the higher level topics above are the things we'll need most discussion on.

core/go/internal/msgs/en_errors.go
```diff
@@ -525,6 +528,10 @@ var (
 	MsgSequencerInvalidEndorserCandidate           = pde("PD012646", "Invalid coordinator endorser candidate identity '%s'")
 	MsgSequencerGrapherAddMinterAlreadyExistsError = pde("PD012647", "Error adding TX %s as minter for state %s, already minted by %s")
 	MsgSequencerStaticCoordinatorNotSet            = pde("PD012648", "Static coordinator mode for contract %s has no configured coordinator node")
+	MsgSequencerStateViewFailed                    = pde("PD012655", "Coordinator state view request %s failed: %s")
```
matthew1001: Is there a reason for the gap in message numbers here?

core/go/internal/sequencer/coordinator/grapher/grapher.go
```diff
@@ -424,3 +434,38 @@ func (g *grapher) ExportStatesAndLocks(ctx context.Context, node string) (*proto
 	log.L(ctx).Debugf("ExportStatesAndLocks: %d output states (filtered from %d), %d locks (node=%q)", len(result.States), len(allStates), len(result.Locks), node)
 	return result, nil
 }
+
+func (g *grapher) SnapshotViewForNode(ctx context.Context, node string) ([]*prototk.SnapshotState, []pldtypes.HexBytes) {
```
matthew1001: It feels like we have a slightly deviation here from the grapher/visibility tracker divide. `ExportStatesAndLocks` takes a node and relies on `stateVisibilityTracker.GetForNode` to enforce the node-specific view. Here we've introduced `SnapshotViewForNode` which also takes a node - so I think there's an argument it should just be `SnapshotView`, and it already correctly uses `tateVisibilityTracker.RangeForNode` to enforce the per-node view.

So a long way of saying I think this should be called `SnapshotView`.

matthew1001:
> This feels like quite a fundamental change in architecture that I'm cautious of us dropping in casually to a performance PR.
> 
> Here are a few high-level thoughts:
> 
> 1. This ought to come with some updated docs/diagrams that clearly describe this new relationship between originator and coordinator. When will remote requests be used, who will satisfy them, why is it safe to offload DB queries to some other node etc.
> 2. There needs to be a clear outline of how this is safe and is not going to expose state data to the wrong nodes. What if a node guesses someone else's assemble ID? Can they just ask for states relating to someone else?
> 
> I'll carry on with the rest of the review at more of a code level, but I suspect the higher level topics above are the things we'll need most discussion on.

Having worked through the code review I think I've shifted on my opinion here. While it's a client/server session with the client requesting data from the server (coordinator), it's really just a refactored internal mechanism for how existing data is shared with the originator. I think the newly introduced concepts in the PR like "sessions" and client/server are the things that push on what extra checks & balances are required. In short, I don't think any are. But I think these terms may not be the most useful for future code maintainers.

Here are some follow on thoughts that largely replace the ones above:

1. Client and server aren't really the best terms to use here imo. We already have a communications channel open between originators and coordinators, so the isn't really a new client/server here in the traditional sense.
2. I think the "session" concept is also worth reconsidering. In some respects we have a number of session-like aspects:
   - There's the currently delegated transactions
   - The currently assembling transaction
   - The currently endorsing transactions
   - ...

The PR adds `assembleSessionID` to `transaction`, but I'm wondering what's different to the `assembleRequestID`. Is there a reason the `StateView` server doesn't just have a view per `assembleRequestID`? A session typically means a long-lived context for multiple requests, but in this PR I think we're using it to refer to a short-lived per-assemble request so perhaps it's useful to consider if it's the right terminology.

For me this is all sort of pushed on by the new `action_OpenStateViewSession` state machine actions. In many respects the state machine for a transaction feels like it's the per-transaction session, rather than this one specific part of the state machine.

The other question is not about the specifics of this PR, but it makes me want to raise it again. At some point we need to decide when we can no longer make breaking changes to the state machine protocol. I think this starts with us defining 2 types of state machine change - backwards compatible and non-backwards-compatible. We can think about doing this before we hit a release where we assert no more breaking changes, and I'd make a case for some of our smaller state machine changes being designed with backwards compatibility in mind. That will push us to consider how a state machine can support > 1 valid discrete paths through to success. Perhaps we start with a PR template update that asks "is the state machine change backwards compatible" and "is the protobuf change backwards compatible". We can answer "no" for the time being, but I think a standard approach for even making the decision would be valuable soon. The reason I bring it up here is that this feels like one of the first PRs that could be an "optional optimisation". If the originator supports it, the coordinator doesn't send a snapshot. If the originator doesn't support it, the assemble has the old-style snapshot included. Not a serious suggestion for refactoring this one - but I just feel we need to not ignore it for much longer

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -285,9 +285,16 @@ var stateDefinitionsMap = StateDefinitions{
 	State_Assembling: {
 		OnTransitionTo: []ActionRule{
 			{Action: action_ScheduleStateTimeout},
+			// Refresh block height for the request, then open a stateview session that captures a
+			// snapshot of the states available to the originator.
+			// OnTransitionFrom closes the session on every exit from this state.
 			{Action: action_RefreshBlockHeight},
+			{Action: action_OpenStateViewSession},
```
matthew1001: `action_CaptureGrapherSnapshot`?

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -285,9 +285,16 @@ var stateDefinitionsMap = StateDefinitions{
...
 			{Action: action_SendAssembleRequest},
 		},
+		OnTransitionFrom: []ActionRule{
+			{Action: action_CloseStateViewSession},
```
matthew1001: `action_DeleteGrapherSnapshot`?

matthew1001 (approved):
Following our conversation about `Provider`/`Reader` I've been struggling to come up with terms that are more useful, and within the context of `stateview` they are probably as good a starting point as any. I think the new state view code probably needs to settle for a bit before it'll become clear what terminology might work better, if anything

PR #1317, opened 2026-08-05
Error handling for prepare and dispatch transaction

annamcallister:
Fixes a bug observed under heavy load, where when domain calls over GRPC start to exceed their 10s timeout, failed `PrepareTransaction` calls cause a transaction to become stuck in `ReadyToDispatch`, as there is no error handling for `PrepareTransaction` failures in the current coordinator transaction state machine. This has two knock on effects:
- the failure is ignored by the dispatch loop, which continues to dispatch transactions after this one, even when they depend on the states that would have been minted by the stuck transaction
- the stuck transactions outputs remain in the grapher, and become available for selection again once the earlier transaction that consumed them reverts on chain. Transactions are allowed to continuously dispatch using these states, since a prereq transaction being in `ReadyForDispatch` signals to a dependent it too can move to `ReadyToDispatch`

This PR does the following:
1. Add error handling and retries to `PrepareTransaction` and dispatch persistence
2. Tidy up the semantics around `ReadyForDispatch`

## Error handling
An error from `PrepareTransaction` can be caused by 
* GRPC connectivity issues/timeouts
* Unexpected error from the domain
* Expected error from the domain (this would be a bad design to allow a transaction which has already succeeded in assembly and endorsement to fail at prepare, but there is nothing to prevent a domain developer from doing this)

To achieve a balance between "likely a transient error" against "caused by a bad transaction", a `PrepareTransaction` call is now retried a configurable number of times before being reset.
### Question?
Should we actually be evicting after a certain number of failures? We don't for endorsement errors, but we do for assembly errors. All three could be signs that an originator is not functioning as expected in its assembly of transactions (e.g. proposing something unendorsable, or unpreparable). Yet its not clear what eviction achieves when an originator will immediately redelegate. On the other hand, we don't evict at the most obvious gate that an originator has gone away - repeated assemble request timeouts. I think the error handling I've added for `PrepareTransaction` in this PR is "good enough" for very rare errors and no worse than our existing error handling, but it highlights an area ripe for a follow on PR with improvements.

---

Persisting a dispatch batch is retried indefinitely since the only errors which should occur are transient database problems and these should not change the order in which we choose to submit transactions to the base ledger. The indefinite retry properties are configurable.

## `ReadyForDispatch` semantics

In the current code a transaction moves to `ReadyToDispatch` when
* It has been successfully endorsed AND
* All it's prereq transactions in `ReadyToDispatch` (or further along)

A transaction moving to `ReadyToDispatch` is immediately put on the dispatch queue, representing a point of no return, but not cleanly since
* A transaction in `ReadyToDispatch` may still be reset by a reverting prereq transaction- the dispatch loop will skip it since the transaction cannot be prepared, but the dispatch may race with the coordinator event loop unwinding the dependency chain, and dispatch the later part of the chain, even when its dependencies are not being dispatched
* A transaction can fail its prepare step, with no clean way of resetting its dependents

As a step of sequencing which can fail, and cause other transaction to reset, preparing a transaction must be done on the main coordinator event loop (or specifically, to avoid blocking latency on the loop, the work must be started by the event loop, and an event indicating its result handled on the event loop). This PR adds a new coordinator transaction state `Preparing` which spawns a goroutine that runs `PrepareTransaction` within a retry on entry to the state. A transaction transitions to `ReadyToDispatch` when the prepare completes successfully, with the transaction being added to the dispatch queue on this transition.

No work is performed in `ReadyForDispatch` now, a transaction being in this states means that
* it is on the dispatch queue, so dependent transactions can be queued behind it
* it has not yet reached the point of no return, it may still be reset by a reverting prereq

Handling `DispatchedEvent` is now the point of no return for a transaction. The dispatch loop checks that handling the event has resulted in the transaction transitioning to `Dispatched`: this indicates that the transaction had not been reset from `ReadyToDispatch`, since no other state handles this event. This approach avoids the race with a transaction unwind described above, since a transaction remains locked the entire time the unwind is cascading down its dependency chain, so in order for `HandleEvent` (which requires the transaction lock) to return, a possible unwind must be complete.

## Change to state and nullifier writing

Moving `PrepareTransaction` out of the dispatch loop and adding retries for persisting dispatch batches required some changes to how states and nullifiers are persisted.

Firstly, there can be no validation at persistence time. Any caused by creating duplicate nullifiers are now caught at the `Preparing` stage, before a transaction is past the point of no return.

Secondly, In order to retry persistence, we must know exactly which states and nullifiers were included in the the failed write. Recent changes to dispatch  (#1294)  have already moved us away from staging writes per transaction during post assembly processing, and then writing them at the next possible flush, to staging in batches in between preparing and persisting a dispatch, guaranteeing that states and nullifiers are always written alongside the dispatch of the transaction that minted them. Making dispatch persistence retriable means we need to formalise this guarantee. This PR removes the remaining domain state writer capability, which is no longer being truly exploited as a continuous, flushable write buffer, and instead includes states and nullifiers as part of the `DispatchBatch`.


matthew1001:
There's a fair amount in this PR but picking out one of the questions in the description:

> Should we actually be evicting after a certain number of failures?

You're referring to non-revert errors in the coordinator? I think we really need a very clear "evicting vs re-pooling" definition, for any other future changes to the state machine.

As you say, evicting only temporarily removes the transaction as re-delegation can happen immediately, so we should be clear about when this is the right choice. Can you propose a rule or rules for when one vs the other is the correct action?

core/go/internal/msgs/en_coordinator_transaction_states.go
```diff
@@ -25,7 +25,8 @@ var (
...
 	MsgCoordinatorTxStateBlocked                = pdm("coordinator.transaction.State_Blocked", "All endorsements have been received but the transaction cannot proceed due to dependencies not being ready for dispatch")
 	MsgCoordinatorTxStateConfirmingDispatchable = pdm("coordinator.transaction.State_Confirming_Dispatchable", "The transaction has been endorsed. Confirmation from the originator is required before the transaction can be dispatched. The originator may still request not to proceed at this point.")
-	MsgCoordinatorTxStateReadyForDispatch       = pdm("coordinator.transaction.State_Ready_For_Dispatch", "Dispatch confirmation has been received from the originator and the transaction is waiting to be collected by the dispatch goroutine")
+	MsgCoordinatorTxStatePreparing              = pdm("coordinator.transaction.State_Preparing", "Dispatch confirmation has been received from the originator and the transaction is being prepared and its dispatch built, with bounded retries, off the coordinator event loop")
```
matthew1001, line 28: Not this PR and slightly off topic but...

Trying to work out what "Dispatch confirmation has been received from the originator" means. Is that the "State_Confirming_Dispatchable" state? Do we actually have any cases where the originator rejects the dispatch request?

core/go/internal/sequencer/coordinator/transaction/dispatch.go
```diff
@@ -31,153 +31,205 @@ import (
...
-// processed before or after, never during.
-func (t *coordinatorTransaction) dispatchPrepare(ctx context.Context) error {
+// prepareAndBuildDispatch runs off the coordinator event loop. It prepares the transaction via the
+// domain, builds the transaction dispatch, resolves state distributions, and builds the nullifier
+// records validated against and linked to the states to be written. The private transaction is only
+// ever read: the prepare outputs are returned by the domain and everything built from them goes into
+// the returned pending dispatch, so nothing is written back to shared transaction state from this
+// goroutine.
+func (t *coordinatorTransaction) prepareAndBuildDispatch(ctx context.Context, pt *components.PrivateTransaction, revertCount int) (*syncpoints.PendingDispatch, error) {
 	// TODO: should this domain query context be populated with a snapshot of the domain's states at the point the transaction
```
matthew1001, line 112: Is this still a relevant TODO?

core/go/internal/sequencer/coordinator/transaction/dispatched.go
```diff
@@ -34,20 +33,14 @@ func action_NotifyDispatched(ctx context.Context, t *coordinatorTransaction, _ c
 	return t.transportWriter.SendDispatched(ctx, t.originatorNode, msg)
 }
 
-// guard_WillDispatchPublicTransaction reports whether dispatching this transaction will send a public
```
matthew1001, line 37: Do you mind a quick comment about why this guard isn't necessary any more?

core/go/internal/sequencer/coordinator/dispatch_loop.go
```diff
@@ -26,13 +26,32 @@ import (
...
+// queuedDispatch carries a transaction and its already-built pending dispatch onto the dispatch queue,
+// along with the time it was enqueued so the dispatch loop can record a metric for how long it waited
+// before being dequeued.
 type queuedDispatch struct {
 	txn        transaction.CoordinatorTransaction
+	prepared   *syncpoints.PendingDispatch
 	enqueuedAt time.Time
 }
 
+// enqueueForDispatch places a prepared transaction and its built dispatch onto the dispatch queue. It is
```
matthew1001: Some of the function commands are quite explicit about state machine specifics, which feels at risk of becoming out of date over time (e.g. I'm not sure we'll think to update this comment when we change some of the things it mentions). Just a thought on how much of the state machine behaviour/definition we want replicated elsewhere. Maybe a description without necessarily mentioning specific states as much would be a good balance of useful + less brittle.

matthew1001 (approved):
Approved but had a few questions/comments

core/go/internal/msgs/en_coordinator_transaction_states.go
annamcallister: The only condition currently that can cause it to be rejected is if originator no longer believes this is the active coordinator, which only applies to Pente today.

core/go/internal/sequencer/coordinator/transaction/dispatch.go
annamcallister: yes, I think this is a fairly long term TODO as it gets at an unresolved tension in the structure of the domain query context.

Every call made to the domain is given the full range of callbacks into core, provided via  a domain query context, including the ability to query states which may or may not be backed by a remote view.

Yet once assembly has selected states, there should be no reason for any subsequent domain calls (endorsement/prepare) to be querying states. So an ahead of chain view provided for a prepare call will almost certainly never be used, and its likely a badly designed domain if it is ever used in future.

That said, the recent changes to perform ahead of chain querying on the coordinator, not the originator, means that keeping hold of the ahead of chain view used at assembly, is now just preserving a set of pointers over a common set of backing data, not keeping hold of a lot of very large json byte arrays, so it would be more feasible to write the "just in case" code and get rid of the TODO, if you think its worth the effort.

config/pkg/pldconf/sequencermgr.go
```diff
@@ -29,6 +29,7 @@ type SequencerConfig struct {
 	CoordinatorEventQueueSize         *int               `json:"coordinatorEventQueueSize"`
 	CoordinatorPriorityEventQueueSize *int               `json:"coordinatorPriorityEventQueueSize"`
 	DelegationBatchInterval           *string            `json:"delegationBatchInterval"`
+	DispatchCommitRetry               RetryConfig        `json:"dispatchCommitRetry"`
 	DispatchMaxBatchSize              *int               `json:"dispatchMaxBatchSize"`
 	EndorseErrorRetry                 RetryConfigWithMax `json:"endorseErrorRetry"`
```
annamcallister, line 34: Since this PR was first authored, another PR added `EndorseErrorRetry` - which was renamed on PR review from `EndorseRetry`. 

Do you think we want the same change here? I think we do, as its also consistency with `SignErrorRetryThreshold` and `AssembleErrorRetryThreshold`.

core/go/internal/sequencer/coordinator/transaction/dispatched.go
annamcallister: The intermediate stages of dispatch, including the prepared transaction, used to be stored on the transaction, but this refactor instead includes the prepared transaction as part of the struct that gets put on the dispatch loop queue, this means that the dispatch loop goroutine never reads/writes the transaction keeping concurrency with the event loop goroutine more straightforward and avoiding the need for locks.

This means that a guard can no longer check whether a public transaction is going to be dispatched just by reading this intermediate storage on the transaction. Instead the result of dispatch is included in the dispatch event, which an action can read, but not a guard.

So it's not unnecessary anymore, just relocated.

PR #1279, opened 2026-06-28
Java SDK: PTX and KeyManager RPC namespace clients

divyam-jha123:
## Summary

Implements the first two RPC-namespace clients for the Java SDK on top of the
HTTP JSON-RPC transport from #1231 — `KeyManagerClient` (`keymgr_*`) and
`PtxClient` (`ptx_*`, private transaction manager). Ports the Go reference
clients `sdk/go/pkg/pldclient/keymgr.go` and `ptx.go`, with full method parity
for the HTTP request/response surface and a unit test per method using mock
HTTP responses.

Part of the Java SDK RFC #1224 (enhancement #4).

## Parts

The change is split into independently-reviewable commits.

### Part 1 — KeyManager (`keymgr_*`)
`KeyManagerClient` covering all 6 methods: `wallets`, `resolveKey`,
`resolveEthAddress`, `reverseKeyLookup`, `queryKeys`, `sign`. Adds the
`core/key` response models (`KeyMappingAndVerifier`, `KeyQueryEntry`,
`KeyVerifier`, `KeyPathSegment`), flattened to the inline JSON wire form.
Promotes the `MockJsonRpcServer` test harness to `public` so namespace-client
tests in any package can reuse it. 7 unit tests.

### Part 2 — PTX transaction lifecycle (`ptx_*`)
`PtxClient` with the lifecycle surface: `sendTransaction(s)`,
`prepareTransaction(s)`, `updateTransaction`, `call`, `getTransaction`,
`getTransactionFull`, `getTransactionByIdempotencyKey`, `queryTransactions`,
`queryTransactionsFull`. Adds `core/transaction` types `Transaction`,
`TransactionFull` (extends `Transaction`), and `TransactionCall` (input,
unwraps `TransactionInput` via `@JsonUnwrapped`). 12 unit tests.

### Part 3 — PTX receipts, state, prepared, public *(planned)*
`getTransactionReceipt(Full)`, `getStateReceipt`, `getDomainReceipt`,
`getPreparedTransaction`, `queryPreparedTransactions`,
`queryTransactionReceipts`, `getPublicTransaction`. Adds
`TransactionReceiptFull`, `TransactionStates`, `PreparedTransaction`,
`PublicTxWithBinding`.

### Part 4 — PTX ABI / decode / resolve / listeners / dispatch *(planned)*
`storeABI`/`getStoredABI`/`queryStoredABIs`, `decodeError`/`decodeCall`/
[...]

Issue #1242, opened 2026-06-16
Enhancement: Java SDK - testing module (mock client, WireMock, Testcontainers)


divyam-jha123:
### What would you like to be added?

Implement the `testing` module that gives downstream Java applications first-class
test support for code that talks to a Paladin node — without standing up a real node
for every unit test.

### Reference implementations
- `sdk/go/pkg/wsclient/wstestserver.go` — in-repo test-server pattern
- Existing `testing` Gradle module scaffolded in #1226

### Scope
- [ ] Mock / in-memory Paladin client for unit tests (no network)
- [ ] WireMock stubs (or equivalent) for HTTP JSON-RPC, with reusable fixtures
- [ ] Testcontainers support to spin up a Paladin node for higher-fidelity tests
- [ ] Helper assertions / fixtures for common types (receipts, transactions)
- [ ] Self-tests demonstrating the mock client and stubs

### Out of scope
The end-to-end integration test suite that runs against a real node (separate Phase 3 issue).

### Acceptance criteria
- A consumer can write a unit test against the mock client with no running node
- WireMock stubs cover the core RPC happy paths
- Testcontainers support starts a node and is exercised by at least one smoke test

### Why is this needed?

Part of the Java SDK RFC #1224

PR #1321, opened 2026-08-07
fix(java-sdk): type keymgr_wallets as WalletInfo objects

### LINKED ISSUE #1320: Java SDK: keymgr_wallets typed as List<String> but node returns WalletInfo objects
### Summary

`KeyManagerClient.wallets()` in the Java SDK declares `CompletableFuture<List<String>>`, but
`keymgr_wallets` returns a list of wallet **objects**. Any call against a live node fails to
deserialize.

### Reproduce

Against a local devnet node (`operator/paladin-kind.yaml`, node 1 on `localhost:31548`):

```sh
curl -s -X POST http://localhost:31548 -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"keymgr_wallets","params":[]}'
```

```json
{"jsonrpc":"2.0","id":1,"result":[{"name":"signer-1","keySelector":".*","keySelectorMustNotMatch":false}]}
```

Calling the same method through the SDK:

```
PaladinRpcException: failed to parse result: Cannot deserialize value of type
`java.lang.String` from Object value (token `JsonToken.START_OBJECT`)
 (through reference chain: java.util.ArrayList[0])
```

### Root cause

The wallet shape is a three-field object — the wallet's name plus the key selector that
decides which key identifiers it claims (`keySelectorMustNotMatch` inverts the match). It
is defined in Go as `pldapi.WalletInfo` (`sdk/go/pkg/pldapi/keymgr.go`) and configured via
`pldconf.WalletConfig` (`config/pkg/pldconf/keymanager.go`). The Java SDK has no equivalent
model, so `wallets()` was typed against a shape the node never sends.

### Why CI did not catch it

`KeyManagerClientTest.wallets()` stubbed the RPC response as `["w-a","w-b"]` — a fabricated
payload that matched the client's incorrect assumption rather than the real wire format. The
test passed, so the mismatch survived review and merged in #1279.

More generally: the Java SDK currently has **no test that talks to a real node**. Every test
is a unit test over hand-written fixtures, so any fixture that encodes a wrong assumption
about the wire format is invisible to CI. This was found by pointing a demo client at a
running devnet, not by the test suite. It is an argument for prioritizing the
Testcontainers/integration-test work (#1242).

### Impact

- `wallets()` is unusable against any real node.
- Fixing it changes the method's return type, so it is a **breaking change** to a public API
  — worth landing before `0.1.0` rather than after.

### Fix

Add a `WalletInfo` model to `sdk/java/core` (`org.lfdt.paladin.sdk.core.key`) and retype
`wallets()` to `CompletableFuture<List<WalletInfo>>`, with the client test stubbing the shape
a node actually returns.

Affected since #1279 (`java-sdk-main`). Part of the Java SDK effort tracked by #1224.

divyam-jha123:
Fixes #1320.

### What

`KeyManagerClient.wallets()` declared `CompletableFuture<List<String>>`, but `keymgr_wallets`
returns a list of wallet objects. Every call against a live node failed with:

```
PaladinRpcException: failed to parse result: Cannot deserialize value of type
`java.lang.String` from Object value (token `JsonToken.START_OBJECT`)
```

This adds the missing `WalletInfo` model and retypes the method.

### Changes

| File | Change |
|------|--------|
| `core/.../key/WalletInfo.java` | New — immutable model for `name` / `keySelector` / `keySelectorMustNotMatch` |
| `core/.../key/WalletInfoTest.java` | New — 4 tests, including one pinned to the literal payload a devnet node returns |
| `client/.../keymgr/KeyManagerClient.java` | `wallets()` retyped to `List<WalletInfo>` |
| `client/.../keymgr/KeyManagerClientTest.java` | Stub corrected to the real wire shape |

### On the test that missed this

The existing `KeyManagerClientTest.wallets()` stubbed the response as `["w-a","w-b"]` — a
fabricated payload matching the client's incorrect assumption rather than the node's actual
output. It passed, which is why the mismatch merged in #1279.

The replacement stubs two wallets in the real shape, including one with
`keySelectorMustNotMatch: true` to cover the inverted-selector case. `WalletInfoTest` also
pins the exact JSON captured from a running devnet node, so a future change to the wire
format fails a test rather than a demo.

### Breaking change

`wallets()` changes return type from `List<String>` to `List<WalletInfo>`. The old signature
never worked against a real node, so nothing functional depends on it — but it is a public
API change, and better landed before `0.1.0`.

### Testing

`./gradlew :sdk:java:core:build :sdk:java:client:build` — green, including Spotless,
Checkstyle, Javadoc doclint and JaCoCo.

- `WalletInfoTest` — 4 tests, 0 failures
- `KeyManagerClientTest` — 7 tests, 0 failures

Verified manually against a local 3-node devnet: `wallets()` now returns
`signer-1 selector=.*`.

### Follow-up

This was found by pointing a client at a real node, not by CI — the SDK has no test that
talks to one. Noted in #1320 as an argument for prioritizing #1242.

### DIFF
--- a/sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/key/WalletInfo.java
+++ b/sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/key/WalletInfo.java
@@ -25,6 +25,11 @@ import java.util.Objects;
  * <p>A wallet claims key identifiers by matching them against its key selector, a regular
  * expression. When {@link #keySelectorMustNotMatch()} is set the sense is inverted: the wallet
  * claims identifiers that do <em>not</em> match. Immutable.
+ *
+ * <p>Selectors are evaluated on the node against <a href="https://golang.org/s/re2syntax">RE2
+ * syntax</a>, not {@link java.util.regex}. RE2 has no backreferences ({@code \1}) and no lookahead
+ * or lookbehind ({@code (?=...)}, {@code (?<!...)}), so a selector using them is rejected by the
+ * node even though it would compile as a Java pattern.
  */
 @JsonPropertyOrder({"name", "keySelector", "keySelectorMustNotMatch"})
 public final class WalletInfo {
@@ -54,7 +59,8 @@ public final class WalletInfo {
   }
 
   /**
-   * The regular expression used to select which key identifiers this wallet holds.
+   * The regular expression used to select which key identifiers this wallet holds, in <a
+   * href="https://golang.org/s/re2syntax">RE2 syntax</a>.
    *
    * @return the key selector expression
    */
--- a/sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/key/WalletInfo.java
+++ b/sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/key/WalletInfo.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright contributors to Paladin, an LFDT project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.lfdt.paladin.sdk.core.key;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import java.util.Objects;
+
+/**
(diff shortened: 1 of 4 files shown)


sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/key/WalletInfo.java
```diff
@@ -0,0 +1,97 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import java.util.Objects;
+
+/**
+ * One wallet configured on the node, as returned by {@code keymgr_wallets}.
+ *
+ * <p>A wallet claims key identifiers by matching them against its key selector, a regular
```
matthew1001, line 25: I wonder if it's worth mentioning in the Java doc that Paladin uses https://golang.org/s/re2syntax as its regex syntax, so Java SDK users should ensure their selectors are compatible with that syntax

matthew1001 (approved):
Approved with one minor comment @divyam-jha123

sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/key/WalletInfo.java
divyam-jha123: added the javadoc

matthew1001 (approved):
Thanks for the new comment 👍 LGTM

PR #1326, opened 2026-08-13
full redelegation after switching coordinator

annamcallister:
This regressed with #1299 and has been causing intermittent test failures where transactions are slow to confirm because they are not redelegated until they are missing from a heartbeat.


core/go/internal/sequencer/originator/delegation.go
```diff
@@ -167,27 +167,35 @@ func sendDelegationRequest(ctx context.Context, o *originator, full bool) error
...
 
-// action_NotifyPartialDelegation indicates to the delegation batching loop that a partial delegation
+// requestPartialDelegation indicates to the delegation batching loop that a partial delegation
 // (only transactions State_Pending / State_Delegated) will be required on its next tick. Sending all
 // transactions in these states is required to preserve FIFO ordering from this originator until first
-// assembly. o.notifyPartialDelegation has length 1, so that multiple notfications result in a single
+// assembly. o.notifyPartialDelegation has length 1, so that multiple notifications result in a single
 // delegation request.
-func action_NotifyPartialDelegation(_ context.Context, o *originator, _ common.Event) error {
+func (o *originator) requestPartialDelegation() {
```
matthew1001, line 175: I assume this was broken out into a separate function for symmetry with `requestFullDelegation()`, even though it's not called from anywhere else but `action_NotifyPartialDelegation`?

matthew1001 (approved):

core/go/internal/sequencer/originator/delegation.go
annamcallister: 👍

Issue #942, opened 2025-12-04
bug: Review conversation around handling endorsements, particularly the comment highlighting the need for clarification on the correct behaviour


matthew1001:
### What happened?

See PR review comment https://github.com/LFDT-Paladin/paladin/pull/704#discussion_r2586614050

Refers to this comment https://github.com/LFDT-Paladin/paladin/blob/main/core/go/internal/sequencer/coordinator/transaction/endorsing.go#L47

### What did you expect to happen?

N/A

### How can we reproduce it (as minimally and precisely as possible)?

N/A

### Anything else we need to know?

N/A

### OS version

N/A

annamcallister:
Moving to V1.1 as this TODO needs addressing.

Use of hashes should be looked at across the sequencer. Where is it being calculated and where is it being used - e.g. there is an originator side check which is _not_ use to reject a dispatch when it could be.

PR #1328, opened 2026-08-19
Handle endorsement reverts correctly in sequencer

annamcallister:
A revert result from endorsement is not currently being handled by the coordinator or coordinator transaction state machine. This hasn't been noticeable so far since:
- no domain currently returns a revert result from endorsement, only errors - this needs fixing in the individual domains
- it should be impossible to revert at endorsement time if all nodes are running correct domain code, since an assemble should never propose a transaction that cannot be successfully endorsed. Endorsement is a protection against a proposed invalid transaction from a bad actor assembling

This PR adds the following semantics around endorsement reverts and endorsement unexpected errors:

Where a transaction requires K of N endorsements:
 - If a transaction receives `N-K+1` endorsement reverts, the transaction is finalised as failed with the revert reasons
 - If a transaction receives `N-K+1` unexpected endorsement errors, the transaction is repooled
 - If a transaction receives a mixture of endorsement reverts and unexpected endorsement errors totalling `N-K+1`, the transaction is repooled
 - If a transaction receives `K` successful endorsements and `N-K` endorsement reverts or unexpected errors the transaction has met the endorsement threshold, despite the reverts, and can proceed to dispatch
 
Today there are no domains which support anything other than 100% endorsement. `N-K+1` is always `1`, meaning a single revert or single errors causes a finalisation/repool respectively, and the combined logic is never exercised.
 
This change prevents an unendorsable transaction from occupying an inflight coordinator slot indefinitely. It is not a fairness algorithm, and doesn't prevent a bad actor from repeatedly delegating unendorsable transactions that take up coordinator resource.

---
This PR also fixes a bug where some transaction finalisations weren't passing in the contract address. This meant that all finalisations from a contract wouldn't necessarily get assigned to the same worker in the flush writer. Ordering of off chain reverted transactions is not significant, so this had no meaningful impact, but it's better to fix it up regardless.


matthew1001 (dismissed):
This feels like a substantial assertion that's worth documenting in the distributed sequencer docs. I think the PR is asserting that endorsement can't revert for a transient issue (i.e. re-assemble would never produce an endorseable transaction). As such we should probably make sure that's clearly described in the docs beyond just the state machine changes

annamcallister:
@matthew1001 I've requested a new review as I've added a few changes to this once I pulled on the thread we discussed in person of retrying unexpected endorsement errors. Specifically
- more rigorous checking of idempotency keys, so that we do not double handle endorsement requests, or apply results from a response we weren't expecting
- removing `uuid.MustParse` from the transport client (and equivalent for the one place we do bytes32 parsing)- so that one node's malformed UUID cannot cause another node to crash.

I've made a note on the longer list of things I'm tracking to update docs with. I think as discussed there's a more detailed doc site page to come about all the states a transaction goes through, error semantics and retries- with the changes in #1315  and #1317 needing docs too.

annamcallister:
And an additional commit to bring coverage back to 100%, and stop it slowly dropping undetected because of rounding, until one PR pushes it over the edge

core/go/internal/sequencer/coordinator/transaction/endorsing.go
```diff
@@ -280,9 +270,9 @@ func guard_EndorseFailureExceedsTolerance(_ context.Context, txn *coordinatorTra
 }
 
 // guard_EndorseRevertExceedsTolerance returns true if the number of endorsement reverts for any
-// single attestation requirement now exceeds its tolerance. That means the threshold can no longer
-// be reached and, unlike guard_EndorseFailureExceedsTolerance, the reverts will recur: repooling
-// would reassemble and collect the same ones again. The transaction is finalized instead.
+// single attestation requirement now exceeds its tolerance, meaning the threshold can no longer be
+// reached. A correctly implemented domain does not assemble a transaction that its own endorsers
```
matthew1001, line 274: 👍

core/go/internal/sequencer/coordinator/endorsing.go
```diff
@@ -147,44 +147,110 @@ func action_AddEndorsementRequestSenderToEndorserCandidates(ctx context.Context,
...
+	}
 	endorseCtx := ctx
 	cancel := func() {}
 	if !e.Expiry.IsZero() {
 		endorseCtx, cancel = context.WithDeadline(ctx, e.Expiry)
 	}
 	go func() {
-		defer cancel()
 		c.handleEndorsementRequest(endorseCtx, e)
+		c.endEndorsement(e.IdempotencyKey)
```
matthew1001, line 177: If this go routine fails, does this expose a memory leak on the `inFlightEndorsements` map?

config/pkg/pldconf/sequencermgr.go
```diff
@@ -83,6 +84,7 @@ var SequencerDefaults = SequencerConfig{
 	CoordinatorPriorityEventQueueSize: confutil.P(500),
 	DelegationBatchInterval:           confutil.P("50ms"),
 	DispatchMaxBatchSize:              confutil.P(100),
+	EndorseRetry:                      *GenericRetryDefaults,
```
matthew1001: For consistency, and to help avoid confusion between "endorse failed" and "endorse reverted", should we:

1. Change this to `EndorseErrorRetry` (similar to `AssembleErrorRetryThreshold`)
2. Change `AssembleErrorRetryThreshold` -> `AssembleErrorRetry` of type `GenericRetryDefaults`?
   - If we think we're safe to assume nobody has actually set `AssembleErrorRetryThreshold` and is currently relying on the default behaviour

core/go/internal/sequencer/coordinator/transaction/endorsing.go
```diff
@@ -35,28 +35,18 @@ type endorsementRequirement struct {
...
+// has matched to the request still outstanding for that attestation requirement and party, which is
+// also what makes the verifier safe to dereference here.
+func (t *coordinatorTransaction) applyEndorsement(ctx context.Context, endorsement *prototk.AttestationResult) error {
+	log.L(ctx).Debugf("endorsement '%s' received for transaction %s from %s", endorsement.Name, t.pt.ID, endorsement.Verifier.Lookup)
+	delete(t.pendingEndorsementRequests[endorsement.Name], endorsement.Verifier.Lookup)
+	t.pt.PostAssembly.CollectedEndorsements = append(t.pt.PostAssembly.CollectedEndorsements, endorsement)
+
+	// MRW TODO - Hashing the TX for dispatch confirmation requires that there is > 0 signatures. Need to follow up where an endorsed TX populates the signatures. Temporarily put this workaround in.
+	// log.L(ctx).Infof("Applying endorsement. Appending %+v to list of endorsements received for transaction %s from %s", endorsement, t.pt.ID, endorsement.Verifier.Lookup)
+	// t.pt.PostAssembly.Signatures = append(t.pt.PostAssembly.Signatures, endorsement)
```
matthew1001, line 49: This stands out as a "we ought to address this or decide we're fine without it" in 1.1. Mind taking a look to decide if a new issue is worth opening to address is specifically (or if it can be removed safely here)?

core/go/internal/sequencer/coordinator/transaction/state_machine.go
```diff
@@ -619,19 +620,21 @@ var stateDefinitionsMap = StateDefinitions{
 					},
 				}},
 			},
-			// Domain returned REVERT: the endorser rejected the assembly as invalid.
+			// Domain returned REVERT: the endorser rejected the assembly as invalid. A correctly
+			// implemented domain does not assemble a transaction that its own endorsers would
+			// revert, so a revert says the transaction cannot be executed.
```
matthew1001, line 625: I don't have the suggestion option for some reason, but I suggest this wording tweak:

"so a revert says the transaction cannot be executed even if it was reassembled."

core/go/internal/sequencer/coordinator/endorsing.go
annamcallister: It can't fail since no function returns an error and there is no early exit.

Two deferred functions did not read well, and didn't provide any extra guarantee of those functions running.

core/go/internal/sequencer/coordinator/transaction/state_machine.go
annamcallister: I was quite keen to drop the "even if reassembled" type wording as I don't believe its accurate. An incorrectly implemented domain could reassemble to a completely different result (e.g. selecting valid coins through chance or intention, where it didn't on the first assemble).

Maybe its a "must not be executed" rather than a "cannot"?

core/go/internal/sequencer/coordinator/transaction/endorsing.go
annamcallister: We have an issue for this- I've added it to the V1.1 board https://github.com/LFDT-Paladin/paladin/issues/942

config/pkg/pldconf/sequencermgr.go
annamcallister: I've done 1. 

2. I don't think makes sense. Assemble error retries have very different semantics as it's the number of times that a transaction receiving an error assemble response from an originator, may be repooled and reselected. This cannot be performed within a `retry.Do` function and the config type doesn't make sense - there is no timed delay, the retry timer is a function of how many other transactions are in the pool and how long they take to assemble.

core/go/internal/sequencer/coordinator/transaction/state_machine.go
matthew1001: Yeah I think the it's just a little unclear what the comment is trying to say. "executed" isn't a term we've just a lot wrt Paladin transactions - it's most about "submissions", "dispatches" etc. so I just wasn't sure what "executed" meant. It's not a big thing, so whatever you think is clear I'm happy to go with.

matthew1001 (approved):

PR #1329, opened 2026-08-19
Pente return revert result for endorse

annamcallister:
Update the errors returned during in endorsement to distinguish between
* A revert result- the domain does not endorse this transaction
* Error - an unexpected error occurred during endorsement, this does not tell us anything about the validity of the transaction being endorsed

#1328 updates the sequencer to handle these reverts correctly

This PR also makes three changes to the endorse error handling to support being more specific in the return type.

1. A reverting EVM execution was returning `SIGN` - this would result in a failed sign request later as there is no attestation to sign
2 `JsonProcessingException` was falling into the generic catch but this is a deterministic failure to parse the info state and should not be retried.
3. A plugin error is marked by assembly as permanent. Endorse now checks for this and returns such errors as a revert.

### DIFF
--- a/domains/pente/src/main/java/io/kaleido/paladin/pente/domain/PenteDomain.java
+++ b/domains/pente/src/main/java/io/kaleido/paladin/pente/domain/PenteDomain.java
@@ -439,12 +439,37 @@ import com.fasterxml.jackson.core.JsonProcessingException;
                      setPayload(ByteString.copyFrom(endorsementPayload)).
                      build());
          } catch (PenteEVMTransaction.EVMExecutionException e) {
+             // Either the EVM reverted, or the account state supplied with the request did not match
+             // what the transaction expects (e.g. a stale nonce): Both are REVERT results.
              LOGGER.error("EVM execution failed during endorsement", e);
              return CompletableFuture.completedFuture(EndorseTransactionResponse.newBuilder().
-                     setEndorsementResult(EndorseTransactionResponse.Result.SIGN).
+                     setEndorsementResult(EndorseTransactionResponse.Result.REVERT).
                      setRevertReason(e.getMessage()).
                      build());
+         } catch (JsonProcessingException | IllegalArgumentException | IllegalStateException e) {
+             // The request was malformed, the info state carrying the transaction would not parse, or
+             // re-executing it did not reproduce the proposed inputs and outputs. Deterministic in every
+             // case: asking us again with the same assembly gets the same answer, so the result is revert.
+             LOGGER.error("Transaction not endorsed", e);
+             return CompletableFuture.completedFuture(EndorseTransactionResponse.newBuilder().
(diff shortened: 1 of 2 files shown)


domains/pente/src/main/java/io/kaleido/paladin/pente/domain/PenteDomain.java
```diff
@@ -439,12 +439,37 @@ protected CompletableFuture<EndorseTransactionResponse> endorseTransaction(Endor
                      setPayload(ByteString.copyFrom(endorsementPayload)).
                      build());
          } catch (PenteEVMTransaction.EVMExecutionException e) {
+             // Either the EVM reverted, or the account state supplied with the request did not match
+             // what the transaction expects (e.g. a stale nonce): Both are REVERT results.
              LOGGER.error("EVM execution failed during endorsement", e);
              return CompletableFuture.completedFuture(EndorseTransactionResponse.newBuilder().
-                     setEndorsementResult(EndorseTransactionResponse.Result.SIGN).
+                     setEndorsementResult(EndorseTransactionResponse.Result.REVERT).
```
matthew1001, line 446: So to recap where I think we landed on the the general design re. endorsement failures
 - It's not expected for an endorse step to fail for a reason that wouldn't have been caught at assembly time, unless the domain has been badly coded or there is some other fundamental code bug in Paladin
 - So endorse is considered to be either successful, or revert - not retry.

domains/pente/src/main/java/io/kaleido/paladin/pente/domain/PenteDomain.java
```diff
@@ -439,12 +439,37 @@ protected CompletableFuture<EndorseTransactionResponse> endorseTransaction(Endor
...
+                 // be revert.
+                 LOGGER.error("Error response from plugin during endorsement", e);
+                 return CompletableFuture.completedFuture(EndorseTransactionResponse.newBuilder().
+                         setEndorsementResult(EndorseTransactionResponse.Result.REVERT).
+                         setRevertReason(e.getMessage()).
+                         build());
+             }
+             return CompletableFuture.failedFuture(e);
          } catch (Exception e) {
+             // Anything else (IO, interrupt, class loading) may be transient, so it is reported as an
```
matthew1001, line 471: I do wonder if we're happy with the exception -> revert/retry classifications here:

```
catch (JsonProcessingException | IllegalArgumentException | IllegalStateException e)
```

 - `JsonProcessingException` seems OK
 - `IllegalArgumentException` less so
 - `IllegalStateException` even less so

One example for the `IllegalArgumentException` is `PersistedAccount` which does:

```
} catch (Exception e) {
   throw new IllegalArgumentException(e);
}
```

but I think more generally `IllegalStateException` is a very broad exception that might include transient conditions.

Did you hit these last 2 specifically during testing and want them included for specific reasons? If not then perhaps we leave `IllegalStateException` to fall into the retry camp?

PR #1331, opened 2026-08-19
Statemgr cleanup

annamcallister:
Two commits to clean up the internals and externals of state manager and the state store.

---
### Commit 1: remove UUID support- domain query contexts no longer need to close

External JSON RPC calls to query the state store claimed the ability to specify a transaction UUID as the status qualifier parameter. This might have made sense in an earlier architecture but is certainly meaningless in the current state tracking/optimistic state tracking model. States stopped being locked on transactions in 10/2024, ahead of the initial release. The remaining implementation suggested a wiring into a domain context, however since the V1 release, ahead of chain state views have been stored in the grapher not the domain context. A queryable ahead of chain view would be achievable on the coordinator by querying via the grapher (similar to how available states are found for assembly), but since this feature has been clearly broken for a long time, and the value of querying a rapidly changing view which may or may not be confirmed on chain feels limited, I have chosen simply to fully remove at this point in time.

There had been some discussion of removing either the `confirmed` or 'available` status qualifier as they are identical in meaning, whereas in an earlier but now replaced architecture, where an confirmed but unspent state could be attached to a sequencer, thereby rendering it unavailable, they had different meanings. I have chosen not to as they still feel both like meaningful ways of describing the same state status, and instead call them out more explicitly as synonyms.

---
### Commit 2: refine external state store interface

These changes came about as I was finding it hard to make what should have been a simple change to the state manager. The knock on effects were repeatedly more impactful and problematic then they should have been. This package has had a lot of changes in functionality recently; this commit

* Remove external functions (on the interface and implementation) which do not have any callers
* Remove the external state query options from components which weren't actually used externally
* Instead of having an options object where the potential permutations increase exponentially with each newly added option, refactor the three permutations of the options that are actually used into three separate internal functions whose names self describe their purpose, layered onto a common base function
* Rename functions which claim to be finding nullifiers to be more explicit that they are finding states backed by nullifiers

### DIFF
--- a/core/go/internal/statemgr/domain_query_context.go
+++ b/core/go/internal/statemgr/domain_query_context.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"maps"
 	"strings"
-	"sync"
 
 	"github.com/LFDT-Paladin/paladin/common/go/pkg/i18n"
 	"github.com/LFDT-Paladin/paladin/common/go/pkg/log"
@@ -61,36 +60,28 @@ func createLogContext(ctx context.Context, domainName string, contractAddress pl
 	return ctx
 }
 
-// Short-lived, registered in the state manager. Always closed by the caller via defer dqc.Close(ctx).
+// Short-lived, and holds no resources of its own - it is collected once its consumer drops it.
 // May carry a remote view (spend exclusions + on-demand state queries) for FindAvailableStates queries.
 type domainQueryContext struct {
 	ss                 *stateManager
 	domainName         string
 	customHashFunction bool
 	contractAddress    pldtypes.EthAddress
-	stateLock          sync.Mutex
-	id                 uuid.UUID
-	closed             bool
+	id                 uuid.UUID // correlates this context's log lines
 	remoteStateView    components.RemoteStateView
 }
 
-// Very important that callers Close domain query contexts they open.
 func (ss *stateManager) NewDomainQueryContext(ctx context.Context, domain components.Domain, contractAddress pldtypes.EthAddress) components.DomainQueryContext {
 	id := uuid.New()
 	log.L(ctx).Debugf("Domain context %s for domain %s contract %s created", id, domain.Name(), contractAddress)
 
-	ss.domainContextLock.Lock()
-	defer ss.domainContextLock.Unlock()
-
-	dqc := &domainQueryContext{
+	return &domainQueryContext{
 		ss:                 ss,
 		domainName:         domain.Name(),
 		customHashFunction: domain.CustomHashFunction(),
 		contractAddress:    contractAddress,
 		id:                 id,
 	}
-	ss.domainContexts[id] = dqc
-	return dqc
 }
 
 // NewDomainQueryContextWithRemoteView creates a domain query context whose FindAvailableStates
@@ -102,10 +93,7 @@ func (ss *stateManager) NewDomainQueryContextWithRemoteView(ctx context.Context,
 	id := uuid.New()
 	log.L(ctx).Debugf("Assembly domain context %s for domain %s contract %s created", id, domain.Name(), contractAddress)
 
-	ss.domainContextLock.Lock()
-	defer ss.domainContextLock.Unlock()
-
-	dqc := &domainQueryContext{
+	return &domainQueryContext{
 		ss:                 ss,
 		domainName:         domain.Name(),
 		customHashFunction: domain.CustomHashFunction(),
@@ -113,8 +101,6 @@ func (ss *stateManager) NewDomainQueryContextWithRemoteView(ctx context.Context,
 		id:                 id,
 		remoteStateView:    remoteStateView,
 	}
-	ss.domainContexts[id] = dqc
-	return dqc
 }
 
 // getSpentStateIDs returns the remote view's spend exclusion set. Returns nil for local-only contexts.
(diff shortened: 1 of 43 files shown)


sdk/go/pkg/pldapi/states.go
```diff
@@ -44,17 +44,27 @@ func (st SchemaType) Options() []string {
...
-// Note this is not modelled as a normal Paladin Enum, as you can fall back to a UUID.
+// Note this is not modelled as a normal Paladin Enum, as the set of qualifiers is open to
+// extension with forms that are not a fixed list of options - such as a reference to a
+// particular point in the history of the chain.
 type StateStatusQualifier string
 
+// States with a confirmation record, and no spend record
 const StateStatusAvailable StateStatusQualifier = "available"
+
+// Synonym for StateStatusAvailable - a state is confirmed for use only while it is also unspent
```
matthew1001, line 58: Reading your thoughts in the PR description, I think I'd make a case for removing one of these. So just leaving this comment to at least have that discussion before merging.

core/go/internal/statemgr/state.go
```diff
@@ -318,144 +317,106 @@ func (ss *stateManager) labelSetFor(schema components.Schema) *trackingLabelSet
...
-	return dqc.FindAvailableStates(ctx, dbTX, schemaID, jq)
+	status pldapi.StateStatusQualifier,
+) (components.Schema, []*pldapi.State, error) {
+	scope := statusScope(ctx, dbTX, status, nil)
+	return ss.findStatesCommon(ctx, dbTX, domainName, contractAddress, schemaID, jq,
+		func(_ persistence.DBTX, q *gorm.DB) *gorm.DB { return scope(q) })
 }
 
-func (ss *stateManager) findNullifiers(
+// findStatesForRemoteViewMerge reads states for a domain context that has a remote view, ready to be
```
matthew1001, line 362: "reads states" here means "reads remote states"?

And "merged with the view's own matches" is presumably, by implication, the "local" states? With the introduction of "remote" as a concept in the state manager, I do wonder if we benefit from starting to use "local" explicitly in some places. It's sort of implicit - but only when you really know the codebase. 

So I also wonder if this should be called `findLocalStatesToMergeWithRemoteView`? But is that an accurate description?

core/go/internal/statemgr/domain_query_context.go
```diff
@@ -214,15 +133,11 @@ func (dqc *domainQueryContext) fetchRemoteViewStates(ctx context.Context, schema
 	return queried, nil
 }
 
-// startRemoteViewFetch launches the remote view query (when a remote view is attached) concurrently
-// with the caller's local DB read, and returns a wait function that blocks until the fetch completes
-// and returns its results. All failures — including a query that cannot be re-marshaled for the
-// round-trip — are reported through the wait function, so callers have a single error path. With no
-// remote view attached the wait function is an immediate no-op.
+// startRemoteViewFetch launches the remote view query concurrently with the caller's local DB read,
```
matthew1001, line 136: Could we add a little to this comment about the intended retry semantics for fetching remote states? It would be useful to future contributors just to touch on it here. Do we fail assembly but not revert, and rely on being re-pooled?

doc-site/docs/reference/types/_includes/statestatusqualifier_description.md
```diff
@@ -1,8 +1,7 @@
 The following qualifiers can be used in queries to the state store:
 
 - `available` - states that have been confirmed by a blockchain transaction, and not yet been spent in a subsequent transaction
+- `confirmed` - a synonym of `available`: a state is only confirmed for use while it also remains unspent
```
matthew1001, line 4: I left a comment elsewhere about whether we keep these, but I do think it's worth some more discussion reading these descriptions again. It feels like `available` should be a subset of `confirmed`, i.e. those states which are both `confirmed` and not `spent`.

core/go/internal/statemgr/domain_query_context.go
matthew1001, line 107: I've got a slight ick about just doing a `nil` check as the way of deciding how we're retrieving states. Feel free to ignore, but if there's a cleaner way that removes the `nil` check as the way of implementing a valid choice of code path then I wouldn't be averse to it :)

sdk/go/pkg/pldapi/states.go
annamcallister: Discussed in person that we are removing available

core/go/internal/statemgr/state.go
annamcallister: I don't think the name is wrong here. There are a set of three functions that read from this nodes database, which are all a variant on finding states.

I've updated the comment to hopefully make the relationship with the remote view clearer, as well as the reason why merging requires a different, more expensive format to be returned.

core/go/internal/statemgr/domain_query_context.go
annamcallister: Comment updated

doc-site/docs/reference/types/_includes/statestatusqualifier_description.md
annamcallister: resolved above

core/go/internal/statemgr/domain_query_context.go
annamcallister: I don't think the nil check is out of place here- one is used if provided, and the nil check is asking exactly that in a fairly idiomatic way for go.

What I have done is rename the function to include Remote- because that is the only way you get spent IDs

doc-site/docs/reference/types/_includes/statestatusqualifier_description.md
```diff
@@ -1,7 +1,6 @@
 The following qualifiers can be used in queries to the state store:
 
-- `available` - states that have been confirmed by a blockchain transaction, and not yet been spent in a subsequent transaction
-- `confirmed` - a synonym of `available`: a state is only confirmed for use while it also remains unspent
 - `unconfirmed` - states where no transaction has yet been processed from the blockchain to confirm the state
+- `confirmed` - states that have been confirmed by a blockchain transaction, and not yet been spent in a subsequent transaction. A state is only confirmed for use while it also remains unspent.
```
matthew1001, line 4: Think this is so much clearer 👍

matthew1001 (approved):
LGTM, thanks for the discussion on the direction to go with re. available/confirmed

PR #1332, opened 2026-08-19
Sort label optimisation

annamcallister:
Addresses a TODO about always preloading labels. 

Initially I followed the improvement in the TODO exactly, implementing sort label projection. Then I realised that was a lot of code when no domain currently sorts on anything other than created, which is already available on a state queried from the DB.

So the actual code in this PR drops the preload AND doesn't implement sort label projection, which is the most efficient code path for the current set of domains, and probably also for most domains in the future. It makes a lot of sense to choose the oldest or the newest available states. Sorting by numeric value could make sense (e.g. assemble by adding together the smallest available coins), sorting by alphabetical value is getting strange. Maybe other timestamps in there could make sense too. Sorting by a value other than created remains functionally supported via recovering the labels (which is an expensive ABI parse), and can be enhanced with label projection if its ever actually needed in the future.

This PR also adds some metrics which have been useful for assessing the impact of making this change, and for some follow on changes in this area.


core/go/internal/statemgr/metrics/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var METRICS_SUBSYSTEM = "state_manager"
+
+// operationBuckets are the second-scale boundaries shared by the state-query histograms, on a 1-2.5-5
+// ladder from 100us to 10s.
+//
+// They are hand-picked rather than generated by prometheus.ExponentialBuckets, because a uniform factor
```
matthew1001, line 31: This is very detailed but I'm not it needs to be quite so in-depth. Conscious of us having more comments than code 😆  Can we slim this down to a couple of lines?

core/go/internal/statemgr/metrics/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+// reported as a number at all, so the range runs well past where these queries are expected to sit,
+// keeping an abnormally slow one on the scale rather than lumping it in with everything else that
+// overflowed. Ten seconds of work should be off the scale.
+//
+// These are the same boundaries the distributed_sequencer operation histograms use, so a query timing
+// can be compared bucket for bucket with the domain call that encloses it.
+var operationBuckets = []float64{0.0001, 0.00025, 0.0005, 0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10}
+
+type StateManagerMetrics interface {
+	// ObserveFindAvailableStates records the wall time of one FindAvailableStates call as the engine
```
matthew1001, line 48: Are we expecting these to be primarily recorded at the coordinator now?

core/go/internal/statemgr/metrics/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+	// domain_call_duration_seconds_count{method=assemble} as queries per assemble.
+	ObserveFindAvailableStates(domain string, d time.Duration)
+	// ObserveStateQueryDB records the wall time of the DB round-trip alone (the pgx/Postgres query that
+	// Go profiles cannot see past), for every state read the store makes rather than only the available
+	// ones. An available-state query's round-trip falls inside its ObserveFindAvailableStates bracket.
+	ObserveStateQueryDB(domain string, d time.Duration)
+}
+
+type stateManagerMetrics struct {
+	findAvailableStates *prometheus.HistogramVec
```
matthew1001, line 61: Just checking...

 - `findAvailableStates` - only for finding available states
 - `stateQueryDB` applies to all state DB queries

so things like `FindAvailableNullifiers` don't have their own metric, but we get some measure of their performance from the latter?

matthew1001 (approved):
LGTM, minor questions/comments

core/go/internal/statemgr/metrics/metrics.go
annamcallister: I felt so conscious of what looked like a series of hardcoded arbitrary numbers that I really felt the need to justify it 😂 

Let me see what I can shorten it to which still retains the explanation of why these are the useful buckets

core/go/internal/statemgr/metrics/metrics.go
annamcallister: No both of these will be recorded on the originator who makes the `FindAvailableStates` call

core/go/internal/statemgr/metrics/metrics.go
annamcallister: yes that's true- they are the metrics that have been used for analysing and optimising this particular use case. 

I've tried to only deliver what's been useful rather than being speculatively complete.

More metrics for things like `FindAvailableNullifiers` could come later

PR #1335, opened 2026-08-25
feat(java-sdk): add blockindex, registry, and transport RPC clients

divyam-jha123:
## What

First PR toward #1239 (remaining RPC modules). Adds three of the five remaining namespace clients on top of the existing async JSON-RPC transport, each with its request/response data models and unit tests:

- **`BlockIndexClient`** (`bidx_*`) — all 10 methods: `getBlockByNumber`, `getTransactionByHash`, `getTransactionByNonce`, `getBlockTransactionsByNumber`, `getTransactionEventsByHash`, `queryIndexedBlocks`, `queryIndexedTransactions`, `queryIndexedEvents`, `getConfirmedBlockHeight`, `decodeTransactionEvents`.
- **`RegistryClient`** (`reg_*`) — `registries`, `queryEntries`, `queryEntriesWithProps`, `getEntryProperties`.
- **`TransportClient`** (`transport_*`) — `nodeName`, `localTransports`, `localTransportDetails`, `peers`, `peerInfo`, `queryReliableMessages`, `queryReliableMessageAcks`.

## Data models (in `core`)

- `blockindex/`: `IndexedBlock`, `IndexedTransaction`, `IndexedEvent`, `EventWithData`, `EthTransactionResult`
- `registry/`: `RegistryEntry`, `RegistryProperty`, `RegistryEntryWithProperties`, `ActiveFilter`
- `transport/`: `PeerInfo`, `PeerStats`, `ReliableMessage`, `ReliableMessageAck`, `ReliableMessageAckNoMsgID`, `ReliableMessageType`

## Notes

- On-chain-location and active-flag fields are flattened directly into the registry types (consistent with the existing `TransactionInput` style).
- `getBlockByNumber` returns `IndexedBlock` as its signature intends.

## Testing

- Per-method client tests using the in-process `MockJsonRpcServer`, asserting both the deserialized result and the outgoing method name / params, plus an error-path test per client.
- Round-trip / equals / toString tests for every new data model.
- `./gradlew :sdk:java:core:check :sdk:java:client:check` passes (tests, Checkstyle, JaCoCo ≥ 0.78, Javadoc doclint, Spotless).

## Scope

WebSocket subscriptions, statestore, pgroup, and the top-level client facade are out of scope here and land in follow-up PRs under #1239.

### DIFF
--- a/sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/blockindex/EthTransactionResult.java
+++ b/sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/blockindex/EthTransactionResult.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright contributors to Paladin, an LFDT project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.lfdt.paladin.sdk.core.blockindex;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ * The outcome of an indexed Ethereum transaction. Serializes to its lower-case JSON token; parsing
+ * is case-insensitive.
+ */
+public enum EthTransactionResult {
+
+  /** The transaction reverted or otherwise failed. */
+  FAILURE("failure"),
+  /** The transaction executed successfully. */
+  SUCCESS("success");
+
+  private final String jsonValue;
+
+  EthTransactionResult(final String jsonValue) {
+    this.jsonValue = jsonValue;
+  }
+
+  /**
+   * The JSON token for this result.
+   *
+   * @return the lower-case JSON token
+   */
+  @JsonValue
+  public String jsonValue() {
+    return jsonValue;
+  }
+
+  /**
+   * Resolves a result from its JSON token, case-insensitively.
+   *
+   * @param s the JSON token to resolve
+   * @return the matching result
+   * @throws IllegalArgumentException if {@code s} is null or not a known result
+   */
+  @JsonCreator
+  public static EthTransactionResult fromJson(final String s) {
+    if (s != null) {
+      for (EthTransactionResult r : values()) {
+        if (r.jsonValue.equalsIgnoreCase(s)) {
+          return r;
+        }
+      }
+    }
+    throw new IllegalArgumentException("unknown transaction result: " + s);
+  }
+}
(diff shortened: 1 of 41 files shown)


sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/blockindex/EthTransactionResult.java
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * Resolves a result from its JSON token, case-insensitively.
+   *
+   * @param s the JSON token to resolve
+   * @return the matching result
+   * @throws IllegalArgumentException if {@code s} is null or not a known result
+   */
+  @JsonCreator
+  public static EthTransactionResult fromJson(final String s) {
```
matthew1001: I have a slightly preference for these types of fields being more descriptive if that's OK. Could we use `resultString` instead of `s`? Similarly something suitable for `r` perhaps

matthew1001 (approved):
Looks good @divyam-jha123 I wonder if we could start bumping the test coverage over the course of these PRs? Maybe try to add a few % points of coverage each PR to slowly work towards 100%?

sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/registry/ActiveFilter.java
```diff
@@ -0,0 +1,67 @@
...
+
+  /**
+   * Resolves a filter from its JSON token, case-insensitively.
+   *
+   * @param s the JSON token to resolve
+   * @return the matching filter
+   * @throws IllegalArgumentException if {@code s} is null or not a known filter
+   */
+  @JsonCreator
+  public static ActiveFilter fromJson(final String s) {
```
matthew1001: Same comment here about `s` and `r`

sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/registry/ActiveFilter.java
```diff
@@ -0,0 +1,67 @@
...
+    if (s != null) {
```
matthew1001: Can we lose the `s != null` check given that `f.equalsIgnoreCase(s)` is safe for `s == null`? If so I think we could remove it from a number of the files in the PR that have this pattern in their `fromJson` function

sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/blockindex/EthTransactionResult.java
divyam-jha123: sure, I think I also missed that, I will prefer more descriptive fields

divyam-jha123: @matthew1001  I renamed the parameters and loop variables descriptively (`resultString/result, filterString/filter, typeString/type`) and dropped the `s != null` guard, since `equalsIgnoreCase(null)` returns false and falls through to the same `IllegalArgumentException`. I applied the same to the four already-merged enums with this pattern (`TransactionType, SubmitMode, StateMutability, EntryType`) so all seven read identically.

divyam-jha123:
Also about the test covarage ,
I'd like to walk `sdk:java:core` towards full coverage over the remaining project, so I've started ratcheting the JaCoCo floor - INSTRUCTION minimum raised from `0.78 to 0.94` in sdk/java/build.gradle, set just under core's current 94.72% (`client is at 97.78%`). The floor is deliberately tight so new under-tested code fails the build rather than riding down the average; the intent is to raise it again as each PR clears the next percent.

Happy to loosen it if you'd rather have more headroom per PR.

the core test covarage - 
image

divyam-jha123:
also once #1335 merges, the 0.94 floor applies to #1336 and the pgroup branch, and their new DTO packages will dilute core. I chose to set it now and fix them if red.

matthew1001 (approved):
Thanks for the updates and for the test coverage bump, love the progress

PR #1336, opened 2026-08-25
feat(java-sdk): add state store RPC client

divyam-jha123:
## What

Second PR toward #1239 (remaining RPC modules). Adds the state store namespace client on top of the existing async JSON-RPC transport, with its data models and unit tests.

- **`StateStoreClient`** (`pstate_*`) — all 7 methods: `listSchemas`, `storeState`, `queryStates`, `queryContractStates`, `queryNullifiers`, `queryContractNullifiers`, `transferPrivateState`.

Note the wire prefix is `pstate_`, not `statestore_`.

## Data models (in `core`, package `statestore`)

- `Schema`, `State`, `StateConfirmRecord`, `StateReadRecord`, `StateSpendRecord`, `StateLock`, `StateNullifier`
- `SchemaType`, `StateLockType` (enums)
- `StateStatusQualifier` — a value type covering the standard qualifiers (`available` / `confirmed` / `unconfirmed` / `spent` / `all`) or a transaction id via `forTransaction(UUID)`

## Behaviour note

`queryStates` and `queryContractNullifiers` **always send** the trailing `StateStatusQualifier` parameter, so the query is scoped correctly on the server. This is deliberate: all four query methods take and forward the qualifier. The tests assert the qualifier is present in the outgoing params.

`State` flattens Go's embedded `StateBase` fields (id, created, domain, schema, contractAddress, data) directly, consistent with the existing `TransactionInput` style.

## Testing

- Per-method client tests using the in-process `MockJsonRpcServer`, asserting both the deserialized result and the outgoing method name / params (including qualifier presence and position), plus an error-path test.
- Round-trip / equals / toString tests for every new data model, including the standard-value and UUID paths of `StateStatusQualifier`.
- `./gradlew :sdk:java:core:check :sdk:java:client:check` passes (tests, Checkstyle, JaCoCo ≥ 0.78, Javadoc doclint, Spotless).

## Scope

WebSocket subscriptions, pgroup, and the top-level client facade are out of scope here and land in the final PR under #1239.

### DIFF
--- a/sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/statestore/SchemaType.java
+++ b/sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/statestore/SchemaType.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright contributors to Paladin, an LFDT project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.lfdt.paladin.sdk.core.statestore;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ * The kind of a state schema. Serializes to its lower-case JSON token; parsing is case-insensitive.
+ */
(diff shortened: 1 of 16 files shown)


sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/statestore/SchemaType.java
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Resolves a schema type from its JSON token, case-insensitively.
+   *
+   * @param s the JSON token to resolve
+   * @return the matching schema type
+   * @throws IllegalArgumentException if {@code s} is null or not a known schema type
+   */
+  @JsonCreator
+  public static SchemaType fromJson(final String s) {
+    if (s != null) {
```
matthew1001: See related comments in https://github.com/LFDT-Paladin/paladin/pull/1335 about the style in these `fromJson` methods

matthew1001 (approved):
Approving with one comment about `fromJson` again and same comment about slowly bumping test coverage

divyam-jha123:
@matthew1001 Increase the test coverage , and made the name fields changes 

image

PR #1337, opened 2026-08-26
Transport test fixes

annamcallister:
Fixes a nil pointer exception that occurs after a test has ran and keeps causing GHA failures on PRs.

Fixes a race identified by running tests with the `-race` flag


matthew1001 (approved):
LGTM

PR #1339, opened 2026-08-28
feat(java-sdk): add privacy group RPC client

divyam-jha123:
Part of #1239 — adds the `pgroup_*` (privacy groups) RPC namespace client to the Java SDK.

This is the last of the three namespace PRs under #1239, alongside #1335 (blockindex/registry/transport) and #1336 (statestore). Branched off `java-sdk-main` and independent of both, so it can be reviewed and merged in any order.

### What's here

**`core/…/privacygroup/`** — the data models:

- `PrivacyGroup`, `PrivacyGroupInput`, `PrivacyGroupTXOptions`
- `PrivacyGroupEVMTXInput`, `PrivacyGroupEVMCall`
- `PrivacyGroupMessage`, `PrivacyGroupMessageInput`
- `PrivacyGroupMessageListener`, `PrivacyGroupMessageListenerFilters`, `PrivacyGroupMessageListenerOptions`

**`core/…/transaction/PublicTxOptions`** — a new shared type. Every other request body in the SDK flattens the base-ledger submission options inline (as `TransactionInput` does); `PrivacyGroupEVMTXInput` is the one that carries them as a nested `publicTxOptions` object, so they needed a type of their own. Note this means `gas`/`value` legitimately appear twice on that body — once flattened as the EVM transaction's own values, once nested as the options for the public transaction that carries it. There's an explicit test pinning that shape.

**`client/…/privacygroup/PrivacyGroupClient`** — all 16 request/response methods: group create/get/query, `sendTransaction`/`call`, message send/get/query, and the six message-listener operations.

### Out of scope

`pgroup_subscribe` / `_unsubscribe` / `_ack` / `_nack` / `_subscription` are deliberately not implemented — they need a WebSocket transport, which is tracked separately by #1240. This matches the "WebSocket subscriptions are out of scope" decision for #1239.

The top-level `PaladinClient` facade also isn't here. It has to reference all seven namespace clients, so it can't compile until #1335, #1336 and this PR have merged; it follows as its own PR.

### Testing

- `PrivacyGroupTypesTest` (core, 15 tests) — JSON round-trips for every DTO, plus the omit-when-unset behaviour and the zero-timestamp normalisation the node relies on.
- `PrivacyGroupClientTest` (client, 13 tests) — one per method group against `MockJsonRpcServer`, asserting method names and parameter order/position on the wire, and RPC error propagation.

`./gradlew :sdk:java:core:check :sdk:java:client:check` passes: tests, Checkstyle, Javadoc doclint, Spotless, and JaCoCo verification. Instruction coverage is 0.8820 (core) and 0.9765 (client), against the 0.78 minimum.

### Reviewer note

The DTOs are derived from the `pldapi` privacy-group structs, and the tests assert self-consistent round-trips rather than agreement with a live node. The `publicTxOptions` nesting on `PrivacyGroupEVMTXInput` is the one field shape I'd most want a second pair of eyes on.

### DIFF
--- a/sdk/java/client/src/main/java/org/lfdt/paladin/sdk/client/privacygroup/PrivacyGroupClient.java
+++ b/sdk/java/client/src/main/java/org/lfdt/paladin/sdk/client/privacygroup/PrivacyGroupClient.java
@@ -1,0 +1,233 @@
+/*
+ * Copyright contributors to Paladin, an LFDT project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.lfdt.paladin.sdk.client.privacygroup;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JsonNode;
+import java.util.List;
+import java.util.Objects;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import org.lfdt.paladin.sdk.client.rpc.RpcClient;
+import org.lfdt.paladin.sdk.core.privacygroup.PrivacyGroup;
+import org.lfdt.paladin.sdk.core.privacygroup.PrivacyGroupEVMCall;
+import org.lfdt.paladin.sdk.core.privacygroup.PrivacyGroupEVMTXInput;
+import org.lfdt.paladin.sdk.core.privacygroup.PrivacyGroupInput;
+import org.lfdt.paladin.sdk.core.privacygroup.PrivacyGroupMessage;
+import org.lfdt.paladin.sdk.core.privacygroup.PrivacyGroupMessageInput;
+import org.lfdt.paladin.sdk.core.privacygroup.PrivacyGroupMessageListener;
+import org.lfdt.paladin.sdk.core.query.QueryJSON;
+import org.lfdt.paladin.sdk.core.types.EthAddress;
+import org.lfdt.paladin.sdk.core.types.HexBytes;
+
+/**
+ * Client for the {@code pgroup_*} RPC namespace (privacy groups).
+ *
+ * <p>Each method maps one-to-one to a JSON-RPC call on the underlying {@link RpcClient} and returns
+ * a {@link CompletableFuture}; failures complete it exceptionally with a {@code PaladinException}
+ * subtype.
+ *
+ * <p>The namespace covers three things: the groups themselves, the EVM transactions and calls
+ * executed inside them, and the off-chain messages members exchange. Message delivery is driven by
+ * named listeners; subscribing to a listener's stream needs a WebSocket transport and is not part
+ * of this client.
+ */
+public final class PrivacyGroupClient {
+
+  private final RpcClient rpc;
+
+  /**
+   * Creates a client over the given RPC transport.
+   *
+   * @param rpc the RPC client used to make calls; must not be {@code null}
+   */
+  public PrivacyGroupClient(final RpcClient rpc) {
+    this.rpc = Objects.requireNonNull(rpc, "rpc");
+  }
+
+  /**
+   * Creates a privacy group ({@code pgroup_createGroup}).
+   *
+   * @param spec the specification for the new group
+   * @return a future completing with the created group; its contract address is filled in once the
+   *     genesis transaction is mined
+   */
+  public CompletableFuture<PrivacyGroup> createGroup(final PrivacyGroupInput spec) {
+    return rpc.callRpc(PrivacyGroup.class, "pgroup_createGroup", spec);
+  }
+
+  /**
+   * Looks up a privacy group by its identifier ({@code pgroup_getGroupById}).
+   *
+   * @param domainName the domain the group belongs to
+   * @param id the group identifier
+   * @return a future completing with the group, or {@code null} if no such group exists
+   */
+  public CompletableFuture<PrivacyGroup> getGroupById(final String domainName, final HexBytes id) {
+    return rpc.callRpc(PrivacyGroup.class, "pgroup_getGroupById", domainName, id);
+  }
+
+  /**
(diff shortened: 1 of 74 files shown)


divyam-jha123:
image

image


current test coverage status

sdk/java/client/src/main/java/org/lfdt/paladin/sdk/client/privacygroup/PrivacyGroupClient.java
```diff
@@ -0,0 +1,233 @@
...
+import org.lfdt.paladin.sdk.core.privacygroup.PrivacyGroupInput;
+import org.lfdt.paladin.sdk.core.privacygroup.PrivacyGroupMessage;
+import org.lfdt.paladin.sdk.core.privacygroup.PrivacyGroupMessageInput;
+import org.lfdt.paladin.sdk.core.privacygroup.PrivacyGroupMessageListener;
+import org.lfdt.paladin.sdk.core.query.QueryJSON;
+import org.lfdt.paladin.sdk.core.types.EthAddress;
+import org.lfdt.paladin.sdk.core.types.HexBytes;
+
+/**
+ * Client for the {@code pgroup_*} RPC namespace (privacy groups).
```
matthew1001, line 36: It looks like `invokeRPC` isn't included in the list of supported RPC methods. Any reason not to also include it?

sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/privacygroup/PrivacyGroupEVMCall.java
```diff
@@ -0,0 +1,357 @@
...
+  private final HexUint256 value;
+  private final JsonNode input;
+  private final AbiEntry function;
+  private final HexBytes bytecode;
+  private final String block;
+  private final String dataFormat;
+
+  @JsonCreator
+  PrivacyGroupEVMCall(
+      @JsonProperty("domain") final String domain,
```
matthew1001, line 67: Could we remove some of the duplication here by re-using your `PrivacyGroupEVMTXInput` as an input field, and using `JsonUnwrapped` like you've done elsewhere e.g. in `TransactionCall`?

sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/privacygroup/PrivacyGroupTXOptions.java
```diff
@@ -0,0 +1,201 @@
...
+ *
+ * <p>The public-transaction options ({@link #gas()}, {@link #value()}, {@link
+ * #maxPriorityFeePerGas()}, {@link #maxFeePerGas()}) are flattened onto the flat JSON wire form
+ * alongside the idempotency key. Supplying either gas-pricing field fixes pricing for the
+ * transaction, disabling the node's gas-pricing engine.
+ */
+@JsonPropertyOrder({"idempotencyKey", "gas", "value", "maxPriorityFeePerGas", "maxFeePerGas"})
+public final class PrivacyGroupTXOptions {
+
+  private final String idempotencyKey;
```
matthew1001, line 36: I've got a similar comment about the duplication of fields, between this class and your new `PublicTxOptions`. Is it possible to remove that duplication by passing in a `PublicTxOptions` field?

sdk/java/client/src/test/java/org/lfdt/paladin/sdk/client/tx/TxBuilderTest.java
```diff
@@ -728,7 +728,10 @@ void waitForReceiptTimesOutWhenNoReceiptArrives() throws IOException {
           assertTrue(
               elapsed.toMillis() >= 60,
               "must wait out the full timeout, waited " + elapsed.toMillis() + "ms");
-          assertTrue(server.requestCount() > 2, "should have polled repeatedly");
+          // One send plus at least one poll. How many polls fit inside the timeout depends on
+          // the machine, so repeated polling is asserted by
+          // waitForReceiptPollsUntilTheReceiptArrives instead.
```
matthew1001, line 733: This might be a typo - there's a `waitForReceiptPollsUntilTheReceiptLands` function - is it that one the comment is referring to?

matthew1001 (commented):
Thanks @divyam-jha123, and apologies for the slow review. Looks good, just a few comments.

One broader comment, which I'm happy to be deferred if you'd prefer to raise an issue and track it there, is that there's a difference between which classes you've added `equals()` and `hashCode()` implementations to, and which you haven't. I suspect we probably want custom implementations on most of the core classes, so it's possible to compare objects and work with Sets etc. correctly. I'll let you decide whether to update anything in this PR or track it as a TODO

sdk/java/client/src/test/java/org/lfdt/paladin/sdk/client/tx/TxBuilderTest.java
divyam-jha123: yes, it `waitForReceiptPollsUntilTheReceiptLands `, my bad

sdk/java/client/src/main/java/org/lfdt/paladin/sdk/client/privacygroup/PrivacyGroupClient.java
divyam-jha123: Added `pgroup_invokeRPC` in commit 80c4709d6. It accepts domainName, groupId, StateStatusQualifier, and DomainInvokeRPC in the server-defined order. I also added client and DTO tests covering the parameter order, request serialization, and JSON result. Thanks for catching the missing method.

sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/privacygroup/PrivacyGroupEVMCall.java
divyam-jha123: Updated in commit 324e397ca to reuse `PrivacyGroupEVMTXInput` with `@JsonUnwrapped`, following the `TransactionCall` pattern. I also added regression coverage in 780a00491 to preserve flat-form deserialization and ensure send-only fields such as idempotencyKey and publicTxOptions are not included in read-only calls.

sdk/java/core/src/main/java/org/lfdt/paladin/sdk/core/privacygroup/PrivacyGroupTXOptions.java
divyam-jha123: Updated in commit 793d11dfc. `PrivacyGroupTXOptions` now stores a PublicTxOptions instance instead of duplicating its four fields. The existing flat JSON wire format and round-trip deserialization are preserved through delegated accessors. Commit 780a00491 also adds regression coverage for empty options and value equality.

matthew1001 (approved):
Thanks for the updates @divyam-jha123 👍

PR #1342, opened 2026-09-01
Public tx mgr logging

annamcallister:
All these log lines are on very hot paths that get called multiple times per transaction so small numbers of unnecessary allocs per line can drive a lot of GC.

* Stop calling `.String()` on format arguments - let the formatter do this _only_ if the log is included at the current level
* Stop including time arguments - these drive allocs even if not passed with `.String()` and we get a timestamp included in the log already. These are a carry over from an earlier version of code where I think they contributed to performance analysis, but at debug level the same data would be better derived from metrics
* Fix a bug where timelines were built but never actually printed (not perf related)

### DIFF
--- a/common/go/pkg/log/log.go
+++ b/common/go/pkg/log/log.go
@@ -241,6 +241,10 @@ func IsDebugEnabled() bool {
 	return atomLevel.Enabled(zapcore.DebugLevel)
 }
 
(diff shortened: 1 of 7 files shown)


matthew1001 (approved):
LGTM

PR #1343, opened 2026-09-01
use sync channels not eventually statements

annamcallister:
Factor eventually statements out of block indexer tests and replace with synchronisation channels. This stops us from having to give a timeout to each statement. It means the tests will run faster in most cases, and where there is an unexpected wait (e.g. runner contention but only briefly) stops spurious test failures.

### DIFF
--- a/core/go/pkg/blockindexer/event_streams.go
+++ b/core/go/pkg/blockindexer/event_streams.go
@@ -38,27 +38,28 @@ import (
 )
 
 type eventStream struct {
-	ctx               context.Context
-	cancelCtx         context.CancelFunc
-	bi                *blockIndexer
-	definition        *EventStreamDefinition
-	signatures        map[string]bool
-	signatureList     []pldtypes.Bytes32
-	batchSize         int
-	batchTimeout      time.Duration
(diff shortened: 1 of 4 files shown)


core/go/pkg/blockindexer/event_streams.go
```diff
@@ -392,6 +394,10 @@ func (bi *blockIndexer) startEventStreams() {
 			_ = es.start(false)
 		}
 	}
+	select {
+	case bi.eventStreamsStarted <- struct{}{}:
+	default: // nobody is listening, or a previous start is still unread
```
matthew1001: if `eventStreamsStarted` is a buffered channel the "nobody is listening" part of this communicate doesn't seem accurate? Should it just be

```
// A previous start is still unread
```

?

matthew1001: If so similar comment for `checkpointCommitted`

matthew1001 (approved):
Approved with small comment

core/go/pkg/blockindexer/event_streams.go
annamcallister: changed

PR #1349, opened 2026-09-02
Fix double-submission race on orchestrator swap

atharrva01:
When a public tx orchestrator gets evicted (idle or stale timeout, or a pool-fairness swap under load), Stop() only signalled the polling loop to exit. It never cancelled the signing and submission goroutines that were already in flight. Once the orchestrator was removed from the pool, a fresh orchestrator for the same signing address could start right away and pick up the same pending transactions, so two goroutines could end up signing and submitting for the same nonce at the same time.

This gives each orchestrator its own cancellable context plus a WaitGroup that tracks in-flight stage actions. Stop() now cancels that context, and the loop waits for any outstanding action to actually finish before the orchestrator is marked stopped and removable. That way a new orchestrator for the same address can never run concurrently with a leftover goroutine from the old one.

## Test plan
- [x] `go test ./core/go/internal/publictxmgr/...` (including the real-DB end-to-end lifecycle test), run repeatedly
- [x] `go test -race ./core/go/internal/publictxmgr/...`

### DIFF
--- a/core/go/internal/publictxmgr/in_flight_transaction_stage_controller.go
+++ b/core/go/internal/publictxmgr/in_flight_transaction_stage_controller.go
@@ -691,7 +691,19 @@ func (it *inFlightTransactionStageController) executeAsync(funcToExecute func(),
 	if it.testOnlyNoActionMode {
 		return
 	}
+	// Tracked on the owning orchestrator's WaitGroup so Stop() can block until this goroutine finishes -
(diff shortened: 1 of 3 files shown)


atharrva01:
hi @matthew1001 , this pr fixes a race where an evicted orchestrator's leftover goroutine could double-submit a transaction at the same nonce as its replacement.

matthew1001 (approved):

PR #1351, opened 2026-09-07
Fix nil pointer panic in PrepareTransaction

atharrva01:
The Transaction field on a domain plugin's PrepareTransactionResponse is optional in the proto, so a plugin can return no error and still leave it unset. We were dereferencing it right away without checking, which panics. Since this runs on the coordinator's event loop with nothing recovering panics there, it takes down the whole node, not just the one transaction.

Added a nil check that returns a normal error instead, in both places that read this field (PrepareTransaction and WrapPrivacyGroupEVMTX).

Also added a test that mocks a plugin returning an empty response and checks we get an error back instead of a crash.

Test plan:
- go test ./internal/domainmgr/... passes, including the new test

### DIFF
--- a/core/go/internal/domainmgr/private_smart_contract.go
+++ b/core/go/internal/domainmgr/private_smart_contract.go
@@ -425,6 +425,9 @@ func (dc *domainContract) PrepareTransaction(ctx context.Context, dqc components
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


atharrva01:
hi @matthew1001 @annamcallister , would love your thoughts on this one.. thanks :)

PR #1357, opened 2026-09-14
Noto endorse error classification

annamcallister:
Assembly and endorsement requests to a domain can result in two types of error response:

#### Revert
* returned as part of the assemble/endorse response
* returned when either the transaction inputs (at assembly) or proposed assembled transaction (at endorsement) are invalid and break the rules of the domain

#### Internal error
* returned in the error parameter
* returned when an unexpected error occurs (e.g. db query failure, timeout) or for errors which should be unreachable except in the case of a code bug

Noto has only implemented this distinction so far for assembly, with all errors at endorsement being returned as internal errors. This impacts how the sequencer responds. An an endorsement request to a domain is retried up to a configurable limit when an internal error occurs, and the transaction is repooled for a later sequencing attempt if the retry limit is reached. When a revert is returned by a domain, the result is immediately considered terminal, and the transaction is finalised as reverted.

The existing implementation for assembly returns a `revert` boolean alongside the error. This approach does not extend well to endorsement. Firstly because there are many more places an error can occur at endorsement and the boolean becomes unwieldy. Secondly because some errors which are internal errors at assembly are reverts at endorsement, since assembly is validating data that has just been produced by this node as an earlier part of assembly, whereas endorsement is validating data that has been produced by an originator likely running on a different node.

This PR modifies `Assemble` and `Endorse` to return a `NotoDomainError`. A `NotoDomainError` implements two functions which report whether the final result of assembly or endorsement should be a revert or internal error.  It embeds `error` so it is usable without type casting/conversion on all existing domain interfaces, and in places which do not need to make the distinction, but the requiring of it on the more specific `NotoDomainHandler` interface means that adding a new error case to assembly or endorsement will not compile unless the error behaviour for assembly and endorsement is specified. There is no default behaviour; every error case must be individually considered.

**Every error classification in this PR has been decided by me. I would recommend that review of this PR carefully considers each revert as this is the primary behaviour change for endorsement.**

This PR also adds a TODO as there is currently no revert handled on calls in Paladin, which is a feature gap. Because the `ParamValidator` interface is shared across transaction and call handlers, this means that  call `ValidateParam` functions must return a `NotoDomainError` who's classification is always ignored. I've chosen to accept this redundancy since it leaves the code in the correct shape for when the Paladin core feature gap is addressed.


matthew1001 (commented):
I certainly see the need to distinguish between errors and reverts at endorse time.

My main comment on the PR is whether the "should we revert" logic is in the right place. They way the PR is currently written, the common functions define how the error they are about to return should be treated - `internalErr`, `revert` or `endorseOnlyRevert`. 

Did you consider an alternative approach of having common functions return a first-class error type - e.g. `findEthAddressVerifier` could return `ErrFindEthAddressVerifier` - and having the caller (`AssembleTransaction()` or `EndorseTransaction()`) decide what that means? This would probably involve an assemble-specific and endorse-specific list of revertable error types:

Assemble revert errors:
 - `ErrValidateParams`
 - ...

Endorse revert errors:
 - `ErrValidateParams`
 - `ErrFindEthAddressVerifier`
 - ...

There might be reasons this doesn't fly, but if an approach like that worked it might be a better way to a) understand the intended behaviour and b) maintain the code

annamcallister:
That is an option I thought quite carefully about, and I think that while it would read more cleanly, anything that keeps a degree of abstraction between the source of the error and how it is used in the response for assemble/endorse is at greater risk of inappropriate reuse.

In almost all cases the difference between a revert and internal error is the source of the data being validated. If it's been created as part of handling this request its an internal error, if its come from an untrusted source it's a revert. An custom that describes the thing being validated could be reused in a place where the provenance of the data is different, meaning the end classification of the error is different.

I don't think there's anyway to get around this without forcing the question to be explicitly answered when the error is created, even if that has to be in a slightly awkward way. The same argument as above still applies to some degree to the implementation proposed here- as a helper function that classifies the error, could be reused in different places where the source has changed- but that's not really the style of the code. In most places it is specific fields of the transaction/states being validated, as opposed to passing in fields of a specific type to be validated in isolation.

`endorseOnlyRevert` is definitely the clunkiest error helper function in this approach, but there's an tension there with the fact that concept of some errors being classified differently at assemble vs endorse is not straightforward, and does require some very intentional thinking.

I did struggle a bit with the fact that there are a lot of helper/validation functions of which some are inherently assembly/endorsement only, but the code structure does not reflect that. I don't know restructuring there is a way to get more clarity for future development, but I'm aware this PR is already making a lot of changes across the code base.

It may also be that if we've discussed this enough, there is enough understanding of the issue at hand amongst maintainers to pick up on future inconsistencies in a code review, and a more elegant code solution becomes feasible.

An appropriate skill might also ensure agentic development at least highlights the question of error classification to a developer, even where we don't want an agent taking full responsibility for the classification.

annamcallister:
Two alternative implementations:

https://github.com/LFDT-Paladin/paladin/compare/main...noto-endorse-validation-2 has lists which classify the errors, with the look up done at the point of building the assemble/endorse response. I initially tried putting custom errors in but they ended up being pure wrapper to the i18n error underneath, so it made more sense to classify the i18n errors. I'm not keen on this approach. It's very easy to miss an error or to not classify an error that comes back from a dependency. It is the most minimal approach though and includes a skill to try keep lists up to date.

--- 

https://github.com/LFDT-Paladin/paladin/compare/main...noto-endorse-validation-3 I think has more promise. The custom errors I think only make sense if they are used to hold the classifications. Any classification done just by listing types is only ever best effort and that to me feels like a poor choice to make when we're using a typed language which has the ability to enforce full compliance at compile time.

This approach has switched to three interfaces for the errors, which means we can declare more structurally that a function and the errors it returns are only used for assemble/endorse, and would need conscious updating to be used on the other path. This means there's no `endorseOnlyRevert` naming, it's now a typing concern. Because of custom error type reuse, the rational behind the error classification is abstracted away in the `errors.go` file, rather than at the source of the error. The detailed comments including the return types mean that hovering over the error type at the point of use gives the full behaviour of the error in line.

This approach is no less involved than the original proposal in this PR (which I still maintain is a good thing, getting the right error classification is important and shouldn't be abstracted away), but I do think it is less clunky.

annamcallister:
Finally https://github.com/LFDT-Paladin/paladin/compare/main...noto-endorse-validation-4 I think combines what works from this PR and `noto-endorse-validation-3`

* Fully enforced by compiler
* revert decision made at source
* assemble/endorse/both handled more elegantly

It does not have specific error types. I can't find a way to do this which doesn't take decision making away from the source. Specific error types also have a real tendency to describe the failure that occurred, and what we care about is where did the data come from.

matthew1001:
Thanks so much for pulling together some alternative approaches.

My preference is option 3 (https://github.com/LFDT-Paladin/paladin/compare/main...noto-endorse-validation-3 to be clear). I think it strikes a good balance of requiring committers to think about what a function returns ("none of these existing error types seems to match what I'm trying to return - I better create a new one and that makes me think about the assemble and endorse behaviours") and code clarity at the calling site (`return invalidStateList{err}` reads pretty clearly imho - no mention on this line of code about assembly or endorsement which feels better to me). 

I agree that option 2 doesn't seem to fit as well. The big lists of

```
...
msgs.MsgUnexpectedSchema:           {assemble: true, endorse: true},
msgs.MsgNoStatesSpecified:          {assemble: true, endorse: true},
msgs.MsgDuplicateNullifierInList:   {assemble: true, endorse: true},
...
```

feel somehow a little clunkier than option 3.

Option 4 moves away (imho) from the clarity of option 3 when deciding what to return:

```
return endorseError{...} 
```

and especially

```
return nil, assembleOrEndorseError{...}
```

...all of a sudden I'm mentioning assembly and/or endorsement in some common function that just feels harder code to comprehend at a glance.

annamcallister:
I've closed this PR - https://github.com/LFDT-Paladin/paladin/pull/1370 raised to replace it with the preferred implementation

PR #1360, opened 2026-09-15
Fix grpc transport connection leak

annamcallister:
We have been leaking a connection every time that a peer has been reaped. We have also been leaking connection streams every time a send is unsuccessful.

This PR fixes that by
* tracking the lifecycle of the sender goroutine and underlying transport connection separately
* always closing the the connection when the transport `outboundConn` is closed
* starting all streams with their own context that can be cancelled after a send


transports/grpc/internal/grpctransport/grpc_transport.go
```diff
@@ -274,8 +274,7 @@ func (t *grpcTransport) DeactivatePeer(ctx context.Context, req *prototk.Deactiv
 
 	existing := t.outboundConnections[req.NodeName]
 	if existing != nil {
-		// Replace an existing connection - unexpected as Paladin shouldn't do this
-		log.L(ctx).Warnf("replacing existing activation for node '%s'", req.NodeName)
+		log.L(ctx).Debugf("deactivating peer '%s'", req.NodeName)
```
annamcallister, line 277: fixes a log that was copied incorrectly from the activate function (just tidy up, not related to the main fix)

core/go/internal/transportmgr/peer.go
```diff
@@ -194,8 +195,9 @@ func (tm *transportManager) reapPeer(p *peer) {
 	log.L(p.ctx).Infof("peer %s deactivating", p.Name)
 	p.close()
 
-	if p.senderStarted.Load() {
+	if p.transportActivated {
```
matthew1001, line 198: Mainly for my understanding, was this a timing window,where `p.senderStarted` might not have been set by the time the peer was reaped? Or was did we fail to clean up 100% of the time?

transports/grpc/internal/grpctransport/outbound_conn.go
```diff
@@ -72,25 +81,55 @@ func (t *grpcTransport) newConnection(ctx context.Context, nodeName string, tran
...
+	// counts as an active call (so the ClientConn never goes idle) and keeps a goroutine alive.
+	// We never read from the stream, so cancelling the context is what releases it.
+	_ = oc.stream.CloseSend()
+	oc.streamCancel()
+	oc.stream = nil
+	oc.streamCtx = nil
+	oc.streamCancel = nil
+}
+
+func (oc *outboundConn) ensureStream() error {
```
matthew1001, line 118: Does `ensureStream()` also need to be called with `sendLock` held? Looks like do hold the lock but not on first connection, but a comment confirming either way might be good

matthew1001 (approved):
Approved with a couple of minor questions for clarification

core/go/internal/transportmgr/peer.go
annamcallister: There's a few timing conditions where `p.sendStarted` can be `false` while the outbound connection has already started. One is on creation where to sender goroutine isn;t started until the connection is established- but that should be impossible to hit now we've fixed the bug where a newly created peer could be reaped before first use.

The main way it could be hit is if the sender goroutine (in transport manager) can stop after a send failure, while the connection in the transport plugin is still active. The goroutine is only restarted if there's another send. So `p.senderStarted` can't be used. There's also a race between deferred closing actions.

The lifecycle of the two things is not strictly tied together, hence the need to track them separately.

transports/grpc/internal/grpctransport/outbound_conn.go
annamcallister: comment added

